PR #7, opened 2019-07-16
Renamed .hgignore to .gitignore

olehermanse:
Prevents committing temporary files, like test logs, pycache, etc.


benmoran56:
Probably a good idea to also ignore popular IDE files, such as the Pycharm .idea subfolder. I'll make another addition for that.

PR #10, opened 2019-07-19
Added new hello_world example

olehermanse:
A few advantages over the one in programming_guide:

* Easier to find for beginners, not hidden in subdirectory
* Less noise from copyright blob, unnecessary for small examples
* Faster to type examples/hello_world.py without subdirectory

Signed-off-by: Ole Herman Schumacher Elgesem <oleherman93@gmail.com>


benmoran56:
Looks good to me!

Issue #11, opened 2019-07-20
Request: friendlier Display API for X11


ixjlyons:
I've found that I need to pass a `ctypes.c_char_p` type as the name argument to `Display` to get it to work, which is somewhat cumbersome:

```python
import pyglet
import ctypes
display = pyglet.canvas.Display(ctypes.c_char_p(":2".encode('ascii'))
```

Would it be reasonable to detect a `str` type passed to `XlibDisplay` and do that conversion for the user so you could just write `display = pyglet.canvas.Display(":2")`?

edit: also note the docstring for [`Display.name`](https://pyglet.readthedocs.io/en/pyglet-1.3-maintenance/modules/canvas.html#pyglet.canvas.Display.name) says it's a str.

benmoran56:
Fixed via PR #12 
Thanks again for your contribution!

PR #12, opened 2019-07-20
Updated display docs and improved XlibDisplay interface

ixjlyons:
I'm new to pyglet and while reading the [context page](https://pyglet.readthedocs.io/en/pyglet-1.3-maintenance/programming_guide/context.html) I came across what seems to be some outdated information. I did my best to update the docs to reflect the current state of the library and clean up some minor issues.

Since the docs claim you can pass a string to `XlibDisplay`, I went ahead and added that functionality as mentioned in #11.

### DIFF
--- a/doc/programming_guide/context.txt
+++ b/doc/programming_guide/context.txt
@@ -39,7 +39,7 @@ of its properties::
     c_int(1)
     >>> config.stereo
     c_int(0)
-    >>> config.sample_buffers     
+    >>> config.sample_buffers
     c_int(0)
 
 Note that the values of the config's attributes are all ctypes instances.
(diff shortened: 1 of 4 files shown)


benmoran56:
Thanks Kenneth!

Lots of legacy APIs were removed in 1.4, and we missed that part in the docs.

The Display update should be fine, but could you confirm that it works on Python 2 and 3? We still need to support both for v1. 4.

ixjlyons:
Ah, I did overlook Python 2 support. The explicit conversion isn't *necessary* for Python 2 but it does work. I tested in Python 3.7.3 and 2.7.16 virtual environments using Xephyr to check combinations of multiple screens with/without xinerama (mostly to check my understanding of the name/x_screen arguments).

benmoran56:
Thanks for that. Merged!

PR #13, opened 2019-07-21
 Changed to explicit imports for cocoapy

olehermanse:
`import *` can be problematic, especially when we want to update this library.
The new version (PyCocoa) has classes like Screen, which conflict with pyglet
classes. More explicit imports can also highlight what's missing in a newer
version, without having to run the affected code (Import will fail).

### DIFF
--- a/pyglet/input/darwin_hid.py
+++ b/pyglet/input/darwin_hid.py
@@ -35,13 +35,19 @@
 from __future__ import print_function
 from __future__ import absolute_import
 from builtins import object
+
+from ctypes import cdll, util, CFUNCTYPE, byref, c_void_p
+from ctypes import c_int, c_ubyte, c_bool, c_uint32, c_uint32, c_uint64
+
 # Uses the HID API introduced in Mac OS X version 10.5
 # http://developer.apple.com/library/mac/#technotes/tn2007/tn2187.html
 
 import sys
 __LP64__ = (sys.maxsize > 2**32)
 
-from pyglet.libs.darwin.cocoapy import *
+from pyglet.libs.darwin.cocoapy import CFSTR, CFIndex, CFTypeID, \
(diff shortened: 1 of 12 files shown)


pyglet/input/darwin_hid.py
```diff
@@ -35,13 +35,19 @@
 from __future__ import print_function
 from __future__ import absolute_import
 from builtins import object
+
```
benmoran56: Instead of using the line break ”\",  it's more readable to just have two lines of imports. For example: 
```from ctypes import cdll, util, CFUNCTYPE, c_int, c_ubyte, c_bool, c_void_p, 
from ctypes import c_uint32, c_uint32, c_uint64, byref
```

olehermanse:
@benmoran56 Thanks for the review. I actually thought that would re-evaluate the module, but apparently not:

```
olehermanse@OHMBP ~ $ cat > module.py
def a():
  pass

def b():
  pass

print("Module evaluated")
olehermanse@OHMBP ~ $ cat > use.py
import sys
sys.path.insert(0, ".")

from module import a
from module import b
olehermanse@OHMBP ~ $ python3 use.py 
Module evaluated
olehermanse@OHMBP ~ $ python use.py 
Module evaluated
olehermanse@OHMBP ~ $ python3
Python 3.7.4 (default, Jul  9 2019, 18:13:23) 
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.path.insert(0, ".")
>>> import module
Module evaluated
>>> import module
>>> 
```

I'll try that later today.

olehermanse:
@benmoran56 Split all the imports into multiple statements. Should be ready now.

benmoran56:
Thanks, and sorry to nitpick!

PR #20, opened 2019-07-26
Fix a few missing imports and unresolved references.

benmoran56:
I ran a code inspection on the codebase with the PyCharm inspector, and it turned up a few missing imports.

### DIFF
--- a/pyglet/canvas/base.py
+++ b/pyglet/canvas/base.py
@@ -183,14 +183,13 @@ class Screen(object):
         """
         configs = None
         if template is None:
-            for template_config in [
-                gl.Config(double_buffer=True, depth_size=24),
-                gl.Config(double_buffer=True, depth_size=16),
-                None]:
+            for template_config in [gl.Config(double_buffer=True, depth_size=24),
+                                    gl.Config(double_buffer=True, depth_size=16),
+                                    None]:
                 try:
                     configs = self.get_matching_configs(template_config)
                     break
-                except NoSuchConfigException:
+                except window.NoSuchConfigException:
                     pass
         else:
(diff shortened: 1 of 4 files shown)


pyglet/canvas/base.py
```diff
@@ -183,14 +183,13 @@ def get_best_config(self, template=None):
         """
         configs = None
         if template is None:
-            for template_config in [
-                gl.Config(double_buffer=True, depth_size=24),
-                gl.Config(double_buffer=True, depth_size=16),
-                None]:
+            for template_config in [gl.Config(double_buffer=True, depth_size=24),
+                                    gl.Config(double_buffer=True, depth_size=16),
+                                    None]:
```
olehermanse: I actually prefer how it was before, this style forces things so far out to the right (but this isn't a very bad example). Just my opinion.

pyglet/window/cocoa/pyglet_delegate.py
```diff
@@ -40,11 +40,13 @@
 from pyglet.libs.darwin.cocoapy import NSApplicationDidUnhideNotification
 from pyglet.libs.darwin.cocoapy import send_super, get_selector
 from pyglet.libs.darwin.cocoapy import PyObjectEncoding
+from pyglet.libs.darwin.cocoapy import quartz
 from .systemcursor import SystemCursor
 
 NSNotificationCenter = ObjCClass('NSNotificationCenter')
 NSApplication = ObjCClass('NSApplication')
 
+
```
olehermanse: No big deal, but this is just a whitespace change, and doesn't belong in this commit.

PR #22, opened 2019-07-27
Add get_pixel_ratio to BaseWindow

einarf:
Making this a method matching `get_size` and `get_framebuffer_size`.

I think in 2.0 it could be an idea making these into `@property` methods.


olehermanse:
Tested it, works well.

benmoran56:
Thanks @einarf!

PR #24, opened 2019-07-31
Issues with uniform block inspection

caffeinepills:
While most people may not use the gl_ModelViewProjectionMatrix, it does cause the uniform block inspection to crash when trying to split it.

### DIFF
--- a/pyglet/graphics/shader.py
+++ b/pyglet/graphics/shader.py
@@ -326,33 +326,36 @@ class ShaderProgram:
 
         block_uniforms = {}
 
-        for index in range(self._get_num_active(GL_ACTIVE_UNIFORMS)):
-            uniform_name, u_type, u_size = self._query_uniform(index)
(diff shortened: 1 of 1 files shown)


benmoran56:
The general idea is OK, but I think we should instead give the block a name such as "UniformBlock0", etc. Something like:
`block_name = "UniformBlock{}".format(index)`
What do you think? 

We can also remove the block comment there about only supported Named Uniform blocks, because this will work around that as well.

einarf:
Just as a reminder: This must work both in 3.1 and 3.3 core to ensure that it works on OS X and in headless mode.

caffeinepills:
I've changed `_introspect_uniform_blocks ` to no longer rely on the instance name to get the block name. This should also fix issues of gl_ built ins getting caught during the old inspection. Also reduces the requirement of pyglet 2.0 requiring GL 3.3, can now be reduced to 3.1.

PR #30, opened 2019-08-02
Resolve 50+ Python undefined names

cclauss:
[__StandardError__ was removed in Python 3](https://portingguide.readthedocs.io/en/latest/exceptions.html#the-removed-standarderror) in favor of __Exception__ which works as expected in both Python 2 and Python 3.  

__flake8__ results:
* before: https://travis-ci.org/pyglet/pyglet/builds/566994718#L1768 62 undefined names
* after: https://travis-ci.org/pyglet/pyglet/builds/567020258#L1776 8 undefined names

[__wraptypes__](https://pyglet.readthedocs.io/en/pyglet-1.3-maintenance/internal/wraptypes.html) needs work.
There is still three instances of __cmp()__ that was removed in Python 3  https://portingguide.readthedocs.io/en/latest/comparisons.html#the-cmp-function

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,5 @@ addons:
 before_install: pip install --upgrade pip flake8 pytest
 install: python setup.py install
 before_script: flake8 . --count --exclude=./.*,*/future/* --select=E9,F63,F7,F82 --show-source --statistics || true
-script: pytest
+script: pytest --ignore=tests/base --ignore=tests/interactive
 # after_success: coveralls
--- a/pyglet/gl/__init__.py
+++ b/pyglet/gl/__init__.py
@@ -40,7 +40,7 @@ functions.  Functions have identical signatures to their C counterparts.  For
 example::
 
     from pyglet.gl import *
-    
+
     # [...omitted: set up a GL context and framebuffer]
     glBegin(GL_QUADS)
     glVertex3f(0, 0, 0)
@@ -48,7 +48,7 @@ example::
     glVertex3f(0.1, 0.2, 0.3)
     glEnd()
 
-OpenGL is documented in full at the `OpenGL Reference Pages`_.  
+OpenGL is documented in full at the `OpenGL Reference Pages`_.
 
 The `OpenGL Programming Guide`_ is a popular reference manual organised by
 topic.  The free online version documents only OpenGL 1.1.  `Later editions`_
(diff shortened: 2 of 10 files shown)


cclauss:
Sure...  Actually my editor (vscode) is set up to whack trailing whitespace so I will need to turn that off.

olehermanse:
@cclauss You don't have to turn it off, just pay attention to when it does it, and commit that separately. If you always use `git add -p` to changes, you're always aware of what you're staging.

cclauss:
At some point we should whack the whole codebase with https://github.com/psf/black

olehermanse:
@cclauss I don't think I agree with that. Stripping trailing whitespace is good. But I prefer if we agree on a style, and use yapf instead. Even then, running it on the whole codebase makes it more ugly / less readable, IMO.

benmoran56:
I will just add that any potential auto-formatting should be considered for pyglet 2.0 only (shaders branch). That branch has already been extensively changed with removal of Python 2 specific code, futures support, etc., and is still in need of much cleanup. 

The 1.4 branch should get bug fixes and minor changes only, since it's going to be the final Python 2, OpenGL 2.0 release.

cclauss:
Google made a brilliant choice with __gofmt__ to decide across the entire Go language that there should be one and really only one way to format code.  There is no way to _configure_ the formatting choices that gofmt makes -- it is hyper-opinionated so the community around that language spends zero time discussing how code should be formatted.

My sense is that the [Python Software Foundation is trying to do the same this with __black__](https://github.com/psf) -- _any color you want as long at it is black_.  This is in contrast to Google's yapf and autopep8 which can be configured.  My sense is that black is they way to go.

olehermanse:
@cclauss 

> Google made a brilliant choice with gofmt to decide across the entire Go language that there should be one and really only one way to format code. There is no way to configure the formatting choices that gofmt makes -- it is hyper-opinionated so the community around that language spends zero time discussing how code should be formatted.

Interesting that you mention gofmt. I agree that it's nice that it's all formatted the same. But it's also frustrating when they make the "wrong" decisions about how code should be formatted. I like Black's choices a lot more than gofmt, but it kind of highlights why configuration can be nice. With clang-format and yapf I've been able to use my preferred style (which is very close to how Black does it), the same config, across projects in C, python, etc. My YAPF config (and the preferred style in Black), is much more friendly to refactoring, diffs, and readability than gofmt, if I was a golang developer, their style choices would annoy me (a little bit).

With that said, all of this is largely based on opinion, it's not **that important** and since Black looks very close to what my config would be, I don't care that much either way.

PR #32, opened 2019-08-02
Travis CI: Add macOS to the testing

cclauss:
Python 2 on Linux,
Python 3 on Linux,
Python 2 on macOS

pytest --ignore=tests/base --ignore=tests/interactive

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 dist: xenial
 language: python
-python: 3.7
 services: xvfb
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


benmoran56:
Thanks for the additions. As a semi related note, I'll paste some additional comments here: 

Looking at the test suite, the media related tests seem to be failing recently where they didn't before. May be due to a recent change in pytest or mock. I'll have to look into those. The tests in general are a mix of old unittest tests and pytest additions, and could use an overhaul. 

The text related tests seem to be failing because they incorrectly assume that the default system font will be "Arial". This is not the case on many Linux systems, including the Travis CI servers. These tests will need to be modified to work around that incorrect assumption.

Issue #33, opened 2019-08-03
cmp() was removed in Python 3


cclauss:
[__wraptypes__](https://pyglet.readthedocs.io/en/pyglet-1.3-maintenance/internal/wraptypes.html) needs work.  There is still three instances of __cmp()__ that was removed in Python 3
https://docs.python.org/3/howto/sorting.html
https://portingguide.readthedocs.io/en/latest/comparisons.html#the-cmp-function

https://travis-ci.org/pyglet/pyglet/builds/567020258#L1776-L1801

benmoran56:
Definitely needs work!  **wraptypes** is still functional under Python2, and does not need to be run often, so unfortunately it has suffered from bitrot. We definitely need to upgrade this with the end of Python 2 right around the corner. 

As you can see, **wraptypes** is using a vendored legacy version of on David Beazley's PLY project. The PLY project itself has been updated to work under Python 2/3 a while ago, but the API also changed along the way.  Perhaps a good short term plan is to attempt to port the existing codebase to Python 3, if it can be made to work without a ton of work. 

A longer term plan would be to do a rewrite based on [SLY](https://github.com/dabeaz/sly), which seems like a much nicer library to work with.

PR #34, opened 2019-08-03
cmp was removed in Python 3

cclauss:
A partial solution to #33 

```python
#!/usr/bin/env python2

for ab in (["a", "b"],
           ["a", "bb"],
           ["aa", "b"],
           ["aa", "bbb"],
           ["aaa", "bb"],
           ["", "b"]):
    old = ab[:]
    old.sort(lambda a, b: -cmp(len(a), len(b)))
    new = ab[:]
    new.sort(key=len, reverse=True)
    print(ab, old, new)
    assert old == new
```


benmoran56:
By the way, have you yet attempted to run wraptypes under Python 3?

PR #58, opened 2019-09-19
Readme merge

einarf:
An attempt to merge in a lot of the information in the wiki. Also converted the `README` file to markdown because it has much wider support.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,6 @@
 include LICENSE
 include NOTICE
-include README.rst
+include README.md
 include CHANGELOG
 include RELEASE_NOTES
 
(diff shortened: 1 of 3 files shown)


caffeinepills:
The markdown fits really nicely, my only critique would be, that the opening paragraph and features list should be organized a little better. In the first paragraph you're already discussing supported features: 

>It supports windowing, user interface event handling, Joysticks, OpenGL graphics, loading images and videos, and playing sounds and music. *pyglet* works on Windows, OS X and Linux.

But you also have a separate features bullet point list (some of which has the same info), I'd say just combine those. Minor gripe, but features should probably be kept in one concise area.

einarf:
I agree about the features. Was battling with that when gluing things together. Maybe that whole list could need a quick revamp anyway. I'm not extremely versed in what features should be  emphasized.

PR #59, opened 2019-09-19
readthedocs theme

einarf:
Not sure if this is the right thing to do, but it's at least not a too radical change in the doc files and can be reverted later without too much trouble.

Because the custom pyglet theme is currently broken, here's some minimal changes to get it working with `sphinx-rtd-theme`. This way we at least have working search.

Changes:
* Created `requirements.txt` for in `doc/` directory.
* The left  menu is now generated from the `toxtree` directives (by adding `:caption:`)
* Added "External Resources" toctree to including the bottom part the index to the left menu

I'm not entirely sure if these to paragraphs should be re-added somewhere else?
* Removed first paragraph from "Programming Guide" as the `toctree` now creates the title
* Removed first paragraph from "API Guide" as the `toctree` now creates the title

There are also some other minor configuration options that can be explored:
https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html

That section also mentions how the left menu is generated:
> Currently the left menu will build based upon any toctree directives defined in your source files. It outputs 4 levels of depth by default, to allow for quick navigation through topics. If no TOC trees are defined, Sphinx’s default behavior is to use the page headings instead.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -189,7 +189,7 @@ modindex_common_prefix = ['pyglet.']
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'pyglet'
+html_theme = 'sphinx_rtd_theme'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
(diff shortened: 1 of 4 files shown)


einarf:
### Landing with menu
Screenshot 2019-09-20 at 00 26 19

### Working search with highlighting

Screenshot 2019-09-20 at 00 27 45

PR #62, opened 2019-09-21
rst docs

einarf:
Changes:
* All docs files now has `.rst` file extensions. This makes editing docs a lot easier as rst linting and preview plugins works better with `rst` files.
* Added note in `doc/ext` (old custom theme)
* All documents should be linted.
* Standardized all documents to use `lf` endings. They were a mix of `lf` and `crlf` (the rst standard wants `lf`)
* Move "External Resource" toctree above the "Development Guide"
* External projects are now proper toctree nodes (in "External Resources")


einarf:
Should be done now. This turned into something much bigger than I originally planned.
I've looked though the generated docs a couple of times and it seems everything still looks reasonable.

einarf:
.. and editing docs with previewer should now work like a breeze.. 
Screenshot 2019-09-21 at 13 10 25

Issue #79, opened 2019-10-14
Multi inheritance causes issue when variable names collide


Torxed:
```python
import pyglet

class ImageObject():
    def __init__(self, image=None, *args, **kwargs):
        if not 'batch' in kwargs: kwargs['batch'] = pyglet.graphics.Batch()
        self.batch = kwargs['batch']

class test(ImageObject, pyglet.sprite.Sprite):
    def __init__(self, *args, **kwargs):
        if not 'batch' in kwargs: kwargs['batch'] = pyglet.graphics.Batch()
        ImageObject.__init__(self, *args, **kwargs)

t = test()
```

Causes

```
    self.batch = kwargs['batch']
  File "/usr/lib/python3.7/site-packages/pyglet/sprite.py", line 313, in batch
    self._vertex_list.delete()
AttributeError: 'NoneType' object has no attribute 'delete'
```

This will partially be really hard/impossible to fix.

Torxed:
Using this as a placeholder for my own internal discussion on this topic. Since it's not really relevant to the main project as this isn't really how one should be doing it.

PR #80, opened 2019-10-14
Torxed multi inheritance

Torxed:
Fixing issue #79 where `Sprite()` might not yet be instantiated in a parent class in a inheritance chain scenario or a multi inheritances scenario.

The problem appears to be the assumption that when setting batch, the java-alike `setter` is called and it assumes that `.batch` is already set in a previous iteration. This isn't true if `Sprite()` isn't instantiated in the right order.

Moving away from this assumption, and not calling `.migrate()` if the `_vertex_list` is empty, seams to solve the corner case.

**Note**: Tested on two smaller projects, no extensive tests have been made!

### DIFF
--- a/pyglet/sprite.py
+++ b/pyglet/sprite.py
@@ -190,6 +190,7 @@ class Sprite(event.EventDispatcher):
     _scale_y = 1.0
     _visible = True
     _vertex_list = None
+    _texture = None
 
     def __init__(self,
                  img, x=0, y=0,
@@ -310,9 +311,11 @@ class Sprite(event.EventDispatcher):
             self._batch.migrate(self._vertex_list, GL_QUADS, self._group, batch)
             self._batch = batch
         else:
-            self._vertex_list.delete()
+            if self._vertex_list:
+                self._vertex_list.delete()
             self._batch = batch
-            self._create_vertex_list()
+            if self._texture:
+                self._create_vertex_list()
 
     @property
     def group(self):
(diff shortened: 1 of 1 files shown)


benmoran56:
Thanks for the PR. I'm not sure this is a good idea, since i's going down the path of "hardening" the Sprite class against being clobbered. Unfortunately the more safety checks are in there, the slower the class becomes. I'm more in favor of keeping the Sprite class as lean as possible since it's a very performance critical path. 

In the linked issue, clobbering the `self.batch` property would be expected to break. I guess we can't really prevent users from doing things like that without lots of extra code everywhere.

Torxed:
It's a fair concern, and I do agree that the class should be generic and not riddled with performance impacts. I think I gotta benchmark some more and maybe skip the `if` cases, because there are way around those in the users code, but the non-existing values before initialization is an issue, so maybe keep those.

I'll have a look : )

benmoran56:
@Torxed  Some thing have changed since this was opened. Firstly, the master branch now contains the development code for pyglet v2.0.  Secondly, we don't really want to introduce any big changes into 1.4, as it's a stable release. Any changes would be better in 2.0, or even a potential 1.5 release (if someone really wants to do one of those).  

With that in mind, shall we close this for now, and potentially look at the pyglet 2.0 (master branch) sprite class? It's been changed quite a bit already from 1.4. 
Let me know what you think.

Torxed:
I think this is a good idea, I'll re-investigate how much/little there is to change in 2.0 and re-investigate the need for this. I've been running these changes in 1.4 for a while, but there's no point in changing 1.4 in such a way since we're moving away to 2.0 anyway : )

PR #86, opened 2019-10-24
Catch StopIteration at the end of an iterator

gbrookshire:
This change avoids a RuntimeError that arises when initializing a text object. I think this error arises because of a change in the way Python 3.7 handles a StopIteration. I've tested this in Psychopy v3.1.5 running Python 3.7.3.

### DIFF
--- a/pyglet/text/runlist.py
+++ b/pyglet/text/runlist.py
@@ -418,7 +418,11 @@ class ZipRunIterator(AbstractRunIterator):
                 start = min_end
                 for i, iterator in enumerate(iterators):
                     if ends[i] == min_end:
-                        starts[i], ends[i], values[i] = next(iterator)
+                        try:
+                            starts[i], ends[i], values[i] = next(iterator)
+                        except StopIteration:
+                            return
+
         except StopIteration:
             return
 


benmoran56:
Thank you very much for the pull request.  I'll have a look at this soon. 
In the mean time, do you have a simple code example that triggers this? Might be a good idea to make a unit test for it as well.

benmoran56:
Unfortunately that doesn't error out for me. Could you confirm your pyglet version? Better yet, could you share the output of `python -m pyglet.info`?

benmoran56:
Yes, looks like you have 1.3.0 installed, which is quite old. There were already some fixes for Python 3.7 that went in around 1.3.2 or so, but were a bit further up the stack in the text layout classes. 

If you can upgrade your pyglet to the latest 1.4.x release, that should fix it.

gbrookshire:
Good to know -- thanks!

benmoran56:
Looks like the requirements in Psychopy have the pyglet version requirements listed as:
`pyglet>=1.3,<1.4`
Maybe that just installs 1.3 by default? I'm not positive how that works, as I've never used that pattern before. In any case,   if you have a chance to update pyglet to 1.4.x, we can confirm if that fixes your issue.

gbrookshire:
I've uninstalled and reinstalled psychopy and pyglet using Anaconda (a
number of times). When installing everything with the default system python
(3.7) instead of specifying 3.6, it installs pyglet 1.4.5 and everything
seems to work correctly.

benmoran56:
Thank you for confirming! I suppose this means that this pull request isn't necessary, but I appreciate the effort.

Thanks again!

gbrookshire:
Sure thing -- thanks for helping me figure it out!

Issue #87, opened 2019-10-26
AttributeError's are suppressed in 1.4.6+


Torxed:
To demonstrate the issue, below is an [executable code](https://gist.github.com/Torxed/c82ff9ca710cf0cb108497f215d53ab4) snippet.

The problem appears to be (in `event.py`):

```python
        except AttributeError:
            pass
```

Commenting it out produces:
```
  File "C:\Python36\lib\site-packages\pyglet\event.py", line 423, in dispatch_event
    if getattr(self, event_type)(*args):
AttributeError: 'UnformattedDocument' object has no attribute 'on_insert_text'
```

Usually, I'm used to getting out the different errors and pyglet crashing, which is more of a desired behavior unless it's internal stuff ofcourse, like some of my more experimental codes I'm used to getting:

```
Traceback (most recent call last):
  File "test.py", line 54, in <module>
    W = window()
  File "test.py", line 50, in __init__
    self.add_sprite('smilyface', smilyface(debug=True, x=100, y=100, width=20, height=20, alpha=0))
  File "test.py", line 16, in __init__
    super(smilyface, self).__init__(*args, **kwargs)
  File "/home/anton/github/pyglet2d/pyglet_gui.py", line 222, in __init__
    ImageObject.__init__(self, texture, *args, **kwargs)
  File "/home/anton/github/pyglet2d/pyglet_gui.py", line 167, in __init__
    self.batch = kwargs['batch']
  File "/usr/lib/python3.7/site-packages/pyglet/sprite.py", line 322, in batch
    self._create_vertex_list()
  File "/usr/lib/python3.7/site-packages/pyglet/sprite.py", line 401, in _create_vertex_list
    4, GL_QUADS, self._group, vertex_format, 'c4B', ('t3f', self._texture.tex_coords))
AttributeError: 'NoneType' object has no attribute 'tex_coords'
```

And I quite like it, but Pyglet still handles all the internal errors pretty well.
Below is the code example.

```python
import pyglet
from math import *

key = pyglet.window.key
tile_width_half = 84/2
tile_height_half = 51/2

class Tile(pyglet.sprite.Sprite):
	def __init__(self, width=84, height=51, image='path.png', *args, **kwargs):
		self.texture = pyglet.image.load(image)
		pyglet.sprite.Sprite.__init__(self, self.texture, **kwargs)

class mathtile():
	def __init__(self, x=42, y=25, width=84, height=51):
		self.max_x = x + (width/2)
[...]

PR #88, opened 2019-10-27
Torxed issue 87

Torxed:
Modified comment https://github.com/pyglet/pyglet/issues/87#issuecomment-546611608 and made the syntax more resembled to the neighboring exceptions.

Seams to be back to the expected result without penalty result.<br>
For instance:

```
[anton@Archinstall pyglet2d]$ python test.py 
[DEBUG] Activating button's object: <__main__.testButton object at 0x7f1ea4261a50>
  File "test.py", line 74, in <module>
    W.run()
  File "/home/anton/github/pyglet2d/pyglet_gui.py", line 700, in run
    event = self.dispatch_events()
  File "/usr/lib/python3.7/site-packages/pyglet/window/xlib/__init__.py", line 870, in dispatch_events
    self.dispatch_platform_event_view(e)
  File "/usr/lib/python3.7/site-packages/pyglet/window/xlib/__init__.py", line 900, in dispatch_platform_event_view
    event_handler(e)
  File "/usr/lib/python3.7/site-packages/pyglet/window/xlib/__init__.py", line 1205, in _event_button
    self.dispatch_event('on_mouse_press', x, y, button, modifiers)
  File "/usr/lib/python3.7/site-packages/pyglet/window/__init__.py", line 1330, in dispatch_event
    if EventDispatcher.dispatch_event(self, *args) != False:
  File "/usr/lib/python3.7/site-packages/pyglet/event.py", line 428, in dispatch_event
    self._raise_dispatch_exception(event_type, args, getattr(self, event_type), e)
  File "/usr/lib/python3.7/site-packages/pyglet/event.py", line 485, in _raise_dispatch_exception
    raise exception
  File "/usr/lib/python3.7/site-packages/pyglet/event.py", line 425, in dispatch_event
    func(*args)
  File "/home/anton/github/pyglet2d/pyglet_gui.py", line 611, in on_mouse_press
    sprite_obj.mouse_down(x, y, button)
AttributeError: 'testButton' object has no attribute 'mouse_down'
```


pyglet/event.py
```diff
@@ -420,10 +420,13 @@ def dispatch_event(self, event_type, *args):
...
-            if getattr(self, event_type)(*args):
+            func = getattr(self, event_type, None)
+            if func:
+                func(*args)
                 return EVENT_HANDLED
-        except AttributeError:
-            pass
+        except AttributeError as exception:
+            self._raise_dispatch_exception(event_type, args, getattr(self, event_type), exception)
+            return EVENT_UNHANDLED
```
Torxed: This guy is probably overkill, since the later `return EVENT_HANDLED` will trigger.

benmoran56:
I think in this case we don't want to specifically catch `AttributeError` anymore. The only reason that was in there as `pass` in the first place, was because the line
**`if getattr(self, event_type)(*args):`** would raise an `AttributeError` if the instance did not have a particular handler. By changing the pattern to:
```python
func = getattr(self, event_type, None)
if func:
    func(*args)
```
we have prevented the `AttributeError` from ever being raised by the event dispatcher internally, and therefor there is no more need to specifically pass it. We can now assume that any exceptions raised must come from the function call itself, so we want to catch *everything* and send it to that `self._raise_dispatch_exception` method. 

Does that make sense? It seems like the best solution would be to catch everything, such as:
`except Exception as exception:` or even `except BaseException as exception:`.

Torxed:
Correct, `getattr(x, y, None)` will safely check if the attribute exists, and if not return `None` as per intention. So removing the `except AttributeError` clause and assuming every exception is from the user-platform end.

pyglet/event.py
```diff
@@ -421,9 +421,6 @@ def dispatch_event(self, event_type, *args):
             if func:
                 func(*args)
                 return EVENT_HANDLED
-        except AttributeError as exception:
-            self._raise_dispatch_exception(event_type, args, getattr(self, event_type), exception)
-            return EVENT_UNHANDLED
         except TypeError as exception:
```
benmoran56: Instead of `TypeError`, we want to catch everything here. `Exception` or `BaseException` is a better choice. If we only catch `TypeError`, other types of exceptions will crash on line 422, and not be passed to `self._raise_dispatch_exception`.

Torxed: Good thinking, as pretty obvious. I'm still not entirely familiar with the whole function chain. So I'm unaware of what not catching the errors will cause other than a hard crash. Modified the code further.

benmoran56:
It looks like the `else` clause is causing trouble. Could you simplify it like this?
```python
        # Check instance for an event handler
        try:
            func = getattr(self, event_type, None)
            if func:
                func(*args)
                return EVENT_HANDLED
        except BaseException as exception:
            self._raise_dispatch_exception(event_type, args, getattr(self, event_type), exception)
            return EVENT_UNHANDLED

        return False
```

After that change, the unit tests will pass, and we can merge it in. 
Thanks!

Torxed:
As per request, the change has been made. Curious as to what sort of issue it was causing, didn't notice anything in my basic tests. Care to share for learning purposes *(only if you have a second over to spare, otherwise, not important)* : )

Torxed:
Added a review request as well to be on the safe side, as to the logic of `if getattr()` vs `if exec()`.

benmoran56:
Great, thanks!

If you check out the little Travis-CI link, you should be able to browse to the current and past builds, including the failed ones with error messages.  

Basically there are three potential ways this function can return. True, False, and None, depending on successful, with exception, or nothing executed. I think previously the else statement was taking us down the wrong path.

PR #89, opened 2019-10-30
Add DWM support for Windows Vista+

caffeinepills:
This adds support for DwmFlush and DwmIsCompositionEnabled for Windows Vista+. DWM has become integrated into the latest Windows platforms and affects the Vsync. 

This helps relieve some jitter or microstutters that some Windows users experience if you are in Windowed mode and with Vsync on.  I believe this primarily affects us because we utilize double buffering. This may need testing with no double buffering configurations.

It currently supports XP by testing the Windows Version and not loading the DLL. Can be removed if XP is no longer supported. However the docs do say XP and above currently.

### DIFF
--- a/pyglet/libs/win32/__init__.py
+++ b/pyglet/libs/win32/__init__.py
@@ -92,6 +92,7 @@ else:
 _gdi32 = DebugLibrary(windll.gdi32)
 _kernel32 = DebugLibrary(windll.kernel32)
 _user32 = DebugLibrary(windll.user32)
+_dwmapi = DebugLibrary(windll.dwmapi)
 
 # _gdi32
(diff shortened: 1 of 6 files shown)


pyglet/libs/win32/__init__.py
```diff
@@ -92,6 +92,7 @@ def f(*args):
 _gdi32 = DebugLibrary(windll.gdi32)
 _kernel32 = DebugLibrary(windll.kernel32)
 _user32 = DebugLibrary(windll.user32)
+_dwmapi = DebugLibrary(windll.dwmapi)
```
benmoran56: :+1:

PR #92, opened 2019-11-04
ENH: Add support for conda-installed FFmpeg on macOS

hoechenberger:
On macOS, pyglet wasn't able to detect the FFmpeg libraries installed into a `conda` environment, as `conda` doesn't add them to `LD_LIBRARY_PATH`. Instead, one has to consider the `CONDA_PREFIX` environment variable, which points to the currectly active `conda` environment.

### DIFF
--- a/pyglet/lib.py
+++ b/pyglet/lib.py
@@ -222,6 +222,11 @@ class MachOLibraryLoader(LibraryLoader):
                 sys.frozen is True and pyglet.compat_platform == 'darwin'):
             search_path.append(os.path.join(sys._MEIPASS, libname))
 
+        # conda support
+        if os.environ.get('CONDA_PREFIX', False):
+            search_path.append(os.path.join(os.environ['CONDA_PREFIX'], 'lib',
+                                            libname))
+
         if '/' in path:
             search_path.extend([os.path.join(p, libname) for p in self.dyld_library_path])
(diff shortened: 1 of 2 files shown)


pyglet/lib.py
```diff
@@ -222,6 +222,11 @@ def find_library(self, path):
                 sys.frozen is True and pyglet.compat_platform == 'darwin'):
             search_path.append(os.path.join(sys._MEIPASS, libname))
 
+        # conda support
+        if os.environ.get('CONDA_PREFIX', False):
+            search_path.append(os.path.join(os.environ['CONDA_PREFIX'], 'lib',
```
benmoran56: If you prefer, we allow a line length of 100 columns in pyglet.

hoechenberger: Thanks, I've changed this in e0ca014

benmoran56:
I thought I might try to change the base from `master` to `pyglet-1.4-maintenance` on my end, but that doesn't work so well. No worries, I'll merge this in manually later today. 

Thanks again!

hoechenberger:
Thank you very much, @benmoran56!

benmoran56:
Merged into master for v2.0, as well as 1.4-maintance branch for the upcoming v1.4.7 point release. Thanks again!

PR #97, opened 2019-11-16
Remove findall, it's slow. Speed up using list comprehension.

caffeinepills:
Using findall is pretty slow and not necessary when we have list comprehension when all we are doing is looking for lines. Seems to speed up atleast 2x on larger images. Already did the normal case a while ago, however I've noticed `get_data` on` ImageRegion` was still slow, so finished removing all of the findalls.

One thing to note that in `get_data` it uses `self._current_data = b''.join(rows)` which is inconsistent with others that use the `compat.asbytes` such as `data = asbytes('').join(rows)` not sure which is correct, as I think they do the same thing, so I'm just leaving it.

### DIFF
--- a/pyglet/image/__init__.py
+++ b/pyglet/image/__init__.py
@@ -1054,8 +1054,8 @@ class ImageData(AbstractImage):
             packed_pitch = self.width * len(current_format)
             if abs(self._current_pitch) != packed_pitch:
                 # Pitch is wider than pixel data, need to go row-by-row.
-                rows = re.findall(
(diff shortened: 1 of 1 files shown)


benmoran56:
Thanks for the PR!  Definitely agreed that list comprehensions are the way to go here. 

As for the "asbytes", it looks like `asbytes('')` and `b''` are going to produce an identical result in both Py2 and 3. This probably snuck in during the initial Python 3 upgrade, out of habit.

PR #102, opened 2019-11-29
Fixed  a bug with Sprite._update_position()

jmiller89:
This method no longer raises an Exception when the sprite is not visible.
I found this when trying out pyglet 2.0.dev0 on a game I'm currently working on. I really like this library, and am excited for 2.0!


benmoran56:
Thanks for the pull request!

PR #108, opened 2019-12-21
darwin: Make context creation more flexible

einarf:
On OS X the opengl major/minor version is really only a lose suggestion because you'll always get the highest core context available no matter what gl version you request. This is in line with how most (if not all) window libraries are handling it.

https://github.com/glfw/glfw/blob/fa602692455d87e11c9ff5a5fb0681ca6403772a/src/nsgl_context.m#L234-L246


einarf:
I can make PR for 1.4 as well. Just waiting for judgement!

benmoran56:
Looks good to me.

PR #110, opened 2019-12-23
darwin: Make context creation more flexible (#108)

einarf:
Cherry-pick darwin gl version fix from master

### DIFF
--- a/pyglet/gl/cocoa.py
+++ b/pyglet/gl/cocoa.py
@@ -189,7 +189,7 @@ class CocoaConfig(Config):
             # check for opengl profile
             # This requires OS-X Lion (Darwin 11) or higher
             version = (
-                getattr(self, 'major_version', None),
+                getattr(self, 'major_version', None) or 2,
                 getattr(self, 'minor_version', None)
                 )
             # tell os-x we want to request a profile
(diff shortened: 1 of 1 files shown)


einarf:
hmm. I guess this needs more attention when version tuple contains `None` values.

benmoran56:
Not to be pedantic, but how about just simplifying it further, but changing the default values to ints when major/minor versions are not provided? Then we can just do simple tuple comparison, and make it easier to read.  Something like this for the defaults *should* cover the bases.  Defaulting to 2.2 should also give a 3.2 context, if someone were to only pass the "major_version" as 3. Does that make sense to you?

```py
            version = (
                getattr(self, 'major_version', 2),
                getattr(self, 'minor_version', 2)
                )
            # tell os-x we want to request a profile
            attrs.append(cocoapy.NSOpenGLPFAOpenGLProfile)

            # check if we're wanting core or legacy
            # Mavericks (Darwin 13) and up are capable of the Core 4.1 profile,
            # while Lion and up are only capable of Core 3.2
            if version >= (4, 1) and _os_x_version >= os_x_release['mavericks']:
                attrs.append(int(cocoapy.NSOpenGLProfileVersion4_1Core))
            elif version >= (3, 2):
                attrs.append(int(cocoapy.NSOpenGLProfileVersion3_2Core))
            else:
                attrs.append(int(cocoapy.NSOpenGLProfileVersionLegacy))
```

einarf:
I guess squash this one if you merge

PR #116, opened 2020-01-02
Force Window Size

Capital-EX:
This change forces the window size to be `(800, 600)` which is what this tutorial expects. Without this code, it's possible for a different window size to be used.  In my case, Pyglet was creating a window of (640, 480).


benmoran56:
Thank you very much for the PR.

PR #118, opened 2020-01-05
Updated data item parameters to current names.

DGriffin91:
I was trying out this example and had to change the data item parameters from vertices2i and colors4Bn to v2i and c4B for the example to run. I'm assuming vertices2i and colors4Bn are old data item parameter names.

### DIFF
--- a/examples/text_input.py
+++ b/examples/text_input.py
@@ -17,8 +17,8 @@ class Rectangle(object):
     def __init__(self, x1, y1, x2, y2, batch):
         self.vertex_list = batch.add_indexed(4, pyglet.gl.GL_TRIANGLES, None,
                                              [0, 1, 2, 0, 2, 3],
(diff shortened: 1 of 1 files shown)


benmoran56:
Hi @DGriffin91, thank you for the pull request.  The master branch is targeting v2.0, which does use the new full length names. Previously pyglet had a half dozen or so short names (v2f, c3B, etc.) because there were only that many default variable attributes available in legacy OpenGL. With pyglet v2.0+ we're targeting OpenGL 3.1+, which lets you use any names you like in the shaders. That's why the longer names are required - to match the new default shaders in pyglet v2.0. 

Sorry if this is a bit confusing, but if you're using pyglet from pip (v1.4.x), you need to check out the "pyglet-1.4-maintenance" branch to get the matching examples.

DGriffin91:
@benmoran56 Makes perfect sense. Thank you for your explanation! Thanks for all the work you do on this project!

PR #126, opened 2020-01-19
Add customizable min_zoom for Camera

t-o-n-y-p:
Current Camera implementation does not allow to zoom out from the default level to .5, .25, etc.
The solution is to make min_zoom customizable.

### DIFF
--- a/examples/camera.py
+++ b/examples/camera.py
@@ -64,7 +64,7 @@ class Camera:
         self.min_zoom = min_zoom
         self.offset_x = 0
         self.offset_y = 0
-        self._zoom = self.min_zoom
+        self._zoom = max(min(1, self.max_zoom), self.min_zoom)
 
     @property
     def zoom(self):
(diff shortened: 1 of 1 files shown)


benmoran56:
Thanks Tony! 

If I could make one suggestion, I would set the default zoom to = 1, instead of min_zoom. 
I think that would be the expected behavior for most users.

t-o-n-y-p:
What if 1 lies outside of [min_zoom, max_zoom] ? What would you expect?

benmoran56:
If we want to cover that case, we could use the same logic as the `zoom.setter`. This should bias towards "1" in case both values are higher or lower than 1: 
```self._zoom = max(min(1, max_zoom), min_zoom)```

Might as well also throw an assertion at the top of `__init__` for sanity. Something like:
```assert min_zoom < max_zoom, "Minimum zoom must be less than maximum zoom"```

t-o-n-y-p:
Maybe just silently swap them for last case?

benmoran56:
It's usually not good to silently do things, so I usually prefer an assertion. This will break execution as soon as the class is instanced and let them know what the issue is.

Changes look good though! Should we merge this in?

t-o-n-y-p:
Great! Now we are good to go.

benmoran56:
Thanks Tony!

Issue #128, opened 2020-01-22
FFMpeg __del__ isn't working right


pvcraven:
In pyglet/media/codecs/ffmpeg.py there is a class FFmpegSource. In the `__del__` method, it assumes attributes like self._packet have been created. This is not always the case. If an exception happens while loading the file, the attribute is not created. Then when the deletion method is run, we get another exception.

benmoran56:
Addressed by recent pull request.
Thanks!

PR #129, opened 2020-01-22
Fix for issue #128, __del__ doesn't work if exception thrown in __init__

pvcraven:


benmoran56:
Thanks Paul!  Looks better now.

PR #133, opened 2020-01-23
Create readthedocs config

einarf:
This will stabelize builds in general instead of always falling back to rtd's default config that is constantly changing.

Most of my projects not having this config file broke yesterday.


einarf:
This also enables generating pdf docs.

benmoran56:
Thanks for that!

Issue #105, opened 2019-12-12
AttributeError incorrectly catched by EventDispatcher.dispatch_event


marcvigoanglada:
The following is a [simple example](https://github.com/pyglet/pyglet/files/3955261/pyglethello.py.txt) of the error we detected. 

We expect calling HelloWorldWindow.on_draw to raise an AttributeError exception. Instead, no error is raised but in the example the label is not shown (in fact, the handler is executed only until the wrong attribute is accessed). 

We traced the error and concluded that EventDispatcher.dispatch_event method incorrectly catches the AttributeError. This method should only catch the error raised by `gettatr.`

We propose to rewrite lines 418-427 in file pyglet/pyglet/event.py:

        # Check instance for an event handler
        try:
            if getattr(self, event_type)(*args):
                return EVENT_HANDLED
        except AttributeError:
            pass
        except TypeError as exception:
            self._raise_dispatch_exception(event_type, args, getattr(self, event_type), exception)
        else:
            invoked = True

like this:

        # Check instance for an event handler
        try:
            handler = getattr(self, event_type)
            try:
                if handler(*args):
                    return EVENT_HANDLED
            except TypeError as exception:
                self._raise_dispatch_exception(event_type, args, getattr(self, event_type), exception)
            else:
                invoked = True
        except AttributeError:
            pass

caffeinepills:
Related to: 
https://github.com/pyglet/pyglet/issues/87

Looks like it was reverted back to the original way for 1.4.8:
`Revert Event Dispatcher changes from v1.4.7 due to possible bug.`

I'm not sure what the bug was but maybe this is a better way of catching the error?

benmoran56:
I reverted this because it was causing wierd issues with some code. I think there might be a hard to find race condition with class instantiation. 

The main issue is that we're relying on an AttributeError to determine if a class has a handler method or not. Maybe we need to rethink how this code works, and make a more robust solution.

pvcraven:
I still get this issue in 1.4.9. Or one like it. The pass below eats a lot of errors:

```
        # Check instance for an event handler
        try:
            if getattr(self, event_type)(*args):
                return EVENT_HANDLED
        except AttributeError:
            pass
        except TypeError as exception:
            self._raise_dispatch_exception(event_type, args, getattr(self, event_type), exception)
        else:
            invoked = True```

benmoran56:
[...]

PR #147, opened 2020-02-12
fixes issue #105 by removing try: except completely

Nearoo:
This solution fixes incorrect `AttributeError` catches by removing the corresponding `try: except` block completely. 

I assume the reason this block was added in the first place was to distinguish between a method not existing and a method returning a falsy value without invoking `hasattr`, for performance reasons.

My solution verifies if the method was invoked by first creating a unique object (`tuple`) that is then returned as a default argument from `getattr`. To check if a method was invoked, that object is then compared to the return value by using the `is` comparator.

### DIFF
--- a/pyglet/event.py
+++ b/pyglet/event.py
@@ -417,23 +417,19 @@ class EventDispatcher:
             except TypeError as exception:
                 self._raise_dispatch_exception(event_type, args, handler, exception)
 
-        # Check instance for an event handler
+        not_called = () # arbitrary unique, falsy value
         try:
-            if getattr(self, event_type)(*args):
-                return EVENT_HANDLED
-        except AttributeError as e:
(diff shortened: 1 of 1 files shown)


Nearoo:
I just noticed that LGTM uncovered an error in my implementation that is not caught by tests yet - the case where event handlers don't handle event by returning pyglet.event.EVENT_UNHANDLED.

Will add tests that catch it along with a fix later.

benmoran56:
Thanks for the PR! I like your solution it's similar to what they do in the standard library lru_cache, with using a "`sentinal`" object.

Unfortunately there is still a race condition that we hit when not hijacking AttributeError. It often happens in slightly complex projects. One way to reproduce this consistently is to try the `examples/text_input.py` example in the `master` branch. Paul and I were talking about this on Discord a bit.

Issue #158, opened 2020-03-08
Pyglet plays audio files for 2 seconds then stops???


davidlondonuk:
Hi,

I'm using Ubuntu 18.04 with python 3.6 and ffmpeg 4.2.

I am using the code below for an audio player project. Pyglet via ffmpeg will play any file of mp3, m4a, ogg or wav for approx 2 seconds then stops. The same audio files play without a problem in well known players like Rhythmbox or Google play music etc. I just can't figure out why pyglet accepts the file but ffmpeg plays 2 seconds then cuts out?

class Player:
    def __init__(self):
        self.player = pyglet.media.Player()
        self.player.volume = 0.6

    def play_media(self, audio_file):
        self.reset_player()
        self.player = pyglet.media.Player()
        self.source = pyglet.media.load(audio_file)
        self.player.queue(self.source)
        self.player.play()

    def reset_player(self):
        self.player.pause()
        self.player.delete()

My apologies if the markdown has mangled the code, the python code runs ok apart from the sound cutting out problem.

Thanks for any help.

benmoran56:
Hi David, 

You'll need a running event loop or the audio buffers won't get refilled. In a typical pyglet application, this is `pyglet.app.run()` at the end of your project. 

If you're using some other event loop, then you'll need to periodically tick pyglet's clock: `pyglet.clock.tick()`.

jonathanjfshaw:
Thanks @benmoran56 . Here's a PR documenting this: #176

benmoran56:
Thanks again for the PR. I'll close this for now.

PR #176, opened 2020-03-28
Document need to tick clock when using pyglet audio outside of pyglet…

jonathanjfshaw:
Addresses #158


benmoran56:
Thanks Jonathan!

PR #183, opened 2020-04-16
Fix crash initializing WMFDecoder

### LINKED ISSUE #174: debugging pyglet on windows, I get a warning/exception related to COM/threading
Hey, I'm using pyglet as a media player for a program I'm writing, and I'm using PyQt5 for the GUI. I was having problems with the program getting stuck when running, and when debugging I got the following error:
```
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "C:\Users\ofer\miniconda3\lib\site-packages\pyglet\media\__init__.py", line 143, in <module>
    add_default_media_codecs()
  File "C:\Users\ofer\miniconda3\lib\site-packages\pyglet\media\codecs\__init__.py", line 175, in add_default_media_codecs
    add_decoders(wmf)
  File "C:\Users\ofer\miniconda3\lib\site-packages\pyglet\media\codecs\__init__.py", line 128, in add_decoders
    for decoder in module.get_decoders():
  File "C:\Users\ofer\miniconda3\lib\site-packages\pyglet\media\codecs\wmf.py", line 887, in get_decoders
    return [WMFDecoder()]
  File "C:\Users\ofer\miniconda3\lib\site-packages\pyglet\media\codecs\wmf.py", line 833, in __init__
    raise(err)
  File "C:\Users\ofer\miniconda3\lib\site-packages\pyglet\media\codecs\wmf.py", line 831, in __init__
    ole32.CoInitializeEx(None, COINIT_MULTITHREADED)
  File "_ctypes/callproc.c", line 935, in GetResult
OSError: [WinError -2147417850] Cannot change thread mode after it is set
Exception ignored in: <function WMFDecoder.__del__ at 0x000001BFF08F9318>
```
If I comment out the guilty line 
```
wmf.py:831 ole32.CoInitializeEx(None, COINIT_MULTITHREADED)
```
the program works fine in debug and regular mode.

The error is reproduced just by running this line in debug
```
import pyglet.media
```

I'm running on Windows 10 with python 3.7.6 and pyglet 1.5.2.

nzjrs:
previously this CoinitializeEx call crashes some applications
at startup, but any failures caused a cascading error at shutdown.

catch the first error better and prevent the crash at shutdown.

fixes #174

### DIFF
--- a/pyglet/media/codecs/wmf.py
+++ b/pyglet/media/codecs/wmf.py
@@ -825,10 +825,13 @@ class WMFSource(Source):
 
 class WMFDecoder(MediaDecoder):
     def __init__(self):
-        # Coinitialize supposed to be called for COMs?
-        ole32.CoInitializeEx(None, COINIT_MULTITHREADED)
+
+        self.ole32 = None
+        self.MFShutdown = None
 
         try:
+            # Coinitialize supposed to be called for COMs?
+            ole32.CoInitializeEx(None, COINIT_MULTITHREADED)
             MFStartup(MF_VERSION, 0)
         except OSError as err:
             raise ImportError('WMF could not startup:', err.strerror)
@@ -873,8 +876,10 @@ class WMFDecoder(MediaDecoder):
             return StaticSource(WMFSource(filename, file))
 
     def __del__(self):
-        self.MFShutdown()
-        self.ole32.CoUninitialize()
+        if self.MFShutdown is not None:
+            self.MFShutdown()
+        if self.ole32 is not None:
+            self.ole32.CoUninitialize()
 
 
 def get_decoders():


caffeinepills:
This code will actually cause the entirety of WMF to not work since it's going to error before MFstartup is called. Better off doing a separate try call for the Coinitializeex, since it's possible it may still work with a different thread mode.

benmoran56:
@caffeinepills @nzjrs  Would it be better to just warn and move on if the first call fails, and then try 
catching on the Coinitializeex? Assuming the MFStartup call will also throw an `OSError`, which it should, how about something like this? 

```py
        try:
            # Coinitialize supposed to be called for COMs?
            ole32.CoInitializeEx(None, COINIT_MULTITHREADED)
        except OSError as err:
            warnings.warn("WMFDecoder: Unable to initialize threading mode.")

        try:
            MFStartup(MF_VERSION, 0)
        except OSError as err:
            raise ImportError('WMF could not startup:', err.strerror)
```

caffeinepills:
@benmoran56 Yeah I think that would be the best solution.

PR #189, opened 2020-04-28
Adds support for ImageMouseCursor to use hardware acceleration

caffeinepills:
Current implementation relies on blitting an image to the screen to display an image cursor. Since you're at the mercy of the rendering pipeline, if it's slowed, or interrupted in anyway, your cursor will stop moving or move slowly. Also depending on how high end your mouse is, the movement can appear to be sluggish since it's possible it's polling faster than 60 FPS in the OS. 

This is changed by adding an acceleration keyword to the ImageMouseCursor. Usage is the same:
```
image = pyglet.image.load("cursor.png")
cursor = pyglet.window.ImageMouseCursor(image, 12, 12)
```
Enabled by default but can be disabled by: `cursor = pyglet.window.ImageMouseCursor(image, 12, 12, acceleration=False)` Debatable on whether it should be default or not.

One possible issue is that, once the icon is created, it would need to be re-created if you change x/y hotspots, or even the image of the cursor. However, I am not sure if the intention was to be able to change everything after cursor creation or to have separate cursor instances for differences. I can make these changes but not sure which ways are intended usage.

By default it turns the drawing attribute off so you do not get two cursors. However, this implementation also allows you to use drawable and acceleration together if you subclass and overwrite the draw function (if you choose).
For instance you may want something drawing with the accelerated cursor, in my usage, I have a label drawing and following the cursor to display information. The label only draws depending on which cursor I am using.

Currently Windows only but thought I'd open this for discussion atleast and get the implementation going since I've been sitting on it for a while.

### DIFF
--- a/pyglet/window/__init__.py
+++ b/pyglet/window/__init__.py
@@ -172,6 +172,7 @@ class MouseCursor:
     #: Indicates if the cursor is drawn using OpenGL.  This is True
     #: for all mouse cursors except system cursors.
     drawable = True
+    acceleration = False
 
     def draw(self, x, y):
         """Abstract render method.
@@ -194,6 +195,7 @@ class MouseCursor:
 class DefaultMouseCursor(MouseCursor):
     """The default mouse cursor #sed by the operating system."""
     drawable = False
+    acceleration = True
 
 
 class ImageMouseCursor(MouseCursor):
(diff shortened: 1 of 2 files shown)


benmoran56:
Thanks for opening the pull request, and not letting this work get lost. 
I have a proof of concept mocked up for X on Linux, and it's working well except for the fact that it only supports two color pointers. Higher color depth images are sampled down to black and white. 
For that reason, and also because we don't yet have a working OSX implementation, it might be good to default to `accelleration=False`, at least for 1.5.x. That way, someone using standard image cursors and updating pyglet won't suddenly get changed behavior (at least on Linux). What do you think?  For pyglet 2.0 I'd lean towards defaulting it on, and just documenting the limitation. 

Allowing both hardware and software cursors to draw sounds like it would make a nice example of a custom subclass. Not really common enough to be included maybe, but a nice technique that can be done.

Issue #193, opened 2020-05-11
on_mouse_scroll returns 0 for small scrolls on macOS


eterevsky:
On macOS for small (one-click) scrolls with a mouse or with the trackpad `on_mouse_scroll` event receives 0 as `scroll_y`. Example code:

```python
import pyglet
window = pyglet.window.Window()

@window.event
def on_mouse_scroll(x, y, scroll_x, scroll_y):
    print('on_mouse_scroll', x, y, scroll_x, scroll_y)

pyglet.app.run()
```

When trying to scroll slowly inside the window, both `scroll_x` and `scroll_y` are 0. For faster scrolls some non-zero values are passed.

The documentation states that `scroll_x` and `scroll_y` parameters are supposed to be integer, so it's possible that the system provides a non-zero floating point value, which is then rounded to an integer.

Machine: Macbook Pro + internal trackpad or external Razer mouse
System: macOS Catalina 10.15.4
Python 3.8
Pyglet 1.5.5

benmoran56:
I had a look at this, and it's clamping to the nearest int and losing the fractional values. 
At the top of `pyglet/window/cocoa/pyglet_view.py` you can see the following function:
```py
def getMouseDelta(nsevent):
    dx = nsevent.deltaX()
    dy = -nsevent.deltaY()
    return int(round(dx)), int(round(dy))
```
In the past, it made sense to return ints, but I don't see any reason for it now. Especially since pyglet 1.5+ is Python 3 only, there are no more int/float conflicts to worry about. 

If possible, could you try modifying that function and seeing if it does the trick?

benmoran56:
Just to add, it turns out we've been returning floats on Windows already, so it should be fine to switch to floats.

benmoran56:
This is merged in for the next release, so I'll close this for now. Thanks again!

PR #194, opened 2020-05-12
Fix window and mouse issues on macOS

### LINKED ISSUE #5: Window buttons on mac cause python/pyglet to crash
**Description:** The window buttons (close, minimize, full screen) cause python to crash in various ways. (Segmentation fault, Recursion Error, Abort Trap, Illegal Instruction).

**Pyglet version(s):** 1.4.1

**Python version(s):** 3.7

**OS:** OS X 10.14.5 (Seems to only affect newer Macbooks)

**Related tickets:** [bitbucket #199](https://bitbucket.org/pyglet/pyglet/issues/199/attempting-to-resize-or-close-pyglet) [bitbucket #229](https://bitbucket.org/pyglet/pyglet/issues/229/closing-window-causes-a-crash-on-mac-os)

**Minimal example:**

```python3
import pyglet

pyglet.window.Window(resizable=True)

pyglet.app.run()
```

**Output 1 (close):**

```
$ python3 example.py 
  File "_ctypes/callbacks.c", line 232, in 'calling callback function'
  File "/usr/local/lib/python3.7/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 1120, in objc_method
    args = convert_method_arguments(encoding, args)
  File "/usr/local/lib/python3.7/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 1000, in convert_method_arguments
    new_args.append(ObjCInstance(a))
  File "/usr/local/lib/python3.7/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 921, in __new__
    if not isinstance(object_ptr, c_void_p):
RecursionError: maximum recursion depth exceeded while calling a Python object
Segmentation fault: 11
$ 
```

**Output 2 (minimize):**

```
$ python3 example.py 
  File "_ctypes/callbacks.c", line 232, in 'calling callback function'
  File "/usr/local/lib/python3.7/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 1120, in objc_method
    args = convert_method_arguments(encoding, args)
  File "/usr/local/lib/python3.7/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 1000, in convert_method_arguments
    new_args.append(ObjCInstance(a))
  File "/usr/local/lib/python3.7/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 921, in __new__
    if not isinstance(object_ptr, c_void_p):
RecursionError: maximum recursion depth exceeded while calling a Python object
Illegal instruction: 4
$ 
```

eterevsky:
Fixes #5 and #193 on my system.

The change to rounding the mouse delta changes the API, so docs and other system implementations should probably be updated.

### DIFF
--- a/pyglet/libs/darwin/cocoapy/runtime.py
+++ b/pyglet/libs/darwin/cocoapy/runtime.py
@@ -492,7 +492,9 @@ def send_super(receiver, selName, *args, preventSuperclassRecursion=False, **kwa
         receiver = receiver._as_parameter_
     superclass = get_superclass_of_object(receiver)
     if preventSuperclassRecursion:
-        superclass = c_void_p(objc.class_getSuperclass(superclass))
+        supersuperclass = c_void_p(objc.class_getSuperclass(superclass))
+        if supersuperclass.value is not None:
+            superclass = supersuperclass
     super_struct = OBJC_SUPER(receiver, superclass)
     selector = get_selector(selName)
     restype = kwargs.get('restype', c_void_p)
--- a/pyglet/libs/darwin/cocoapy/runtime.py
+++ b/pyglet/libs/darwin/cocoapy/runtime.py
@@ -1045,7 +1045,7 @@ def convert_method_arguments(encoding, args):
 # if you do so, you must call the __init__ method with register=False,
 # and then call the register method after the ivars have been added.
 # But rather than creating the ivars in Objective-C land, it is easier
-# to just define python-based instance yvariables in your subclass's init
+# to just define python-based instance variables in your subclass's init
 # method.
 #
 # This class is used only to *define* the interface and implementation
--- a/pyglet/libs/darwin/cocoapy/runtime.py
+++ b/pyglet/libs/darwin/cocoapy/runtime.py
@@ -481,14 +481,18 @@ class OBJC_SUPER(Structure):
 
 OBJC_SUPER_PTR = POINTER(OBJC_SUPER)
 
-#http://stackoverflow.com/questions/3095360/what-exactly-is-super-in-objective-c
-def send_super(receiver, selName, *args, **kwargs):
+# http://stackoverflow.com/questions/3095360/what-exactly-is-super-in-objective-c
+#
+# Setting preventSuperclassRecursion to True prevents a bug in which the method
+# is called on the same object, leading to infinite recursion, see
+# https://github.com/pyglet/pyglet/issues/5. This is a temporary hack that
+# solves the problem, but doesn't resolve the underlying issue.
+def send_super(receiver, selName, *args, preventSuperclassRecursion=False, **kwargs):
     if hasattr(receiver, '_as_parameter_'):
         receiver = receiver._as_parameter_
     superclass = get_superclass_of_object(receiver)
-    superclass_ptr = c_void_p(objc.class_getSuperclass(superclass))
-    if superclass_ptr.value is not None:
-        superclass = superclass_ptr
+    if preventSuperclassRecursion:
+        superclass = c_void_p(objc.class_getSuperclass(superclass))
     super_struct = OBJC_SUPER(receiver, superclass)
     selector = get_selector(selName)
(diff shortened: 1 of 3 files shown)


einarf:
My my 2013 macbook I get the following error when trying to resize/min/maximize the window
```
2020-05-12 14:16:14.818 python[58436:3308746] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[PygletWindow nextEventMatchingMask:untilDate:inMode:dequeue:]: unrecognized selector sent to instance 0x7fb9a2a1cb10'
*** First throw call stack:
(
        0   CoreFoundation                      0x00007fff38bf88d7 __exceptionPreprocess + 250
        1   libobjc.A.dylib                     0x00007fff719d6a9e objc_exception_throw + 48
        2   CoreFoundation                      0x00007fff38c77a16 -[NSObject(NSObject) __retain_OA] + 0
        3   CoreFoundation                      0x00007fff38b5ce4a ___forwarding___ + 829
        4   CoreFoundation                      0x00007fff38b5ca78 _CF_forwarding_prep_0 + 120
        5   _ctypes.cpython-37m-darwin.so       0x000000010a434177 ffi_call_unix64 + 79
        6   ???                                 0x00007ffeead67370 0x0 + 140732838343536
)
libc++abi.dylib: terminating with uncaught exception of type NSException
Abort trap: 6
```

olehermanse:
@eterevsky Thank you very much for working on this! This seems to fix the issue of crashes (segfault / recursion errors) on my mac. Unfortunately, it seems to break functionality for the older macs which don't have the issue.

Interestingly, when I tried reproducing the issue today, I initially couldn't. My touch bar crashed a few days ago, and so it was black. After rebooting my mac, my touch bar is working again, and the issue is reproducible. A theory is that the new event data pyglet receives is somehow related to the touch bar, maybe.

eterevsky:
It's a pity. I'll try to think of some other way to detect when we need to apply this hack. Ideally we could just compare the pointer to the class to the class of the current object and do another step only when needed, but I don't quite see how to get it.

Alternatively maybe it's possible to just get the pointers to parent classes directly (we know they are NSWindow and NSView).

eterevsky:
@olehermanse That's a great find! I just closed the lid on my laptop, leaving only the external monitor and switching off the touchbar, and I got exactly the same behavior as you and @einarf. So it seems that it *is* related to touchbar.

So in addition to the potential solutions I mentioned in my previous comment, it should be possible to apply the fix only when the touchbar is active.

eterevsky:
@olehermanse @einarf I made a different fix for #5 where I just resolve the superclass by name in two places. It seems to work correctly both with touchpad on and off. Could you please try it on your Macs?

einarf:
Looks good here with the last change 👍

pyglet/libs/darwin/cocoapy/runtime.py
```diff
@@ -1046,7 +1045,7 @@ def convert_method_arguments(encoding, args):
 # if you do so, you must call the __init__ method with register=False,
 # and then call the register method after the ivars have been added.
 # But rather than creating the ivars in Objective-C land, it is easier
-# to just define python-based instance variables in your subclass's init
+# to just define python-based instance yvariables in your subclass's init
```
benmoran56, line 1048: Maybe a type here with "yvariables"?

benmoran56:
That you very much for your work on this Eterevsky, and thanks to everyone else for the collaborative effort!

pyglet/libs/darwin/cocoapy/runtime.py
eterevsky: Fixed, thank you.

benmoran56:
Hey @olehermanse, this looks good to merge in. If it's OK with you, please go ahead (or let me know and I'll merge it).

benmoran56:
Thank you @eterevsky!

eterevsky:
My pleasure!

PR #200, opened 2020-05-18
Register event type decorator example

t-o-n-y-p:
Added example of the other way to register an event - using a class decorator.


benmoran56:
Looks good. Would it be possible to add a short docstring at the top of the file, that explains what's going on? Something that explains that  when you create your own EventDispatcher classes, you can use this class decorator as an alternative way of registering events.

t-o-n-y-p:
Done!

benmoran56:
Thanks Tony!

Issue #205, opened 2020-05-24
Intercepting events in a class implementing EventDispatcher


eterevsky:
## TL;DR

I would like to add an ability for the class inheriting from `EventDispatcher` to add high-priority handlers, in addition to low-priority handlers that they have now.

## Current behavior

Currently the logic of dispatching events in the `EventDispatcher` is as follows:

* Go through `self._event_stack` and call the handlers until one returns `EVENT_HANDLED`.
* If none of the handlers returned `EVENT_HANDLED`, call the method with the same name on the class itself.

## The problem

I need to first call the handler on the class, and then call the external handlers.

My use case is as follows: I'm trying to implement a simple UI (see [current version](https://github.com/eterevsky/seer/blob/2551d3ab2613f080ab4473249f7d46861b0c517e/ui.py)). In particular I have a layout class `StackLayout` that owns some `Pane` objects and routes events to them. Then there's code that listens to events on `Pane` objects and does the actual rendering. The problem is that that application code is called *before* `Pane.on_event`, and it means for example that when the handler for `on_resize` is called, the `Pane` object fields are not updated yet with the new size.

It seems that the current logic of `EventDispatcher` was written under the assumption that all the events of the class are raised by the code in it, and `DispatcherClass.on_event` methods are just fallback handlers. But in my case it's more convenient to raise the events from the parent class.

## Proposal

In my mind the cleanest solution for this would be to add optional decorators for the event handlers that would classify them as either "default" (or maybe "fallback") or "intercept":

```python
    @events.intercept
    def on_draw(self, ...):
      # ...

    @events.default  # Optional. Same behavior as now.
    def on_resize(self, ...)
      # ...

    @events.intercept('on_resize')
    def on_resize_intercept(self, ...):
      # ...
```

The intercept handler will be called *before* the main event stack, and the default handler will be called *after*, provided that all handlers have returned `EVENT_UNHANDLED`.

This behavior could optionally be generalized to the external event listeners by introducing the second event stack (or more precisely queue), that would grow in the opposite direction, so that the handlers are called in the following order `intercept_handler1`, `intercept_handler2`, `intercept_handler3`, `default_handler3`, `default_handler2`, `default_handler1`, where the handler with the lower number was registered earlier:

```python
    @dispatcher.event(intercept=True)
    def on_resize(width, height):
        # ...
```

or

```python
    @dispatcher.event
    @events.intercept
    def on_resize(width, height):
        # ...  
```

## Alternatives

[...]

PR #207, opened 2020-05-25
Add the ability to create high-priority event handlers

eterevsky:
See #205 for the detailed proposal.

This change introduces a decorator `@intercept` that marks an event handler as higher-priority. Handlers with this decorators are invoked in order of registration, as opposed to the normal handlers which are invoked in the reverse order.

I haven't yet updated the documentation for `event` module. I'll do it if the maintainers are ok with the proposed API.

Also, I raise EventException instead of AssertionError in case an unknown event is raised.

### DIFF
--- a/pyglet/event.py
+++ b/pyglet/event.py
@@ -162,6 +162,12 @@ EVENT_HANDLED = True
 EVENT_UNHANDLED = None
 
 
+def intercept(handler):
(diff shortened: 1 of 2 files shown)


eterevsky:
Closing, since I sent another implementation in #218 218

PR #216, opened 2020-05-29
Atlas border fix and enhancement

caffeinepills:
Fix issue where `TextureBin` was not passing border attribute to `TextureAtlas`. Resource module not actually utilizing borders because of that, as `border` is false by default.

Added `atlas_border` attribute to the `resource.image` function if you wish to specify if it adds pixel borders.

Made change where border attribute is the amount of pixels to separate, rather than enabled as true or false. This allows more pixels to be padded in rarer cases where 1 pixel is not enough of a border. 0 can still be set for no border if you wish to save texture memory and don't have any bleeding/filtering.

### DIFF
--- a/pyglet/image/atlas.py
+++ b/pyglet/image/atlas.py
@@ -178,7 +178,7 @@ class Allocator:
 class TextureAtlas:
     """Collection of images within a texture."""
 
-    def __init__(self, width=2048, height=2048, border=False):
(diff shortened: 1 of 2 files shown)


benmoran56:
Thanks for this. I'm not sure how we missed the border not being passed before, but it's good to make it flexible as well.

PR #218, opened 2020-06-01
Refactor events framework and add event priorities.

eterevsky:
See #205 for context.

This PR contains almost complete rewrite of the events framework. The API remained mostly the same, but most of the code except `_raise_dispatch_exception` and `event` decorator was rewritten to accommodate the new representation of event handlers stack/queue, and hopefully to simplify the code and the API. The changes include:

* The notion of "event frames" has been removed. The queues of event handlers for each event types are now independent.
* The methods `set_handlers`, `set_handler`, `pop_handlers` have been removed. The main remaining methods for manipulating the handlers are `push_handlers` and `remove_handlers` that accept exactly the same arguments. There are also `push_handler` and `remove_handler` methods for adding/removing only one handler, but they aren't really necessary, and aren't mentioned in the documentation.
* The handlers are now stored in lists of (priority, handler) pairs for each event type. New handlers are added after all existing handlers with strictly higher priority, but before handlers with lower or equal priority. This preserves the current convention that the last added handler is executed first.
* The dispatcher object is not treated differently from other handlers, but is added via `self.push_handlers(self)` the first time any handler is added or an event is fired. This means that overwriting a method on the dispatcher class will not always work for adding a new handler. I removed this recommendation from the docs, because it was error-prone anyway. The three recommended way of adding handlers are now: `@event` decorator, `push_handlers`, and subclassing the dispatcher class.
* Parts of the module documentation are rewritten to reflect the changes. I'm not sure what documentation generator Pyglet is using, so I just wrote in plain Markdown.
* I've added a bunch of new unit tests.

Please have a look at the code and tell me what you think.


benmoran56:
Sorry I've been a bit busy, but I'll give this a look over today. Thanks!

benmoran56:
Sorry for the delay! It's been a really busy week. I've had a chance to look over the code, but not actually test it yet. I've got it cloned locally so I'll test it later this event. 

Just want to say: excellent work on the documentation!

eterevsky:
Thanks for the heads up! There's no rush. I haven't yet tested this in Pyglet 2 since it's currently broken on Mac, but I tried using it in my own small project, and it worked fine so far.

LennyPhoenix (approved):
Looks great to me 👍 
As Benjamin said: Great work on the docs!

benmoran56 (approved):

benmoran56 (approved):

PR #219, opened 2020-06-01
Added examples of using sprite sheets for simple animation

olehermanse:
Open to suggestions for how to do this in a better / cleaner way.


caffeinepills:
I think the better way would be to use the `Animation` class from pyglet. At first glance there may not be a lot of flexibility with it, and maybe that's a problem with the docs but you can accomplish this in a few ways. I think this boils down to: how can I make my own Animations from a grid?

While you can use `pyglet.image.Animation.from_image_sequence` it falls short when your list may not be in the order, or you want to customize it with effects or other behavior. Since the sprite just essentially plays the `Animation` in order, you can treat the whole thing like a list.

Here is a quick example of creating an animation using a list of indexes in an image grid, and even a couple of common animation effects:
```
def animation_from_grid(image_grid, index_list, duration=0.2, effect=None):
    frames = [AnimationFrame(image_grid[idx], duration) for idx in index_list]
    if effect == 'reverse':
	frames.reverse()
    elif effect == 'pingpong':
	frames += reversed(frames)

    return Animation(frames)
```
This way you can do something like
```
move_animation = animation_from_grid(image_grid, [0, 1, 2, 3])
attack_animation = animation_from_grid(image_grid, [3, 5, 0, 2])
```

You can also access an animation's frames from the `frames` property and re-use the frames (as long as you don't modify the frames individually). 
`move_backwards = Animation(reversed(move_animation.frames))`

Hope that helps.

benmoran56:
I agree that the documentation could use some work, and show a few more examples. As caffeinepills pointed out there is already a lot of flexibility in creation, and reuse of frames, but maybe it's not so obvious. I _did_ update the docstrings a little while ago, but the programming guide could use some work.

benmoran56:
Since this was started, we actually ended up adding functionality to the Sprite class for this. 
(You can set `Sprite.paused` and `Sprite.frame_index`). 
Since it kind of duplicates these ideas, would it be alright to close this one?

benmoran56:
I'm going to close this for now since the Sprite class updates include it for the most part. Lets follow up if necessary!

PR #220, opened 2020-06-04
Pyglet 1.0 maintenance

planetbyter:
pip install pyglet

### DIFF
--- a/pyglet/__init__.py
+++ b/pyglet/__init__.py
@@ -55,7 +55,7 @@ import sys
 #:    >>> parse_version(pyglet.version) >= parse_version('1.0')
 #:    True
 #:
-version = '1.0'
+version = '1.0.1'
 
 def _require_ctypes_version(version):
     # Check ctypes version
--- a/pyglet/__init__.py
+++ b/pyglet/__init__.py
@@ -55,7 +55,7 @@ import sys
 #:    >>> parse_version(pyglet.version) >= parse_version('1.0')
 #:    True
 #:
-version = '1.0.1'
+version = '1.0.2'
 
 def _require_ctypes_version(version):
     # Check ctypes version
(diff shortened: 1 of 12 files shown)


benmoran56:
Hello, 
Your pull request is attempting to merge changes from the 1.0-maintenance branch into master.
Are you indeeding attempting to update the 1.0 branch? If so, could I ask what the use case is? This branch has long been unsupported.

DoAltPlusF4:
Yeah... I don't think merging changes from 2008 is really a good idea...

benmoran56:
I'm going to close this due to lack of reply (and other potential issues). If you read this, and want to give it another try, please update here or open a new PR. 
Thanks!

bl-ue:
Why isn't this closed?  It's likely just an experiment.  @planetbyter didn't even write the PR title; GitHub generated it based on the name of branch.

benmoran56:
When I tried to close this initially, it also had the side effect of deleting the 1.0 branch. It seems like this might have been a GitHub bug of some sort.

Realistically though, these extremely old branches probably don't need to be kept around anymore. There is no chance of anyone doing any maintenance releases for such old releases, since they don't work anymore on supported Python and modern OSs.

bl-ue:
>When I tried to close this initially, it also had the side effect of deleting the 1.0 branch. It seems like this might have been a GitHub bug of some sort.

Weird.  Well you may be able to restore the branch if it gets deleted, but 🤷🏻‍♂️

LennyPhoenix:
You should just be able to push the branch from your local repo if it tries to delete branch

bl-ue:
Also, GH lets you restore branches when you delete them, I think even right from the PR's page itself.

benmoran56:
Looks like I was able to close it this time _without_ it automatically deleting the branch. 
Thanks guys!

PR #223, opened 2020-06-05
BUG: Garbage collect more carefully

larsoner:
Fixes several errors of the form:
```
  File "/home/larsoner/python/pyglet/pyglet/media/codecs/ffmpeg.py", line 586, in __del__
TypeError: 'NoneType' object is not callable
```
I tried also adding attributes like `self._avcodec = avcodec` and `self._ffmpeg_unref_packet = ffmpeg_unref_packet` to try and get these not to get GC'ed before the objects themselves, but it did not work. This at least suppresses the warnings that these things can't be GC'ed.

### DIFF
--- a/pyglet/media/codecs/ffmpeg.py
+++ b/pyglet/media/codecs/ffmpeg.py
@@ -429,7 +429,8 @@ class _Packet:
         self.timestamp = timestamp
 
     def __del__(self):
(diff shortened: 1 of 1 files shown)


larsoner:
Changed base to `1.5-maintenance`, let me know if that's not correct and I can change

Issue #186, opened 2020-04-19
The colors are not correct when playing videos with "pyglet.media.Player()"


zhavtom:
```python
import pyglet
 
vidPath = 'resource/video.mp4'
window= pyglet.window.Window()
player = pyglet.media.Player()
source = pyglet.media.StreamingSource()
MediaLoad = pyglet.media.load(vidPath)
 
player.queue(MediaLoad)
player.play()
 
 
@window.event
def on_draw():
    if player.source and player.source.video_format:
        image = player.get_texture()
        image.blit(0,0)
 
pyglet.app.run()
```
I tried the above code, the video played in a different color than the original video file.
The brightness didn't change, but the hue was not right (red was shown as blue).
Here's a comparison between the original video and the displayed video(left: original right: pyglet):
<img width="451" alt="diff_color_py" src="https://user-images.githubusercontent.com/38828702/79687313-68de1b80-8281-11ea-9e37-f9ad9c584993.png">

pyglet version is 1.5.3 .

zhavtom:
Opreating System: Windows 10 Pro 64bit

caffeinepills:
Are you using any sort of blue light filter like Windows 10 nightlight or flux?

zhavtom:
No.
Sorry, the image I posted earlier is unedited; the two windows are next to each other.
The other windows are normal in color, but only Pyglet looks like this.

benmoran56:
Is it possible to share a short video clip somewhere?

zhavtom:
https://drive.google.com/file/d/1nREN9nYlMbD0vdN6KxEqbszNuybVHAIP/view?usp=sharing
This is the original video.

benmoran56:
Thanks for that. We'll take a look. In the mean time, if you install FFmpeg, you can use the FFmpeg decoder which should work OK. To enforce the use of a specific decoder: 

```py
decoder = pyglet.media.codecs.ffmpeg.FFmpegDecoder()
src = pyglet.media.load(filename, decoder=decoder)
```

Instructions on how to install FFmpeg for Windows can be found here: 
https://pyglet.readthedocs.io/en/latest/programming_guide/media.html#ffmpeg-installation

zhavtom:
```python
Traceback (most recent call last):
  File "C:\python\test\pyglet_ffmpeg.py", line 6, in <module>
    decoder = pyglet.media.codecs.ffmpeg.FFmpegDecoder()
[...]

PR #224, opened 2020-06-05
Image format incorrect for WMF playback.

caffeinepills:
Despite format saying RGB, Windows is almost always BGRA . This fixes the color issues where blue's and reds are incorrect.

#186


benmoran56:
Thanks for that.

PR #227, opened 2020-06-11
Added QoL properties to ScissorGroup.

DoAltPlusF4:
Added some quality of life properties to the ScissorGroup class at `examples/opengl_scissor.py`.
These include X, Y, width, and height.

### DIFF
--- a/examples/opengl_scissor.py
+++ b/examples/opengl_scissor.py
@@ -84,18 +84,6 @@ class ScissorGroup(pyglet.graphics.Group):
     def unset_state(self):
         glDisable(GL_SCISSOR_TEST)
 
-    # For efficient drawing, pyglet will internally consolidate equivalent
-    # Groups into the same draw call when you add objects to a Batch. For
-    # this to work, __eq__ and __hash__ methods must be defined so that the
-    # Groups can be checked for equality.
-    def __eq__(self, other):
-        return (self.__class__ is other.__class__ and
-                self.parent is other.parent and
-                self._area == other._area)
-
-    def __hash__(self):
-        return hash((id(self.parent), self._area))
-
 
 ###################################################
 # Create an instance of our Group, and some Sprites
--- a/examples/opengl_scissor.py
+++ b/examples/opengl_scissor.py
@@ -46,43 +46,20 @@ class ScissorGroup(pyglet.graphics.Group):
 
     def __init__(self, x, y, width, height, parent=None):
         super().__init__(parent)
-        self._area = x, y, width, height
+        self.x, self.y = x, y
+        self.width, self.height = width, height
 
     @property
-    def x(self):
-        return self._area[0]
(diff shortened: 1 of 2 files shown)


benmoran56:
Thanks for the pull request. The reason I didn't make these attributes public, was because it might cause confusion if they are changed. If two groups are created with the same area, they will be consolidated internally: 
```py
groupa = ScissorGroup(10, 10, 100, 100)
groupb = ScissorGroup(10, 10, 100, 100)
groupa == groupb   # True
```
If you then change the area these groups, only one of them will actually have any effect (as the other one is discarded internally, and not used when drawing). 

We could dump the hash/eq methods so that different instances of this group are not consolidated. 
It would avoid issues like this, but would cause performance issues if someone were to create multiple instances of the same Group instead of reusing the same instance. Maybe that would be better though.

DoAltPlusF4:
Hmm, I didn't think about that. :/

As an end-user, I would probably prefer it if they stayed as separate instances, though I'm not 100% sure on the performance implications of this.

DoAltPlusF4:
I also realise that I accidentally added another commit (Anchor constants to `pyglet.text`), this was by mistake, do you know how I should remove it? (My fault for not making a separate branch for it :P )

benmoran56:
If you're just going to make a few, then there won't be any performance problem. The problems arise if you try to do something like this: 
```py
for i in range(tile_width):
    group = ScissorGroup(10, 10, 100, 100)
    sprite = Sprite(....group=group)
```
For something like `OrderedGroup(order)` that pattern is fine, since they don't change and can get consolidated together internally. Without the hash/eq methods, however, you would end up with every single Sprite instance being in a separate Group, and hence a separate internal draw call. The solution is to do something like this: 
```py
group = ScissorGroup(10, 10, 100, 100)
for i in range(tile_width):
    sprite = Sprite(....group=group)
```
That way you have only a single Group instance, so eq/hash methods don't come into play since no consolidation will happen. 


As for the stray commit, I think the only thing you can do is to revert the file back the way it was an commit again. Other than that I'm not sure.

benmoran56:
Since this Group can't be consolidated if the scissor area is different, maybe we just dump the hash/eq methods, and just expose x, y, width, height attributes directly. It will simplify the class, and won't need the extra properties.

In that case we should also write in the docstring that you should use the same exact instance for all objects that will be in the group, and not make a new instance each time. What do you think?

DoAltPlusF4:
Sounds good 👍

benmoran56:
How about dropping all of the properties/setters, and just exposing `x, y, width, height` attributes directly?

DoAltPlusF4:
I removed the `x`, `y`, `width` and `height` properties in favour of exposing them directly.
I also added an `area` property.

benmoran56:
Thanks!

PR #228, opened 2020-06-12
Add on_mouse_drag event to MouseStateHandler.

DoAltPlusF4:
I have noticed that on_mouse_drag blocks on_mouse_motion, meaning that the MouseStateHandler would not track the mouse position while a mouse button is help. This PR should fix that.


benmoran56:
That's just the way the OS sends the events, so you really do need to account for both cases. Your PR will fix that. Thanks!

PR #230, opened 2020-06-16
Fix for issue regarding numba.

Torxed:
Converts `<class 'OSError'>` into `str(err)` when calling `warnings.warn(err)`.

Relevant gist: https://gist.github.com/Torxed/10f3bd35ddbd8ceabda8f7a002fdff6c

Code to reproduce:
```py
import pyglet
import numba

game_window = pyglet.window.Window(600, 400)
icon = pyglet.image.load("test.png")
game_window.set_icon(icon)
pyglet.app.run()
```

Original report: https://stackoverflow.com/questions/62383035/error-while-import-numba-module-in-pyglet-projects

### DIFF
--- a/pyglet/image/codecs/wic.py
+++ b/pyglet/image/codecs/wic.py
@@ -289,7 +289,7 @@ class WICDecoder(ImageDecoder):
         try:
             ole32.CoInitializeEx(None, COINIT_MULTITHREADED)
         except OSError as err:
-            warnings.warn(err)
+            warnings.warn(str(err))
 
         ole32.CoCreateInstance(CLSID_WICImagingFactory,
                                None,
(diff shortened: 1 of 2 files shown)


benmoran56:
Thanks, looks good!

caffeinepills:
This is bizarre as `warnings.warn` seems to take exceptions:
```
>>> import warnings
>>> test = OSError("Failed")
>>> warnings.warn(test)

Warning (from warnings module):
  File "__main__", line 1
UserWarning: Failed
>>> 
```

benmoran56:
The best we can figure is that Numba is mutating things somehow during the JIT process. The traceback indicated that `warn` would only accept string/bytes, but that certainly doesn't seem to be the case in my Python interpreter either. 

In any case, this seems like an easy enough fix for whatever is causing it.

Torxed:
I was having issues using the `-m trace --trace` module to diagnose the issue. Not sure the JIT optimization liked what I was doing.

PR #239, opened 2020-07-01
Update events.py to use KeyStateHandler 

Hibot12:
Now KeyStateHandler is used, instead of the 'old' approach.

### DIFF
--- a/examples/programming_guide/events.py
+++ b/examples/programming_guide/events.py
@@ -47,15 +47,21 @@ from pyglet.window import mouse
 
 window = pyglet.window.Window()
 
+key_handler = key.KeyStateHandler()
 
-@window.event
-def on_key_press(symbol, modifiers):
-    if symbol == key.A:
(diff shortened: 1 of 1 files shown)


DoAltPlusF4:
Window events are not an "old" approach. It entirely depends on what you are trying to do.
Keep in mind that the KeyStateHandler uses events itself to track input. 
(https://github.com/pyglet/pyglet/blob/c32556e92e854b9001da39fba70c4deb63c28878/pyglet/window/key.py#L67)

The events.py example is meant to be an example of events anyway, not the KeyStateHandler.

benmoran56:
I agree. The KeyStateHandler is more of a generic thing, for convenience. For advanced uses, you want to use the event system directly.

We could add a separate example just for the KeyStateHandler though.

Hibot12:
I think creating another example would be great for new users. Definitely should do it.

Hibot12:
I will open another PR for this.

PR #240, opened 2020-07-06
Create events_key_state_handler.py

Hibot12:
As intended in #239, this PR proposes creating another example of using KeyStateHandler.

### DIFF
--- a/examples/programming_guide/events_key_state_handler.py
+++ b/examples/programming_guide/events_key_state_handler.py
@@ -33,6 +33,10 @@
 # POSSIBILITY OF SUCH DAMAGE.
 # ----------------------------------------------------------------------------
 
+"""
+This example shows the KeyStateHandler polling approach when it comes to catching keyboard events.
+"""
+
 import pyglet
 
 window = pyglet.window.Window()
--- a/examples/programming_guide/events_key_state_handler.py
+++ b/examples/programming_guide/events_key_state_handler.py
@@ -33,14 +33,6 @@
 # POSSIBILITY OF SUCH DAMAGE.
 # ----------------------------------------------------------------------------
 
-"""
-"""
-
-from __future__ import print_function
-
-__docformat__ = 'restructuredtext'
-__version__ = '$Id$'
-
 import pyglet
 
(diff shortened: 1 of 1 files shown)


examples/programming_guide/events_key_state_handler.py
```diff
@@ -0,0 +1,71 @@
...
+
+import pyglet
+
+window = pyglet.window.Window()
+
+key_handler = pyglet.window.key.KeyStateHandler()
+window.push_handlers(key_handler)
+
+
+def update(*args):
```
benmoran56, line 52: To to be clear what's being passed, could you change *args to "dt"? Even if it's unused, it shows what the update method is being passed.

examples/programming_guide/events_key_state_handler.py
```diff
@@ -0,0 +1,71 @@
...
+
+"""
+"""
+
+from __future__ import print_function
+
+__docformat__ = 'restructuredtext'
+__version__ = '$Id$'
+
+import pyglet
```
benmoran56, line 44: Could you remove everything above the "import pyglet" line, and below the license section? The future import and docstring declarations are not needed anymore.   I know that they're still in the other examples, but that will get cleaned out eventually.

benmoran56:
Thank you!  I made a few comments, but other than that it looks good. If you make new commits on your end, it will automatically update the PR.

DoAltPlusF4:
In the docstring I would recommend making it clear that the KeyStateHandler is more of a polling approach to getting user input, and that both methods have different use cases.

examples/programming_guide/events_key_state_handler.py
Hibot12: Yep. Doing it right now...

examples/programming_guide/events_key_state_handler.py
Hibot12: Done.

Hibot12:
Sounds great for me, @DoAltPlusF4 .

benmoran56:
Thanks for the changes. Merged!

PR #243, opened 2020-07-11
Added transpose, invert, some row operations for Mat4

kcmw3e:
Passed some simple tests; more may be needed in case something's a little off.
Transpose could potentially quicken inverse if matrix passes a certain condition (more research needed) --> in some cases the transpose of a matrix is the inverse.
Row operations used in invert.


benmoran56:
Thanks! I haven't tested it yet, but since it's not modifying any existing code I'll go ahead and merge it in.
We can fix it if necessary after we get some tests made.

Issue #246, opened 2020-07-15
Possiblity to create windows with transparent background


roipoussiere:
I would like to create a window transparent background, where only shapes I draw on it are visible.

I used `window.Window.WINDOW_STYLE_BORDERLESS` to remove window borders:

```py
from pyglet import app, window, shapes
win = window.Window(width=400, height=400, style=window.Window.WINDOW_STYLE_BORDERLESS)
```

But how to remove the black background?

benmoran56:
Unfortunately this is not currently possible to do in pyglet.

roipoussiere:
May I ask why it's not possible? Is this technically not possible by design or just not implemented for now? Do you think I will be able to do it at lower lever with OpenGL functions?

Sorry to ask, but transparent background is required for my new project so this point will basicaly determine if I'm going to chose Pyglet or not. :)

benmoran56:
It would need to be implemented in the window platform code for all platforms, but so far no one has done that work. It should be possible, but I don't think anyone is planning to work on it at the moment.

benmoran56:
Implemented via https://github.com/pyglet/pyglet/pull/434

PR #249, opened 2020-07-16
Add `transparent_background` option and implement it on xlib

roipoussiere:
This PR adds the `transparent_background` option to the window and is a first try of implementing it on xlib.

Related issue: #246

I don't have any computer running on Microsoft Windows or MacOS, so I can't work on adding support for these systems, but I will be very happy to see contributions for this.

Code sample:

```py
from pyglet import app, window, shapes

win = window.Window(width=400, height=400,
                    style=window.Window.WINDOW_STYLE_BORDERLESS, transparent_background=True)

@win.event
def on_draw():
    win.clear()

    circle = shapes.Circle(x=200, y=200, radius=200, color=(255, 255, 255))
    circle.draw()

if __name__ == '__main__':
    app.run()
```

Reviews welcome!

Thanks to @Goutte for the pair-programming session on this.

**TODO**
- [ ] when setting the disc color to `(128, 128, 128)`, the disc appears half-transparent instead of opaque grey;
- [ ] ability to define the color alpha channel with `color(r, g, b, a)`.

### DIFF
--- a/pyglet/window/__init__.py
+++ b/pyglet/window/__init__.py
@@ -394,6 +394,9 @@ class BaseWindow(with_metaclass(_WindowMetaclass, EventDispatcher)):
     #: A window style without any decoration.
     WINDOW_STYLE_BORDERLESS = 'borderless'
 
+    #: The default background opacity.
+    TRANSPARENT_BACKGROUND_DEFAULT = False
+
     #: The default mouse cursor.
     CURSOR_DEFAULT = None
     #: A crosshair mouse cursor.
@@ -468,6 +471,7 @@ class BaseWindow(with_metaclass(_WindowMetaclass, EventDispatcher)):
     _caption = None
     _resizable = False
     _style = WINDOW_STYLE_DEFAULT
+    _transparent_background = TRANSPARENT_BACKGROUND_DEFAULT
     _fullscreen = False
     _visible = False
     _vsync = False
@@ -503,6 +507,7 @@ class BaseWindow(with_metaclass(_WindowMetaclass, EventDispatcher)):
                  caption=None,
                  resizable=False,
                  style=WINDOW_STYLE_DEFAULT,
+                 transparent_background=TRANSPARENT_BACKGROUND_DEFAULT,
                  fullscreen=False,
                  visible=True,
                  vsync=True,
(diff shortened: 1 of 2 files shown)


benmoran56:
Thank you for the PR. I had a look, and it does seem like this is quite easy to do in X. 
I'll have to do some reading to see if this is possible to achieve in Windows and OSX. 

If it's possible, then we can try to find some help implementing it on those platforms. 
If it's **not** possible to achieve on Windows and/or OSX, then we may have to rethink this. We generally don't want to add platform specific features, but it would still be possible to add as a "hidden" feature for Linux. For example, something like a class variable to set before instantiating the Window: 
`pyglet.window.Window.transparent_background = True`

roipoussiere:
At a first look, it seems possible:
- [on MacOS](https://stackoverflow.com/questions/34531118/how-can-i-create-a-window-with-transparent-background-with-swift-on-osx);
- [on Windows](https://stackoverflow.com/questions/3970066/creating-a-transparent-window-in-c-win32)

I hope we will implement this cross-platform!

By the way, I suggest an easier way to create transparent-background windows: since we usually don't want to create a transparent-background window with opaque borders, we could add a `WINDOW_STYLE_BACKGROUNDLESS` instead of passing the `transparent_background` option to the constructor, ie.:

- this: `stylewindow.Window(width=400, height=400, style=window.Window.WINDOW_STYLE_BACKGROUNDLESS)`
- instead of this: `stylewindow.Window(width=400, height=400, style=window.Window.WINDOW_STYLE_BORDERLESS, transparent_background=True)`

I will work on this PR the next days to fix the problem with the wrong colors.

benmoran56:
Sorry for the delay. 

I do like the idea of simplifying the API by making it just another `style`.  Does setting it to transparent automatically hide the border & window controls?

roipoussiere:
Hello! I was interested for the transparent background for a personal project, but by the meantime I switched to an other programming language for this project, so I don't think I will continue to work on this PR, sorry.

However I hope that my small contribution will encourage you and other contributors to finish the implementation of transparent background support in Pyglet. :)

benmoran56:
This is implemented via #434, so we can finally close this.
Thanks for getting the ball rolling @roipoussiere, even though you didn't end up using it yourself.

PR #255, opened 2020-08-01
Added CenteredCamera class.

DoAltPlusF4:
This class has `(0, 0)` located in the center of the screen, as opposed to the original class having `(0, 0)` at the bottom left.
It sub-classes the standard `Camera` class for simplicity, modifying `begin` and `end`.

Note that the new class requires access to a Window reference to function, this is so that it knows the dimensions of the viewport.

### DIFF
--- a/examples/camera.py
+++ b/examples/camera.py
@@ -128,3 +128,39 @@ class Camera:
 
     def __exit__(self, exception_type, exception_value, traceback):
         self.end()
+
+
+class CentredCamera(Camera):
+    """A simple 2D camera class. 0, 0 will be the centre of the screen, as opposed to the bottom left."""
+
+    def __init__(self, window: pyglet.window.Window, *args, **kwargs):
+        self.window = window
+        super().__init__(*args, **kwargs)
+
+    def begin(self):
+        x = -self.window.width//2/self._zoom + self.offset_x
+        y = -self.window.height//2/self._zoom + self.offset_y
+
+        pyglet.gl.glTranslatef(
+            -x * self._zoom,
+            -y * self._zoom, 0
+        )
+
+        pyglet.gl.glScalef(
+            self._zoom,
+            self._zoom, 1
+        )
+
+    def end(self):
+        x = -self.window.width//2/self._zoom + self.offset_x
+        y = -self.window.height//2/self._zoom + self.offset_y
+
+        pyglet.gl.glScalef(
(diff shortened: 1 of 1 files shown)


examples/camera.py
```diff
@@ -96,23 +96,71 @@ def move(self, axis_x, axis_y):
     def begin(self):
         # Set the current camera offset so you can draw your scene.
         # Translate using the zoom and the offset.
-        pyglet.gl.glTranslatef(-self.offset_x * self._zoom, -self.offset_y * self._zoom, 0)
+        pyglet.gl.glTranslatef(
```
benmoran56, line 99: We're targeting a 100-column line width for the pyglet codebase, so I think this is better as one line.

examples/camera.py
```diff
@@ -96,23 +96,71 @@ def move(self, axis_x, axis_y):
...
+
+class CenteredCamera(Camera):
+    """A simple 2D camera class. 0, 0 will be the centre of the screen, as opposed to the bottom left."""
+
+    def __init__(self, window: pyglet.window.Window, *args, **kwargs):
+        self.window = window
+        super().__init__(*args, **kwargs)
+
+    def begin(self):
+        x = -self.window.width//2/self._zoom + self.offset_x
```
benmoran56, line 141: We may need to go with `Window.get_framebuffer_size()` here for support on HiDPI displays. 
```py
width, height = self.window.get_framebuffer_size()
....
```

einarf: Using the framebuffer size would only really matter when setting viewport unless you intend to have 4 x larger projections for retina screens.

benmoran56: Ah, you're right!

examples/camera.py
DoAltPlusF4: Ok, I'll change it back

examples/camera.py
DoAltPlusF4: Its probably better safe than sorry, I'll use this instead 👍

benmoran56: @einarf was right and it's not needed. We're already handling that when setting the viewport, so if we used it again, we would be overcompensating.

DoAltPlusF4: Gotcha, ill leave it the same then.

benmoran56:
Thanks for this!

PR #260, opened 2020-08-08
Updated joystick.py

Bervenia:
Hello,

I've updated the joystick.py example to use the "pyglet.shapes" module as opposed to the raw OpenGL. I've also added some labels and change the layout to provide more visual feedback with inputs from joysticks.

### DIFF
--- a/examples/joystick.py
+++ b/examples/joystick.py
@@ -33,6 +33,7 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 # ----------------------------------------------------------------------------
+
 '''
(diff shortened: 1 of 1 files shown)


Bervenia:
Honestly it was left over code from the original, but it forces the screen to update. This because with its removal, nothing visual updates even on button press. Ie an empty update function to force draw calls.

DoAltPlusF4:
Hmm, I'm pretty sure I remember @benmoran56 mentioning that the `on_draw` event will be separated from update calls and events, so this may not be required any more. Was this made in 2.0?

Bervenia:
This was updated in pyglet version 1.5.7. my thoughts being that 1.5.7 could use the update and for 2.0 we can remove the line if it still works with its removal.

PR #262, opened 2020-08-19
Added current_frame_image property to the Sprite class.

DoAltPlusF4:
This is a small utility property that users may find helpful when creating visual effects or just general animation.
Just thought I'd make this PR just before 1.5.8 is released.

### DIFF
--- a/pyglet/sprite.py
+++ b/pyglet/sprite.py
@@ -663,6 +663,13 @@ class Sprite(event.EventDispatcher):
         self._visible = visible
         self._update_position()
 
+    @property
+    def current_frame_image(self):
+        if not hasattr(self.image, "frames"):
+            return self.image
+        frame = self.image.frames[self._frame_index]
+        return frame.image
+
     def draw(self):
         """Draw the sprite at its current position.
 
--- a/pyglet/sprite.py
+++ b/pyglet/sprite.py
@@ -179,6 +179,8 @@ class Sprite(event.EventDispatcher):
 
     _batch = None
     _animation = None
+    _frame_index = 0
+    _paused = False
     _rotation = 0
     _opacity = 255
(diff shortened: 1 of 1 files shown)


benmoran56:
Not to get off on a tangent here, but this is some code we were tossing around in the past: 

```py
class AnimationPlayer(pyglet.sprite.Sprite):

        super().__init__(*args, **kwargs)
        self._paused = False

    def pause_animation(self):
        if self._paused or not hasattr(self, '_animation'):
            return
        pyglet.clock.unschedule(self._animate)
        self._paused = True

    def resume_animation(self):
        if not self._paused or not hasattr(self, '_animation'):
            return
        frame = self._animation.frames[self._frame_index]
        self._texture = frame.image.get_texture()
        self._next_dt = frame.duration
        if self._next_dt:
            pyglet.clock.schedule_once(self._animate, self._next_dt)
        self._paused = False

    @property
    def frame_index(self):
        return self._frame_index

    @frame_index.setter
    def frame_index(self, index):
        # Bound to available number of frames
        self._frame_index = max(0, min(index, len(self._animation.frames)-1))
```

Maybe it's a good idea to add these to the Sprite class finally.

DoAltPlusF4:
I replaced the current_frame_image property with those new functions and properties.
I have also moved the _frame_index attribute to the top of the class, with things like `_animation`, `_batch`, `_paused` and `_rotation`:
```py
class Sprite(event.EventDispatcher):
    # ...

    _batch = None
    _animation = None
    _frame_index = 0
    _paused = False
    _rotation = 0
    _opacity = 255
    _rgb = (255, 255, 255)
    _scale = 1.0
    _scale_x = 1.0
    _scale_y = 1.0
    _visible = True
    _vertex_list = None

    # ...

PR #265, opened 2020-08-24
Fix docs for VertexList

r58Playz:
Update vertexattribute.py docs based on default shaders.
The docs were incorrect.


benmoran56:
Thanks for the PR.

Just a heads-up, but the way the vertex lists, and vertex attributes are working is in flux at the moment. It's possible that some of these things might change drastically, so we should hold off on major documentation changes or updates for the immediate future.

benmoran56 (approved):

Issue #272, opened 2020-08-29
Pyglet crashes on macOS when configuring the OpenGL context version


kollmats:
Hi,

I'm trying to use GLSL 3.3 on macOS High Sierra 0.13.1, pyglet 1.5.7. However, when trying to set the version like so:

```
from pyglet.window import Window
from pyglet.gl import Config
Window(config=Config(major_version=int(a), minor_version=int(b)))
```

I get the following crash:

> Traceback (most recent call last):
  File "test.py", line 6, in <module>
    from pyglet.window import Window
  File "/usr/local/lib/python3.7/site-packages/pyglet/window/__init__.py", line 648, in __init__
    self._create()
  File "/usr/local/lib/python3.7/site-packages/pyglet/window/cocoa/__init__.py", line 197, in _create
    self.context.attach(self.canvas)
  File "/usr/local/lib/python3.7/site-packages/pyglet/gl/cocoa.py", line 300, in attach
    self._nscontext.setView_(canvas.nsview)
AttributeError: 'NoneType' object has no attribute 'setView_'

Additional info:

It seems like pyglet invokes [NSOpenGLContext](https://developer.apple.com/documentation/appkit/nsopenglcontext/1436178-initwithformat?language=objc) of Cocoa with invalid parameters, causing it to return a None value (that does not happen for, e.g., OpenGL 2.1). This can happen due to an inconsistency in pixel formats (which I've checked and ruled out) or if the "share" parameter points to an invalid OpenGL context.

kollmats:
After some further digging, I was able to get around this by disabling the pyglet shadow context. I also noticed that the documentation addresses this [here](https://pyglet.readthedocs.io/en/latest/programming_guide/context.html?highlight=shader#sharing-objects-between-contexts), but that the actual fix is somewhat different. 

The documentation suggests that the shadow context must be disabled before a ``Window`` is _created_, but my observation is that the shadow context has to be disabled before _importing_  ``pyglet.window``.

I took the liberty of creating a [PR](https://github.com/pyglet/pyglet/pull/273).

einarf:
Just linking moderngl-window's pyglet window as well in case someone need a more specific example :  https://github.com/moderngl/moderngl-window/blob/master/moderngl_window/context/pyglet/window.py

einarf:
Also, master branch makes 3.3 core context currently by default.

einarf:
[...]

PR #273, opened 2020-08-29
Updated context page in programming guide.

kollmats:
Fixing error in documentation as described in issue #272


benmoran56:
I think the previous wording was just assuming something like: 
```py
import pyglet

```
It might be a bit pedantic, but perhaps we should say something like "before creating a Window, or importing `pyglet.window`". What do you think?

benmoran56 (approved):

benmoran56:
Thanks for the PR! I went ahead and made that small update.

Issue #286, opened 2020-10-11
static analysis tools problems with pyglet


Eric-Arz:
this is based on #https://github.com/microsoft/pylance-release/issues/474

some static analysis tools including pylance have trouble providing correct and meaningful code completion for pyglet. After opening an issue for pylance their suggest to fix/improve this was this:

change pyglet/__init__.py

```
if True:
    app = _ModuleProxy("app")
    ...

if False:
    from . import app
    ...
```

If you change this to the following, it will retain its lazy loading but also work fine with Pylance:

```
from typing import TYPE_CHECKING

if not TYPE_CHECKING:
    app = _ModuleProxy("app")
    ...

if TYPE_CHECKING:
    from . import app
    ...
```

LennyPhoenix:
This would probably be a good idea to add, it might fix a couple other problems with compilers such as Nuitka too.

Eric-Arz:
Should i open a PR?

LennyPhoenix:
Sure 👍

Eric-Arz:
I made a pull request for the change: #https://github.com/pyglet/pyglet/pull/287

PR #287, opened 2020-10-13
Added TYPE_CHECKING for lazy loading

Eric-Arz:
Adresses this issue: #https://github.com/pyglet/pyglet/issues/286

Following the suggestion from one of the pylance maintainers: #https://github.com/microsoft/pylance-release/issues/474

This change makes static analysis tools like pylance able to provide code completion for all of pyglet. Without it they fail to understand the lazy loading and don't work for the  submodules.

### DIFF
--- a/pyglet/__init__.py
+++ b/pyglet/__init__.py
@@ -370,7 +370,7 @@ if not TYPE_CHECKING:
 
 # Fool py2exe, py2app into including all top-level modules
 # (doesn't understand lazy loading)
-if TYPE_CHECKING:
+else:
     from . import app
     from . import canvas
     from . import clock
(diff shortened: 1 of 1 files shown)


benmoran56:
It seems a bit odd to use TYPE_CHECKING here, since we're not using any type definitions in pyglet. Looking at the source code of the typing module, however, it just defaults to `TYPE_CHECKING = False`. It should be fine to make this change, if it will help with their inspection.

benmoran56 (changes requested):
Instead of two `if` checks, could you change this to an if/else? 

```py
if not TYPE_CHECKING:
    ....
else
    ....
```
Thank you!

benmoran56 (approved):

PR #290, opened 2020-10-16
Fix data type for text_input.py example.

charles-l:
The (older?) `vertices2i` and `colors4Bn` type specifiers were causing this example to blow up for me. This patch fixes the text_input.py script.


benmoran56:
Hi Charles, the master branch is tracking the pyglet 2.0 development code.  In order to use the examples, you'll want to checkout the 1.5-maintenance branch instead.  This code is correct for 2.0.

PR #295, opened 2020-10-23
Fix getattr

CourchesneA:
When loading a texture in a headless server, we get an AttributeError:

```
 File "/home/circleci/project/src/project/graphics.py", line 90, in load_texture
    img = pyglet.image.load(tex_path)
  File "/home/circleci/.local/lib/python3.7/site-packages/pyglet/image/__init__.py", line 190, in load
    image = decoder.decode(file, filename)
  File "/home/circleci/.local/lib/python3.7/site-packages/pyglet/image/codecs/pil.py", line 78, in decode
    image_data_fn = getattr(image, "tobytes", getattr(image, "tostring"))
AttributeError: 'Image' object has no attribute 'tostring'
```

The `default` part of `getattr` should not be run if the first attribute is found.

### DIFF
--- a/pyglet/image/codecs/pil.py
+++ b/pyglet/image/codecs/pil.py
@@ -135,7 +135,10 @@ class PILImageEncoder(ImageEncoder):
 
         # fromstring is deprecated, replaced by frombytes in Pillow (PIL fork)
         # (1.1.7) PIL still uses it
-        image_from_fn = getattr(Image, "frombytes", getattr(Image, "fromstring"))
+        try:
+            image_from_fn = getattr(image, "tobytes")
+        except AttributeError:
+            image_from_fn = getattr(image, "tostring")
         pil_image = image_from_fn(fmt, (image.width, image.height), image.get_data(fmt, pitch))
 
         try:
--- a/pyglet/image/codecs/pil.py
+++ b/pyglet/image/codecs/pil.py
@@ -75,7 +75,10 @@ class PILImageDecoder(ImageDecoder):
 
         # tostring is deprecated, replaced by tobytes in Pillow (PIL fork)
         # (1.1.7) PIL still uses it
-        image_data_fn = getattr(image, "tobytes", getattr(image, "tostring"))
+        try:
+            image_data_fn = getattr(image, "tostring")
+        except AttributeError:
+            image_data_fn = getattr(image, "tobytes")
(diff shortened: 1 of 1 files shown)


LennyPhoenix (approved):

pyglet/image/codecs/pil.py
```diff
@@ -76,9 +76,9 @@ def decode(self, file, filename):
         # tostring is deprecated, replaced by tobytes in Pillow (PIL fork)
         # (1.1.7) PIL still uses it
```
ccanepa, line 77: Pillow version comes from PIL.__version__ 
PIL.version is a compatibilty thing with the original PIL package, frozen at 1.1.7
The latest pillow is 8.0.1, does not have tostring
The latest pillow 7.x series is 7.2.0 and it has tostring

LennyPhoenix:
As ccanepa mentioned, it might be a good idea to add this to the encoder too.

pyglet/image/codecs/pil.py
CourchesneA: I seems to me like the new code is working for both version, while what we had before would throw an error if the `tostring` was not found

CourchesneA:
I added the fix to the encoder

pyglet/image/codecs/pil.py
```diff
@@ -135,7 +135,10 @@ def encode(self, image, file, filename):
 
         # fromstring is deprecated, replaced by frombytes in Pillow (PIL fork)
         # (1.1.7) PIL still uses it
-        image_from_fn = getattr(Image, "frombytes", getattr(Image, "fromstring"))
+        try:
```
ccanepa, line 138: no, here this is wrong on two accounts
 - needs to be 'frombytes' and 'fromstring', not 'tobytes' and 'tostrings'
 - it should be Image , not image 
see #297 if any doubt

CourchesneA: Ooops really sorry, that was me trying to fix stuff quickly over the weekend. Thanks for looking at this, you are right about your changes.

benmoran56 (approved):

benmoran56:
Thanks @CourchesneA for the pull request, and thanks @ccanepa for your help. 
I'll merge this in. I'm preparing a new point release in the next day or so, which will have this in it.

PR #298, opened 2020-10-26
Refactor WIC module to work with new com module

caffeinepills:
Com module was updated. WIC module needs to be updated otherwise it throws errors.


benmoran56 (approved):

benmoran56:
Thanks as always!

PR #303, opened 2020-11-14
typo in examples directory

samlaf:
Also... I can't find the graphics.py module anywhere. (not in version 1.5.9 either, which the latest docs point to)

### DIFF
--- a/doc/programming_guide/installation.rst
+++ b/doc/programming_guide/installation.rst
@@ -46,4 +46,4 @@ As mentioned above, you can also clone the repository using Git:
 
     git clone https://github.com/pyglet/pyglet.git
     cd pyglet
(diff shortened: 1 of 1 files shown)


benmoran56:
Thanks for the PR. We recently reorganized things into subfolders, since the examples folder was getting a bit messy. 

It might be best to just point it towards the simple Hello World example, like: 
```py
python examples/hello_world.py
```

benmoran56 (changes requested):
How about updating it to use `examples/hello_world.py` instead? That might make more sense anyway.

samlaf (commented):
done

benmoran56 (approved):
Thanks, perfect!

PR #305, opened 2020-11-15
OpenGL core fixes

einarf:
Fixes various issues when using a pure core context

* Aux buffers are no longer a thing
* Stencil buffer querying is different (and may need more tweaks later)
* Tweaking examples


benmoran56 (approved):
Thanks for your work on this!

PR #306, opened 2020-11-16
Model refactor + extended example

einarf:
- Separate shaders for each group type
- view matrix calculated with Mat4
- New BaseMaterialGroup
- `math` tweaks


benmoran56 (approved):
Looks good!

Issue #274, opened 2020-09-05
Import Error: Can't find framework %s." % path


akarsh1995:
When I try to run `from pyglet.gl import *`

The code line 262 tries to find OpenGL inside the directory `/System/Library/Frameworks/OpenGL.framework`. 
https://github.com/pyglet/pyglet/blob/d563b7a001506758a77fde035c22c786476d0093/pyglet/lib.py#L262
The function `return None` which in turn `raise ImportError("Can't find framework %s." % path) # on line 283`

-----
**System Information**

OS: Mac OS Big Sur Public 11.0 Beta (20A5354i).

```sh
Darwin Akarshs-MacBook-Pro.local 20.0.0 Darwin Kernel Version 20.0.0: Fri Aug 14 00:25:13 PDT 2020; root:xnu-7195.40.44.151.1~4/RELEASE_X86_64 x86_64
```

Tree structure of the dir `/System/Library/Frameworks/OpenGL.framework` in my system. 
```sh
 akarshjain@Akarshs-MacBook-Pro  /System/Library/Frameworks/OpenGL.framework  tree 
.
├── Libraries -> Versions/Current/Libraries
├── Resources -> Versions/Current/Resources
└── Versions
    ├── A
    │   ├── Libraries
    │   │   ├── 3425AMD
    │   │   │   ├── CVMCompiler
    │   │   │   └── libCoreVMClient.mono.dylib
    │   │   ├── 3600
    │   │   │   ├── CVMCompiler
    │   │   │   └── libCoreVMClient.mono.dylib
    │   │   ├── CVMCompiler
    │   │   ├── CVMServer
    │   │   ├── DumpGPURestart
    │   │   ├── ReportGPURestart
    │   │   ├── libGLVMPlugin.dylib
    │   │   ├── libGLVMPlugin31001.dylib
    │   │   ├── libGLVMPlugin3802.dylib
    │   │   └── libGLVMPlugin3902.dylib
    │   ├── Resources
    │   │   ├── BridgeSupport
    │   │   │   └── OpenGL.bridgesupport
    │   │   ├── CVMSPlugin.exp
    │   │   ├── CVMSPluginAll.exp
    │   │   ├── CVMSPluginNoExec.exp
    │   │   ├── GLByteCodes.arm64.bc
    │   │   ├── GLByteCodes.x86_64.bc
    │   │   ├── GLEngine.bundle
    │   │   │   └── GLEngine
    │   │   ├── GLRendererFloat.bundle
[...]

PR #307, opened 2020-11-16
Fix issues with Big Sur

jerekshoe:
This should fix issues (specifically those being mentioned in #274) for anyone using macOS Big Sur, regardless of python version. These changes should be unnecessary once CPython is updated, however this should fix things for the time being.

### DIFF
--- a/pyglet/lib.py
+++ b/pyglet/lib.py
@@ -250,6 +250,16 @@ class MachOLibraryLoader(LibraryLoader):
     @staticmethod
     def load_framework(name):
         path = ctypes.util.find_library(name)
+
+        # Hack for compatibility with macOS > 11.0
+        if path is None and sys.platform == 'darwin':
+            frameworks = {
+                'AGL': '/System/Library/Frameworks/AGL.framework/AGL',
+                'OpenAL': '/System/Library/Frameworks/OpenAL.framework/OpenAL',
+                'OpenGL': '/System/Library/Frameworks/OpenGL.framework/OpenGL'
+            }
+            path = frameworks.get(name)
+
         if path:
             lib = ctypes.cdll.LoadLibrary(path)
             if _debug_lib:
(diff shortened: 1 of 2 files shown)


benmoran56 (approved):
It's not ideal to add hacks of course, but it will be good to get people going again until upstream fixes are ready. Thanks for your work on this!

pyglet/lib.py
```diff
@@ -250,6 +250,16 @@ def find_library(self, path):
     @staticmethod
     def load_framework(name):
         path = ctypes.util.find_library(name)
+
+        # Hack for compatibility with macOS > 11.0
+        if path is None and sys.platform == 'darwin':
```
benmoran56, line 255: Since this method is part of the `MachOLibraryLoader` class, I don't think we need the `sys.platform` check here.

pyglet/lib.py
```diff
@@ -250,6 +250,16 @@ def find_library(self, path):
...
+            frameworks = {
+                'AGL': '/System/Library/Frameworks/AGL.framework/AGL',
+                'OpenAL': '/System/Library/Frameworks/OpenAL.framework/OpenAL',
+                'OpenGL': '/System/Library/Frameworks/OpenGL.framework/OpenGL'
+            }
+            path = frameworks.get(name)
+
         if path:
```
benmoran56, line 263: I suppose the `if path` check is never reached now, since we are explicitly setting it above. If we're planning on removing that hack later, then maybe we can just leave this alone though.

benmoran56 (commented):
I made a small pedantic comment, but don't worry about it. I'll merge this in and fix it on my side.

pyglet/lib.py
jerekshoe: Yeah I wasn't sure about that one, but you're right, it could definitely be removed from both this file and the other file in this pull request

pyglet/lib.py
jerekshoe: Yeah I left it for 2 reasons, the first being that hopefully we will be able to remove this someday. The other being that if for some reason we decided to load a framework other than the 3 in the dictionary above for some reason, I wanted it to behave the same way it would if the dictionary wasn't there.

jerekshoe:
That's fine, I appreciate the feedback!

PR #310, opened 2020-11-18
Purge some Python2 things from the docs

Torxed:
Mainly converted `print '...'` into `print('...')`.
But also the way the exceptions were thrown in `xlib` where `exception.message` no longer exists (?) and type conversion issue.


Torxed:
I apparently had some unstaged changes in the `xlib` exception handling that popped up on the raspberry pi as well.

benmoran56 (approved):
Thanks for this!

PR #313, opened 2020-11-20
Add error code to exception glError exception

electromatter:
The GLException is not sufficient. Sometimes gluErrorString doesn't have an error message for a given error, in which case the message is None which leaves the user high and dry when trying to figure out how to fix the problem. I ran into this issue with GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT, which raised the not-so-helpful GLException(None).

This change adds the GL error code in hex to the error message so the user can look it up manually. Hex for convenience since that's how the GL headers define the error constants.

### DIFF
--- a/pyglet/gl/lib.py
+++ b/pyglet/gl/lib.py
@@ -103,7 +103,7 @@ def errcheck(result, func, arguments):
     error = gl.glGetError()
     if error:
         msg = ctypes.cast(gl.gluErrorString(error), ctypes.c_char_p).value
-        raise GLException(msg)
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

benmoran56:
Thanks @electromatter, this is definitely a useful fix. 

Also, as you may have noticed, the `master` branch is under development at the moment. 
In case you haven't seen it, there are some Framebuffer related classes located at: 
https://github.com/pyglet/pyglet/blob/master/pyglet/image/imagebuffer.py
These have not been completely fleshed out and tested yet, but they may be of interest
already.

PR #316, opened 2020-11-22
GLU removal

einarf:
GLU is deprecated because it mostly interacts with the fixed pipeline.

Tested on:
- [x] Windows
- [x] OS X
- [x] Linux


pyglet/gl/lib.py
```diff
@@ -102,14 +102,22 @@ def errcheck(result, func, arguments):
         raise GLException('No GL context; create a Window first')
     error = gl.glGetError()
     if error:
-        msg = ctypes.cast(gl.gluErrorString(error), ctypes.c_char_p).value
+        # These are the 6 possible error codes we can get in opengl core 3.3+
```
benmoran56, line 105: This is much nicer, thanks!

benmoran56 (dismissed):

benmoran56:
Great work! 
Looks like you got everything on Linux. It works fine as far as I can tell. 
For good measure, I also did a `grep -i glu` of the codebase, and didn't find any stray code fragments.

pyglet/gl/lib.py
einarf: Not really any choice when we don't have `gluErrorString` 😄

benmoran56 (approved):

PR #326, opened 2020-12-10
Implementation of raw keyboard input to fix left and right shift keys

caffeinepills:
Adds raw keyboard input so both shift keys behave as other keys. This is required as `WM_KEYDOWN` and `WM_KEYUP` legacy events will not properly detect shift keys without major hacks.

Still retains shift modifier behavior for key presses, motion, etc.

Also add some missing constants related to input.


benmoran56:
I was able to test this in my Windows 10 VM, and it works as expected. I also confirmed that it's functionally identically to my Linux machine. 

Thanks!

PR #327, opened 2020-12-10
* Set proper default orientation for XAudio2's listener.

caffeinepills:
By default it doesn't have an orientation like the other drivers since it's a client created listener rather than from the driver. This just sets the correct orientation on creation.


benmoran56 (approved):
Thanks! Good to have them all in a consistent state.

PR #328, opened 2020-12-11
XBUTTONS: Add Win32 implementation.

LennyPhoenix:
This adds XButton (Mouse4, Mouse5) support for **windows only**.

I cannot add support for other platforms as I do not have access to machines to test with, so further support will require other contributors to help.

(Credit to @t-o-n-y-p for helping out with the implementation)


benmoran56 (approved):
Thanks Lenny!

PR #333, opened 2020-12-25
Some headless egl work.

mahiuchun:
Want to make sure everything is on the right track. Obviously we still need to implement methods in `HeadlessWindow` etc.

I moved `egl.py` to its own module to avoid circular. `libs` might be an even better location?


benmoran56 (approved):
Looks great!

benmoran56 (approved):

PR #336, opened 2021-01-03
_keyboard_exclusive and _mouse_exclusive alignment for all platforms

jerekshoe:
BaseWindow's defines `_mouse_exclusive` (it should also define `_keyboard_exclusive` but doesn't). This property isn't used by any of BaseWindow's sub-classes (except for XlibWindow). This PR makes all BaseWindow derivatives use `_mouse_exclusive` and `_keyboard_exclusive` as defined in BaseWindow


benmoran56 (approved):
Nice cleanup.

PR #338, opened 2021-01-03
Using super for set_size on window classes

jerekshoe:
### Purpose
This PR uses super to simplify the implementation of `set_size` for all BaseWindow sub-classes. I also added type hinting to the `set_size` functions

### Changes
* Using type hints for all instances of `set_size` in window classes
* Using super to reduce reused code for `set_size`
* Removed `_window` and `_height` from XlibWindow as these are defined in BaseWindow

### DIFF
--- a/pyglet/window/__init__.py
+++ b/pyglet/window/__init__.py
@@ -1019,8 +1019,10 @@ class BaseWindow(with_metaclass(_WindowMetaclass, EventDispatcher)):
         """
         if self._fullscreen:
             raise WindowException('Cannot set size of fullscreen window.')
-        self._width = max(width, 1)
-        self._height = max(height, 1)
+        if width < 1 or height < 1:
+            raise ValueError('width and height must both be positive integers')
+
+        self._width, self._height = width, height
 
     def get_pixel_ratio(self):
         """Return the framebuffer/window size ratio.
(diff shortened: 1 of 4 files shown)


pyglet/window/__init__.py
```diff
@@ -1017,7 +1017,10 @@ def set_size(self, width, height):
                 New height of the window, in pixels.
 
         """
-        raise NotImplementedError('abstract')
+        if self._fullscreen:
+            raise WindowException('Cannot set size of fullscreen window.')
+        self._width = max(width, 1)
+        self._height = max(height, 1)
```
jerekshoe, line 1023: I'm not sure if this is worth doing here or if it should just be:
```python
self._width = width
self._height = height
```
The cocoa version did it with max(), but xlib just takes the user input. Win32 doesn't update these values on calls to set_size at all

benmoran56: We definitely want to do the `max` call here to be safe. Some OpenGL drivers will error out with zero values for width and height (such as viewport settings). It's common for users to use the `on_resize` method to update these things, so we don't want to allow a zero value.

jerekshoe: Ok cool, I'll go ahead and leave that in then

jerekshoe: The other option is to raise an exception if the user entered a value less than 1. Which is the preferred method?

benmoran56: That's a good point. Raising a `WindowException` (or similar) would also be acceptable.

benmoran56 (approved):
Much better to have the logic in the base class. Thanks!

PR #339, opened 2021-01-04
Update set_minimum_size and set_maximum_size window functions

jerekshoe:
### Changes
* Promote `_minimum_size` and `_maximum_size` attributes to `BaseWindow` class
* Use `super()` to simplify code for `set_minimum_size()` and `set_maximum_size()` function
* Enforce restriction on `set_minimum_size()` and `set_maximum_size()` functions that `width` and `height` parameters must be positive integers


benmoran56 (approved):
Thanks! Much more consistent.

PR #340, opened 2021-01-08
Promote get_size to BaseWindow

jerekshoe:
## Purpose
Promotes `get_size` function in XlibWindow and Win32Window to BaseWindow. CocoaWindow is left untouched besides adding type hints

### DIFF
--- a/pyglet/window/cocoa/__init__.py
+++ b/pyglet/window/cocoa/__init__.py
@@ -36,7 +36,6 @@
 from ctypes import *
 
 import pyglet
-from pyglet import gl
 from pyglet.window import BaseWindow
 from pyglet.window import MouseCursor, DefaultMouseCursor
 from pyglet.event import EventDispatcher
--- a/pyglet/window/cocoa/__init__.py
+++ b/pyglet/window/cocoa/__init__.py
@@ -34,10 +34,11 @@
 # ----------------------------------------------------------------------------
 
 from ctypes import *
+from typing import Tuple
 
 import pyglet
 from pyglet import gl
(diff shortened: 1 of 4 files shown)


pyglet/window/cocoa/__init__.py
```diff
@@ -395,7 +395,7 @@ def set_location(self, x, y):
         origin = cocoapy.NSPoint(x, screen_height - y - rect.size.height)
         self._nswindow.setFrameOrigin_(origin)
 
-    def get_size(self):
+    def get_size(self) -> Tuple[int, int]:
```
jerekshoe, line 398: I'm not sure about the intended functionality of `get_size`. I see two options:
1) Return the self._width and self._height as we are doing in the other classes
2) Return the size of the window as reported by each os specific library
#1 is simpler and platform agnostic, #2 is in theory more accurate

benmoran56: I believe 1) would be the way to go. We have a more recent `get_framebuffer_size` method, which will return the _actual_ size if applicable.

jerekshoe: So it's safe to remove this function like I did for the others?

benmoran56: Yes I believe it should be, but we will want to have someone with a Mac confirm it. I think we can merge it for now, and then revert later after testing.

jerekshoe: Ok sounds good

benmoran56 (approved):

PR #342, opened 2021-01-08
Promote set_vsync to BaseWindow

jerekshoe:
## Purpose
Promotes set_vsync to BaseWindow class

## Notes
I didn't make any changes to Win32Window because it doesn't set `_vsync` and it does some weirdness where it sets `vsync` (the argument) to 0. I didn't really want to mess with that without understanding more of what was going on

### DIFF
--- a/pyglet/window/__init__.py
+++ b/pyglet/window/__init__.py
@@ -1133,7 +1133,7 @@ class BaseWindow(with_metaclass(_WindowMetaclass, EventDispatcher)):
         """
         raise NotImplementedError('abstract')
 
-    def set_vsync(self, vsync):
+    def set_vsync(self, vsync: bool) -> None:
         """Enable or disable vertical sync control.
 
         When enabled, this option ensures flips from the back to the front
@@ -1154,7 +1154,7 @@ class BaseWindow(with_metaclass(_WindowMetaclass, EventDispatcher)):
                 If True, vsync is enabled, otherwise it is disabled.
 
         """
-        raise NotImplementedError('abstract')
+        self._vsync = vsync
 
     def set_mouse_visible(self, visible=True):
         """Show or hide the mouse cursor.
--- a/pyglet/window/win32/__init__.py
+++ b/pyglet/window/win32/__init__.py
@@ -339,8 +339,9 @@ class Win32Window(BaseWindow):
                 # This seems odd, as vsync is normally a bool
                 # Win32Context.set_vsync casts vsync to an int anyways
                 # so this should probably be changed
-                vsync = 0
+                vsync = False
 
+        super().set_vsync(vsync)
         self.context.set_vsync(vsync)
 
     def switch_to(self):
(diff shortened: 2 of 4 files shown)


pyglet/window/__init__.py
```diff
@@ -1133,7 +1133,7 @@ def maximize(self):
         """
         raise NotImplementedError('abstract')
 
-    def set_vsync(self, vsync):
+    def set_vsync(self, vsync: bool) -> None:
```
jerekshoe, line 1136: Not sure if this should default to True maybe?

pyglet/window/__init__.py
```diff
@@ -1154,7 +1154,7 @@ def set_vsync(self, vsync):
                 If True, vsync is enabled, otherwise it is disabled.
 
         """
-        raise NotImplementedError('abstract')
+        self._vsync = vsync
```
jerekshoe, line 1157: I could include:
```python
if self._context:
    self._context.set_vsync(vsync)
```
but there are a few questions:
1) Is checking for existing of self._context necessary. Only CocoaWindow performs this check
2) Is it a good idea to include the extra logic here

pyglet/window/__init__.py
benmoran56: I'd probably leave it as-is, just to force users to be explicit. 

However, it _might_ make sense to just turn some of these things into `@property`s instead of using the `set_xxxx` methods. (Since it's for v2.0 it's a chance to modernize a bit).  What do you think?

jerekshoe: I've thought about it and I'm thinking that I would like to clean the code up first, then look at making many of these into properties

jerekshoe: I'm thinking that things like this and mouse_exclusive/keyboard_exclusive would be prime candidates for properties

benmoran56: That sounds reasonable.

pyglet/window/__init__.py
benmoran56: There seem to be a lot of similar checks in the CocoaWindow, but I'm not sure if they're really necessary. The context is set during `__init__`, so this shouldn't be any different from other Windows. I think we should try to confirm on a Mac.

jerekshoe: I'll test it on my mac in a bit

jerekshoe: Context seems to be there from my testing on macOS. Can't see anything in the code that would suggest it wouldn't be there, so I'm going to remove the check for it

pyglet/window/win32/__init__.py
```diff
@@ -336,6 +336,9 @@ def set_vsync(self, vsync):
         if not self._fullscreen:
             # Disable interval if composition is enabled to avoid conflict with DWM.
             if self._always_dwm or self._dwm_composition_enabled():
+                # This seems odd, as vsync is normally a bool
+                # Win32Context.set_vsync casts vsync to an int anyways
+                # so this should probably be changed
                 vsync = 0
```
benmoran56, line 342: It's probably best to just set this to a `bool` (False) for consistency.

benmoran56 (approved):
Much cleaner. Thanks!

PR #343, opened 2021-01-11
Fixes issue with PygletTextView

jerekshoe:
## Purpose
Fixes an issue created in #340 pyglet/window/cocoa/\_\_init\_\_.py (L50). This PR should allow the window to function properly again in macOS

### DIFF
--- a/pyglet/window/cocoa/pyglet_view.py
+++ b/pyglet/window/cocoa/pyglet_view.py
@@ -36,6 +36,7 @@ from pyglet.window import key, mouse
 from pyglet.libs.darwin.quartzkey import keymap, charmap
 
 from pyglet.libs.darwin import cocoapy
+from .pyglet_textview import PygletTextView
 
 
 NSTrackingArea = cocoapy.ObjCClass('NSTrackingArea')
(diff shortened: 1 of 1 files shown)


pyglet/window/cocoa/pyglet_view.py
```diff
@@ -117,7 +118,6 @@ def initWithFrame_cocoaWindow_(self, frame, window):
         # "Option-e", "e" if the protocol isn't implemented.  So the easiest
         # thing to do is to subclass NSTextView which *does* implement the
         # protocol and let it handle text input.
-        PygletTextView = cocoapy.ObjCClass('PygletTextView')
```
jerekshoe, line 120: I'm not entirely sure if this is correct or not. Originally PygletTextView was being imported in pyglet/window/cocoa/__init.py__. Let me know if there is a better way to handle this

benmoran56: I don't see an issue with this as long as it works.
It doesn't actually seem to be used in `__init__.py` after the import, so I'm not sure why it was being imported there.

jerekshoe: This should be good to go then

benmoran56 (approved):
Thanks!

PR #351, opened 2021-02-06
Arc Refactoring

kozbot:
Changed so Arc by default is open, and segments is the number of segments along the arc path.  Added closed argument to draw chord if true.  Added rotation which equates to the angular offset to the start point.

image


benmoran56 (approved):
Thanks for this! 
I'll make some small tweaks after merging it it.

PR #353, opened 2021-02-07
New gengl module using opengl-registry

einarf:
We now generate two modules:
* `gl.py` containing 4.6 core enums and functions
* `gl_compat.py` containing 4.6 compatibility enums and functions

We use the core version by default and users can opt in to compatibility mode by using `gl_compat` instead. Using compatibility mode will cause issues on certain platforms and drivers where only core is available.


einarf:
I kept the old `gengl.py` around because the new one doesn't generate `glx`, `agl` or `wgl` etc. It only generates `gl` and `gl_compat`

einarf:
TODO:
- [ ] Update docs
- [ ] Put requirements somwhere (opengl-registry)
- [ ] Remove old gl modules

einarf:
I think this is ready to merge now unless you find something. It worked fawlessly on MacOS.

benmoran56:
Just cloned your branch, and it seems to work fine on my Linux machine as well. Great work! 
I'm going to give the code a quick manual review, then merge it in.

einarf:
I think `gengl.py` is not as smooth as it can be, but at least it's not horrible.

benmoran56 (approved):

PR #359, opened 2021-02-19
Optional DirectWrite font renderer

caffeinepills:
I posted a while ago looking for feedback on fonts, and I have done some implementation create a DirectWrite as an alternative and updated font renderer for Windows . I have been sitting on it for a while, but most of the implementation and testing are done, but now am looking for any one willing to test their fonts and give feedback. Maybe an extra set of eyes on anything missed, discussion on the changes, or adjustments to improve things.

For now, I have created a separate pyglet option for enabling this (Windows 7+ only). This option is called "advanced_font_features". (See examples/text/advanced_font.py). This will force your pyglet to use the new font renderer. 

I made this because some fonts may have a lot of features that we simply don't take advantage of, and can lead to cut off, missing, or just ugly looking results without them.

Below is a list of some of the features it takes advantage of.

Pros
- Able to specify more font options such as stretching and different font weights (bold, black, light, semilight, etc)
- Supports OpenType features such as kerning, substitutions, alternates, swashes, and more. (some features are Windows 10 creators+)
- Expanded Unicode icons such as emoticons or icons.
- Supports colored glyphs such as colored emoticons or icons. (Windows 8.1+)
- Optional render to a single texture.
- More anti-alias rendering options such as ClearType, Greyscale, or none can be specified.
- Will now give you an error if your custom font name was not found in the collection instead of silently ignoring it and choosing the default font. 

Cons
- Rendering performance is much slower on initial glyph rendering. Speed is closer to GDI once glyphs are cached.
	Some performance examples rendering two labels, initially and then again to show cached performance:

```
	DirectWrite:
	LABEL ONE. LENGTH: 14 TIME: 0.14654
	LABEL ONE. LENGTH: 14 TIME: 0.0033509000000000455
	LABEL TWO. LENGTH: 911 TIME: 0.4112479
	LABEL TWO. LENGTH: 911 TIME: 0.2442897999999999

	GDI:
	LABEL ONE. LENGTH: 14 TIME: 0.04658639999999997
	LABEL ONE. LENGTH: 14 TIME: 0.0021353000000000066
	LABEL TWO. LENGTH: 911 TIME: 0.31095700000000004
	LABEL TWO. LENGTH: 911 TIME: 0.30949589999999993
```
- Bidirectional fonts not supported (possibly supported on render to texture, but untested) 

Limitations:
- Some options are version specific to Windows. 
- Color glyphs adhere to the color of the text. There isn't an intuitive way to omit certain glyphs from being colored.
image

Here is a comparison (DirectWrite left, GDI Right) using the Segoe UI and Arial font with various types.
image

Here is an example of the Zapfino font, which has a lot of large characters, kerning, and some substitution. Looks much more correct with DirectWrite.
image


benmoran56:
Thanks for this. It all looks pretty good. I should be able to merge it in soon, and then confirm on my local box.

benmoran56 (approved):
Looks good to me!

PR #366, opened 2021-03-26
DirectWrite fixes

caffeinepills:
Support for glyph offsets which fixes issues with diacritics and other special characters not being rendered.
Fix for advance cache being shared among instances and a crash. This should solve some spacing issues as well.
Fixes some LSB issue and storing LSB in glyph.
Fixes a crash when using render to image.


benmoran56 (approved):

benmoran56:
Thanks, and sorry for the delay!

PR #376, opened 2021-04-20
BUG: Pad buffers for FFmpeg

### LINKED ISSUE #222: BUG: Simple video example does not work
Based on `git bisect`, 6a409810cdb32bd1d73bd9b4aa3234c50372709d broke this minimal video example: it fails (black screen) on that commit but passes (plays video) on its parent 27ed0d9893372740cddccfee7d39f06c3d88638e. This example code uses [this video file](https://github.com/LABSN/expyfun-data/blob/master/video/example-video.mp4) but I imagine it would work for any:

```
import os
import pyglet

path = os.path.dirname(__file__) + '/example-video.mp4'
window= pyglet.window.Window()
MediaLoad = pyglet.media.load(path)

player.queue(MediaLoad)
player.play()


@window.event
def on_draw():
        player.get_texture().blit(50,50)
    else:
        raise SystemExit(0)


pyglet.app.run()
```
The code seems to have diverged quite a bit so `git revert` does not work. But selective `git checkout` from the parent shows it was the changes to the `window` code -- adding this before `pyglet.app.run()` causes things to work again on 6a409810cdb32bd1d73bd9b4aa3234c50372709d:
```
@window.event
def on_resize(width, height):
    from pyglet import gl
    gl.glMatrixMode(gl.GL_PROJECTION)
    gl.glLoadIdentity()
    gl.glOrtho(0, width, 0, height, -1, 1)
    gl.glMatrixMode(gl.GL_MODELVIEW)
```
But on `master` this is an OpenGL error. Any ideas how to fix this properly?

larsoner:
Closes #222

### DIFF
--- a/pyglet/media/codecs/ffmpeg.py
+++ b/pyglet/media/codecs/ffmpeg.py
@@ -559,7 +559,7 @@ class FFmpegSource(StreamingSource):
         if self.audio_format:
             # Buffer 1 sec worth of audio
             nbytes = ffmpeg_get_audio_buffer_size(self.audio_format)
-            self._audio_buffer = (c_uint8 * nbytes)(*([0] * nbytes))
+            self._audio_buffer = (c_uint8 * nbytes)()
 
         self.videoq = deque()
         self._max_len_videoq = 50  # Need to figure out a correct amount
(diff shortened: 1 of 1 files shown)


larsoner:
I originally had the bytes initialized with zeros (the docs say that the extra ones should be at least), but it slowed down the video decoding. There is probably a better way to do it with ctypes but I'm not sure offhand

PR #383, opened 2021-05-06
Add warning when a detected GL config is less than 2

### LINKED ISSUE #375: Warn user when driver appears to be Generic GDI on windows + ways of addressing it
This is partly @benmoran56 's idea from discord:
> [ I was wondering though if it would make any sense for pyglet to throw a warning if the renderer is "Generic GDI" :thinking:](https://discord.com/channels/438622377094414346/477394331112701952/833175648712392735)

 It was inspired by #374 , which turned out to be an issue with older intel integrated hardware that doesn't report GL capabilities properly under windows 10. To address this, the following steps could be taken:
1. Add notes to the documentation about what a generic GDI driver means, how to diagnose the cause, and how it can be addressed,
2. Display a warning when a generic GDI driver is detected and recommend the user check the documentation for ways to diagnose the cause and address it

If we also have in-depth access to hardware information, we might be able to save users and maintainers a lot of trouble by building in some checks that specifically detect troublesome generations of intel hardware to suggest the cause is something other than the driver not being installed.

pushfoo:
I think this should close #375 . I'd like someone to check that the changes are in the right place since I wasn't sure to put them.

I used `config.major_version`'s value rather than calling `pyglet.gl.gl_info.have_version` because `gl_info`'s default context may be blank at the point when a window is created on intel chips under windows.

We could also move these changes into `pyglet.canvas.base` under `Screen.get_best_config` since a portion of the code is mostly identical between it and the `Window` constructor. However, that seems like it would spam warning messages every time a rendering context is created, which seems excessive.


pushfoo:
Per the discussion in discord, I could also add some tests for this if you'd like. We could patch the supporting method calls with `unittest.mock.patch`.

pushfoo:
Closing this one as it's easier to make a new PR than completely change source and destination branches.

PR #390, opened 2021-05-13
Fix support for GL warnings

pushfoo:
### Summary of Actions
1. Remove the broken warning from `canvas/base.py`
2. Improve encapsulation for detecting and providing GL version information in `GLInfo` objects
3. Add a working warning for GL < 2 as a finale to `BaseWindow.__init__` after the context is created and set as active
4. Ran minimal window creation on local system with `shadow_window` both on and off

### What's needed before merging
- [ ] Run it on a non-windows system
- [x] Run it against a windows host reporting GL 1.1
- [ ] Figure out ways to test it against examples to make sure it doesn't break things
- [ ] Feedback on code quality
- [ ] Feedback on error message placement

### Additional Concerns / Comments
I looked for tests of Context, GLInfo, and Window, but didn't see meaningful ones related to this issue. Should I add them given this is a maintenance branch?
Also, I might be able to get a windows VM up if Ben can't get to it in the next few days.

If this passes sniff checks, it should close #383

### DIFF
--- a/pyglet/canvas/base.py
+++ b/pyglet/canvas/base.py
@@ -32,8 +32,6 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 # ----------------------------------------------------------------------------
-import warnings
-
 from pyglet import gl
 from pyglet import app
 from pyglet import window
--- a/pyglet/canvas/base.py
+++ b/pyglet/canvas/base.py
@@ -194,17 +194,7 @@ class Screen:
         if not configs:
             raise window.NoSuchConfigException()
 
-        best = configs[0]
-        if best.major_version < 2:
-            version = best.major_version, best.minor_version
-            warnings.warn(
-                f"OpenGL version lower than 2.0 {version!r}. "
-                f"Many features may not work if batching fails. "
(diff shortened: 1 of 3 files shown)


pushfoo:
Removed the unused import flagged by lgtm-com

pushfoo:
Closing this PR as commit ec7eb763 and @benmoran56's [feedback in discord](https://discord.com/channels/438622377094414346/477394331112701952/842621163535859722) indicate it's not the right solution for 1.5:
> I had some time to go over this today. Your changes weren't bad, but for v1.5 it's probably not a good idea to rewrite the version parsing code. 
I actually tried to do this myself before, and it caused some bug reports shortly after.
Also, despite the fact that glGetIntegerv with GL_MAJOR_VERSION/GL_MINOR_VERSION 
is the right way to do it on modern OpenGL, it turns out to cause issues.
This is because these calls will fail on older cards. Therefore, you can't actually use them to check
if the driver supports a specific version or not. If not, it'll just crash.
The crusty old glGetString on the other hand, does work on all machines.
I came up with a solution that I think works OK.
This uses the existing gl_info.have_version(2, 0) function, which works fine. The trick is that it needs to be placed after the call to self._create() in the Window initializer. 
I think that was the missing piece that we didn't see before. I had to look through the code manually to understand the order.

Issue #352, opened 2021-02-06
Error -13 returned loading media from io.Bytes using FFmpegDecoder


mcgoober:
Problem:
on Windows 10, using the pyglet-ffmpeg module and when calling -
pyglet.media.load(io.Bytes(media_data_as_byte_array), streaming=False, decoder=pyglet.media.codecs.ffmpeg.FFmpegDecoder)
I get the following error: pyglet.media.codecs.ffmpeg.FFmpegException: avformat_open_input in ffmpeg_open_filename returned an error opening file C:\Users\user\AppData\Local\Temp\tmppgpnk36v Error code: -13

Debugging the problem:
When FFmpegSource is constructed in ffmpeg.py it creates a named tempfile which the media_data is written to. However I observed that the tempfile is 0 bytes, and cannot be opened by ffmpeg because python still has it open. Reading the docs about NamedTemporaryFile, it mentions that on Windows the tempfile cannot be opened by a second process unless the file is first closed by Python (which would remove the tempfile).
If I change the NamedTemporaryFile in the constructor for FFmpegSource in ffmpeg.py to not delete the tempfile on close, write the data and then close the tempfile before ffmpeg loads the data, then I can load the media and play it correctly (but leaves the tempfile on disk, which could be removed with a file delete later).
```
if file:
            file.seek(0)
            self._tempfile = tempfile.NamedTemporaryFile(delete=False)
            self._tempfile.write(file.read())
            self._tempfile.close()
            filename = self._tempfile.name
```

benmoran56:
Interesting, I was unaware that Windows had that limitation with tempfiles. 
Maybe the easiest solution would just be to move the close logic into `__del__` for the FFMpegSource?

```py
def __del__(self):
    ....
    if self._tempfile:
        self._tempfile.close()
   ....
```

mcgoober:
Thanks Ben, unfortunately we need Python to close the file handle before FFMPEG can open it. The only way to do that with NamedTemporaryFile is to call close(), so FFmpegSource needs to take control of the file deletion.
Basically self._tempfile_close() is needed before ffmpeg_open_filename(). 
I ended with my changes above, plus this mod to __del__.

```python
def __del__(self):
    ....
    if self._file and ffmpeg_close_file is not None:
        ffmpeg_close_file(self._file)
    if hasattr(self, '_tempfile'):
        os.remove(self._tempfile.name)
```
[...]

PR #401, opened 2021-06-13
Fix bug when opening FFmpegStreams under Windows

### LINKED ISSUE #400: Bug when opening FFmpegStreams under Windows
In `pyglet.media.codecs.ffmpeg.py` in line 464 a NamedTemproraryFile is created. The file is then left open, and the filename is later given (via `asbytes_filename`) to `ffmpeg_open_filename` in line 468, which will try to open and read from that file again.

However, according to the official Python documentation, this does not work on Windows: "Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later)." (https://docs.python.org/3.8/library/tempfile.html#tempfile.NamedTemporaryFile).

The solution is to instantiate `NamedTemporaryFile` with the `delete=False` parameter in line 464, and then close it after writing to it (and delete the file after closing it in the `__del__` method).

fladd:
This should fix #400, by creating a `NamedTemporaryFile` that is not deleted when being closed, closing it after writing to it so it can be opened again by `ffmpeg_open_filename`, and then deleting it in the `FFmpegSource` class destructor.

### DIFF
--- a/pyglet/media/codecs/ffmpeg.py
+++ b/pyglet/media/codecs/ffmpeg.py
@@ -36,6 +36,7 @@
 """Use ffmpeg to decode audio and video media.
 """
 
+import os
 import tempfile
 from ctypes import (c_int, c_uint16, c_int32, c_int64, c_uint32, c_uint64,
                     c_uint8, c_uint, c_double, c_float, c_ubyte, c_size_t, c_char, c_char_p,
@@ -461,8 +462,10 @@ class FFmpegSource(StreamingSource):
 
         if file:
             file.seek(0)
-            self._tempfile = tempfile.NamedTemporaryFile(buffering=False)
+            self._tempfile = tempfile.NamedTemporaryFile(buffering=False,
+                                                        delete=False)
             self._tempfile.write(file.read())
+            self._tempfile.close()
             filename = self._tempfile.name
 
         self._file = ffmpeg_open_filename(asbytes_filename(filename))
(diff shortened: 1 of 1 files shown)


benmoran56:
Unfortunately we discussed doing something like this, and ultimately decided against it. 
There is no elegant way to guarantee the temp files will be cleaned up if the program terminates abruptly. Since pyglet is used a lot in the scientific community, and other cases where this might go unnoticed, it didn't seem like a good tradeoff. 

If you have a look at the bottom of this issue here, you can see some local code for a workaround: 
https://github.com/pyglet/pyglet/issues/352

The next point release of pyglet will raise an error if trying to pass in a file-like object. After that, it's up to someone to potentially extend the FFmpegDecoder itself to handle it without the tempfile hack.

fladd:
I am not sure I understand the reasoning. You are afraid that the tempfile would not be deleted in case of a crash, even when registering the deletion with atexit (which actually gets called when Python crashes)? It seems to me that this is also not guaranteed (and maybe even less so) in the current scenario either.

Sure, sub-classing FFmpegStream will work for me, thanks. Just thought this would be better fixed on the Pyglet side.

benmoran56:
The tempfile workaround was a bit of a hack, and really isn't good practice for us to be creating files without the users explicit knowledge. The `atexit` module was considered, but it's not guaranteed to be called. For example:

```py
import atexit
import ctypes
import pyglet


def atexitfunc():
    print("Called atexit")


def cause_segfault(dt):
    ctypes.string_at(0)


atexit.register(atexitfunc)

pyglet.clock.schedule_once(cause_segfault, 0.1)
pyglet.app.run()
```
One tempfile is not a big deal, but pyglet is commonly used by other project which may be running dozens or hundreds of iterations of some things. 

The ideal solution would be to update the FFmpegDecoder module to support the data stream directly, but that would take a bit of work.

benmoran56:
Closing this for now, as per our previous discussions.

Issue #408, opened 2021-06-30
GUI widget - allow control through code


samuelznewton:
I think it would be helpful if the GUI widgets could be controlled through code.  So, for instance, if I have a toggle button, I should be able to toggle its state without a user actually clicking on it.  Or if I have a slider, I should be able to write code to set its value to 0.7 without a user touching it.  My understanding is that this should just be a few simple changes to the widgets.py file, so I could try coding it tomorrow if you'd like me to.

The only thing I'm not sure of is whether it would be better for that to dispatch any associated events or not.  I figure a boolean parameter can solve that, but I'm not sure which would be a better default.

benmoran56:
I can see the use of that. Perhaps a general `value` property on the WidgetBase class? 
Something like: 
```py

    @property
    def value(self):
        raise NotImplementedError

    @value.setter
    def value(self, value):
        raise NotImplementedError
```

My thought is that these should _not_ dispatch the events. It doesn't seem to make sense to dispatch the "on_press/on_release" events, for example, when no user interaction has taken place.  Instead, I can see this code as being a way to set defaults, or restore values to a previous session. 
The value property should however, update the appropriate graphics as you would expect. 

Please feel free to give it a try! If this is for the current stable release, make sure to target the `pyglet-1.5-maintenance` branch.

samuelznewton:
Okay, I gave it my best shot!  I've never actually contributed to an open-source project before, so please let me know if I've done anything wrong so I can fix it.

benmoran56:
Thanks for the PR! 

Just to follow up on the discussion, I can see the use of a way to "disable" widgets.
Let me know if you have any other ideas, but I can think of two ways to make that work: 
1. Have the Widget's event handlers do an additional check for `if _enabled` or something similar. 
2. Remove the Widget from receiving the Window events, effectively disabling it.

samuelznewton:
Method 1 is what I had in mind from the start, so that's what I did.  Thanks!

samuelznewton:
2 contributions from 1 issue.  What a bargain!

Thanks for your help and patience!

PR #410, opened 2021-06-30
Added getters and setters for widgets

samuelznewton:
Per [issue 408](https://github.com/pyglet/pyglet/issues/408), this pull adds a value property to pyglet.gui widgets.  It must be implemented separately for each widget, and it is currently implemented for all existing widgets.  I've done a little bit of testing and it seems to work properly.  This is my first time contributing to an open-source project, so I'm really just hoping I've done everything right.  If not, please let me know, and I'll be glad to try to make things right!

### DIFF
--- a/pyglet/gui/widgets.py
+++ b/pyglet/gui/widgets.py
@@ -73,6 +73,17 @@ class WidgetBase(EventDispatcher):
     @property
     def aabb(self):
         return self._x, self._y, self._x + self._width, self._y + self._height
+    
+    # The value property should be overridden in such a way that the widget
+    # can be 'activated' or changed without requiring user input.
+    # The normal events for control activation should NOT be triggered.
+    @property
+    def value(self):
+        raise NotImplementedError("Value depends on control type!")
+    
+    @value.setter
+    def value(self, value):
+        raise NotImplementedError("Value depends on control type!")
 
     def _check_hit(self, x, y):
         return self._x < x < self._x + self._width and self._y < y < self._y + self._height
@@ -124,6 +135,17 @@ class PushButton(WidgetBase):
         self._sprite = pyglet.sprite.Sprite(self._depressed_img, x, y, batch=batch, group=bg_group)
 
         self._pressed = False
+
+    # Override
+    @property
+    def value(self):
+        return self._pressed
+    
+    # Override
+    @value.setter
+    def value(self, value):
+        self._pressed = value
+        self._sprite.image = self._pressed_img if self._pressed else self._depressed_img
 
     def update_groups(self, order):
         self._sprite.group = OrderedGroup(order + 1, self._user_group)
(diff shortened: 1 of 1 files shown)


pyglet/gui/widgets.py
```diff
@@ -73,6 +73,17 @@ def height(self):
     @property
     def aabb(self):
         return self._x, self._y, self._x + self._width, self._y + self._height
+    
+    # The value property should be overridden in such a way that the widget
+    # can be 'activated' or changed without requiring user input.
+    # The normal events for control activation should NOT be triggered.
+    @property
+    def value(self):
```
benmoran56, line 81: Rather than having all of the block comments, we should probably just add in the doc strings. Something like: 
"""Query or set the Widget's value.

This properly allows you to set the value of the Widget directly, 
without any user input. This can be useful to restore the Widgets
to a previous state. Note that events are not dispatched when
changing the value directly. 
"""

pyglet/gui/widgets.py
```diff
@@ -125,6 +136,17 @@ def __init__(self, x, y, pressed, depressed, hover=None, batch=None, group=None)
...
 
+    # Override
+    @property
+    def value(self):
+        return self._pressed
+    
+    # Override
+    @value.setter
+    def value(self, value):
+        self._pressed = value
```
benmoran56, line 147: It might be a good idea to add assertions in the setters, to prevent strange bugs. 
Something like: 
```py
assert type(value) is bool, "Value must be True or False."
or
assert type(value) in (int, float), "Value must be an int or float."
```

benmoran56 (changes requested):
Looks good overall! I made some inline comments, and we should probably get rid of the # comments, but other than that you're on the right track. 

I think however, that the PushButton probably won't work well with this, because it's a momentary Widget. Instead, we can move those property/setter down to the ToggleButton, where they would work as expected. We

benmoran56:
Also, congratulations on your first contribution!

samuelznewton:
> I think however, that the PushButton probably won't work well with this, because it's a momentary Widget. Instead, we can move those property/setter down to the ToggleButton, where they would work as expected. We

I see your point, but that's only if this is being used for the one use case you have in mind: restoring a previous state.  However, I think there are more uses than that.  I've actually brought up this issue for a different use case, where an event in my program triggers the change of a boolean - a boolean normally controlled by a button.  In my case, yes, it's a ToggleButton.

But imagine that you're trying to program a game where the player has to push some on-screen buttons.  And suppose you wanted to create a tutorial, and you show what happens when you press certain buttons.  Then it makes perfect sense to control a PushButton through code, I would think.

Although, in order to make this more effective, it would probably be a good idea to allow a developer to disable a Widget - perhaps a boolean property.  This, too, would have more uses than just the one I described.  I could actually use a disable feature in my program, as well.

samuelznewton:
I added the docstring to the value property, removed all of my # comments, and added assertions.  I did not move the value property definition out of PushButton into ToggleButton, though I can if you still feel that's best.

If you would like a disable feature, I can probably add that, too.  Though, I don't know whether that would be something to add to this pull request or something totally separate, so if you do want that, please let me know how to proceed.

benmoran56:
The changes look good!

I understand your point about the Push Button widget. That does make sense. Lets leave the potential "disable" discussion for a separate PR, so I think we can merge this in for now.

PR #414, opened 2021-07-02
Enable/Disable pyglet.gui Widgets

samuelznewton:
Per discussion in [issue 408](https://github.com/pyglet/pyglet/issues/408) and [pull request 410](https://github.com/pyglet/pyglet/pull/410), added the ability to enable and disable pyglet.gui Widgets.  The base class only has a new variable, `_enabled`, and a new property, `enabled`.  The subclasses all honor this property by checking it at the start of event handlers and returning if it is False.

Something I've noticed while looking at this PR's commits is that it has the commits from the last PR, then the re-merge I did, then the actual new commit.  Is that a problem?  Is there a better way to go about this so that those old commits don't show up in PR?  I can create a new PR if needed.

### DIFF
--- a/pyglet/gui/widgets.py
+++ b/pyglet/gui/widgets.py
@@ -73,6 +73,17 @@ class WidgetBase(EventDispatcher):
     @property
     def aabb(self):
         return self._x, self._y, self._x + self._width, self._y + self._height
+    
+    # The value property should be overridden in such a way that the widget
+    # can be 'activated' or changed without requiring user input.
+    # The normal events for control activation should NOT be triggered.
+    @property
+    def value(self):
+        raise NotImplementedError("Value depends on control type!")
+    
+    @value.setter
+    def value(self, value):
+        raise NotImplementedError("Value depends on control type!")
 
     def _check_hit(self, x, y):
         return self._x < x < self._x + self._width and self._y < y < self._y + self._height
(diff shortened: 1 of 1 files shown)


pyglet/gui/widgets.py
```diff
@@ -204,25 +221,33 @@ def update_groups(self, order):
         self._sprite.group = OrderedGroup(order + 1, self._user_group)
 
     def on_mouse_press(self, x, y, buttons, modifiers):
+        if not self._enabled:
+            return
         if not self._check_hit(x, y):
```
benmoran56, line 226: In the interest of keeping the module from getting too bloated, it would be better to combine the two "early out" checks. 
```py
if not self._enabled or not self._check_hit(x, y):
```

benmoran56 (changes requested):
I think this will be a fine addition.

benmoran56:
Just FYI, there were some other Widget changes that just got pushed in. I don't think it'll cause any conflict with your open PR, but my appologies if so.

pyglet/gui/widgets.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, x, y, width, height):
         self._height = height
         self._bg_group = None
         self._fg_group = None
+        self._enabled = True
```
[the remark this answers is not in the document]
samuelznewton: Yeah, you're right, I tend to future-proof a bit too much sometimes.  I'll make that change.

pyglet/gui/widgets.py
samuelznewton: Oh yikes did I actually do that?  That's embarrassing.  Will fix.

benmoran56 (approved):
Looks good!

benmoran56:
Thanks for making those changes.

PR #415, opened 2021-07-03
Polygon shape in shapes.py

Ross-TheBoss:
This adds a polygon shape making use of the GL_POLYGON flag. 

The Polygon class allows for:
- rotation around the first coordinate.
- moving the first coordinate to a different position.

**Potential Issues:**
- When multiple polygons are drawn, they are connected to the start and end point of the first polygon (as shown below).
Polygon issue

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,8 +1,14 @@
 pyglet 1.5.19
 
+Bugfixes
+--------
+- Ensure that the FFmpegDecoder returns the requested number of bytes.
+
 Maintenance
 -----------
 - Re-introduce background threads for refilling Player buffers.
+- Allow directly setting Widget values.
+- Add docstrings for gui.widgets.
 
 
 pyglet 1.5.18
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -68,6 +68,7 @@ If this is your first time reading about pyglet, we suggest you start at
    modules/font
    modules/gl
    modules/graphics/index
+   modules/gui
    modules/image/index
    modules/info
    modules/input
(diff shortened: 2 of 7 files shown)


benmoran56:
Thanks for opening the PR.

For the issue with multiple polygons, it's due to the way pyglet's Batching works. It's unfortunately not possible to use GL_POLYGON without the shapes affecting each other. I don't think this will be what users expect, so it could cause confusion.  

It's a bit wasteful, but I think a Polygon Shape would be better off if we can construct it from triangles or a triangle strip. What do you think?

(Just for reference)
https://pyglet.readthedocs.io/en/latest/modules/graphics/#drawing-modes

einarf:
`GL_POLYGON` will also not work in pyglet 2.0. It definitely needs to be expressed as triangles in some way.

Ross-TheBoss:
Thanks for letting me know why this shape should be expressed as triangles. I wasn't entirely sure how to split the polygon into triangles so I went with the easier implementation before.

benmoran56:
If we limit this to convex polygons, then it shouldn't be too difficult.  We just need to find the approximate center, using the min/max of the x and y coordinates. After that we can iterate over the coordinates similar to how the circle does it. I'm sure there are probably better ways to do it as well, but that should work.

For concave polygons, it's a bit more difficult.

Ross-TheBoss:
[This article](https://www.codeproject.com/articles/44370/triangulation-of-arbitrary-polygons) describes how to split any convex or concave polygon into triangles. For now, I'll try implementing this class to work with only convex polygons.

benmoran56 (approved):
Thanks for the changes. This looks good. I'll test it locally.

benmoran56:
The anchor points weren't matching up correctly, but I was able to fix that locally. I'll merge this in, and then make another push to fix that.

Ross-TheBoss:
Thanks

PR #417, opened 2021-07-07
Add more utility to math module

Cleptomania:
Add some extra utility functions to the Vector and Mat4 classes, as well as adds a Mat3 class with some basic functionality.

### DIFF
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -46,15 +46,19 @@ operations.
 
 import math as _math
 import warnings as _warnings
-
 from operator import mul as _mul
+
+
+def clamp(num, min_val, max_val):
+    return max(min(num, max_val), min_val)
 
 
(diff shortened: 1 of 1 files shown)


pyglet/math.py
```diff
@@ -130,12 +156,45 @@ def __truediv__(self, other):
     def __abs__(self):
         return _math.sqrt(self[0] ** 2 + self[1] ** 2 + self[2] ** 2)
 
+    def cross_product(self, other):
```
einarf, line 159: Could also be named `cross`

benmoran56: I'm in favor of that.  We already just have other things like "dot", instead of "dot_product", so I think it makes sense to keep them consistent.

benmoran56 (approved):
Merging this in, and will refactor it a bit locally.

PR #423, opened 2021-07-20
cocoa: set internal window size in setFramSize

einarf:
#340 broke window size for the cocoa window since the custom `get_size()` was removed and nothing was added to update the internal `width` and `height` values in the `BaseWindow`.

I'm not sure if this is the right way to go, but at least it's somewhat similar to win32 version except that we're not querying the real surface size. Can look into that but just wanted a branch with a fix to get this started because it's a blocker for arcade 2.6.


benmoran56 (approved):
This seems to make the most sense to place it here.

PR #427, opened 2021-07-30
refactoring : add __all__s to __init__s from pyglet.window modules & submodules

### LINKED ISSUE #426: Lack of __all__ causes easily avoidable warnings
PyInstaller's warnings (found at`./build/*/warn-*.txt` where * is the name of your packaged script) :
```python
missing module named pyglet.window.BaseWindow - imported by pyglet.window (top-level), pyglet.window.headless (top-level), pyglet.window.cocoa (top-level), pyglet.window.win32 (top-level), pyglet.window.xlib (top-level)
```
Inspecting the code made it clear that this issue was caused by the lack of `__all__`s in the aforementionned packages' `__init__.py`s

loldebyte:
Closes #426
I have not yet tested if this _actually_ silences the warnings, and expect many other changes could (should?) be included in here but i figured i'd better ask you guys' thoughts about this first.
All kinds of feedback appreciated !

### DIFF
--- a/pyglet/window/__init__.py
+++ b/pyglet/window/__init__.py
@@ -1907,3 +1907,19 @@ else:
 if not _is_pyglet_doc_run:
     pyglet.window = sys.modules[__name__]
     gl._create_shadow_window()
+
+__all__ = [
+    "WindowException",
+    "NoSuchDisplayException",
+    "NoSuchConfigException",
+    "NoSuchScreenModeException",
+    "MouseCursorException",
+    "MouseCursor",
+    "DefaultMouseCursor",
+    "ImageMouseCursor",
+    "Projection",
+    "Projection2D",
+    "Projection3D",
+    "BaseWindow",
+    "FPSDisplay"
+    ]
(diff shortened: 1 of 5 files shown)


benmoran56:
Thanks for the PR, and sorry for the delay!
I had a quick look, and some of the classes don't necessarily need to be included in`__all__`. 

I would also say that it's preferrable to keep things to a single line, where they fit. 
We generally stick to PEP8 formatting, with the exception of allowing 100 (or occasionally 120) columns instead of 80.

If you have any questions, please let me know.

benmoran56 (changes requested):
Sorry for leaving this hanging for so long. It looks good now, but I have a small request. 
To keep things PEP8 compliant, could you remove the spaces around the list items? 
For example: 
```py
[ "BaseWindow" ] ---> ["BaseWindow"]
```
Thanks in advance!

benmoran56 (approved):
I went ahead and made some small PEP8 cleanups. 
I'll merge this is now. Thanks again for your PR.

Issue #428, opened 2021-08-02
Add support for circular segments in the pyglet.shapes module.


Ross-TheBoss:
I think the **parameters 'angle' and 'start_angle'** and the **property 'rotation'** could be added to the **pyglet.shapes.Circle** class to allow for drawing circular segments using it. This would be implented in the same way that the pyglet.shapes.Arc class calculates points along a arc.

![A circular segment](https://upload.wikimedia.org/wikipedia/commons/thumb/d/da/Circle_arc.svg/1200px-Circle_arc.svg.png)

benmoran56:
This sounds useful. As we talked about on Discord, it probably makes the most sense to make a new shape, named `Sector`.

benmoran56:
Closing now, thanks to your pull request!

PR #429, opened 2021-08-03
Added a sector class to shapes.py

Ross-TheBoss:
I added a sector class to the shapes.py file and fixed some typos in the pyglet.shapes.Circle class.


Ross-TheBoss:
Addresses issue #428.

benmoran56 (approved):

PR #430, opened 2021-08-03
Bug fixes and cleanup in shapes.py.

Ross-TheBoss:
This change fixes a **bug in shapes.py** where the **Sector's first and last points of its outer arc** are joined. It also adds the **Sector class to __all__** and removes a temporary todo message.


benmoran56 (approved):
Great! Thanks for the fixes.

PR #431, opened 2021-08-08
Fixed typos and made shapes.py conform more to PEP8.

Ross-TheBoss:
This commit makes the following tweaks:
- Fixing typos.
- Removing a bare try, except clause (I'm not sure what the purpose of it is and it just makes it harder to debug.)
- Removing redundant parenthesis as suggested by caffeine pills on discord.
- Adding spaces after commas in a tuple (PEP8).


flowerbug:
class BorderedRectangle(_ShapeBase) is missing a description of the parameter border.

Ross-TheBoss:
Ok, I've added some descriptions for the border and border_color parameters of the BorderedRectangle class.

Ross-TheBoss:
```python
class Star(_ShapeBase):
    def __init__(self, x, y, outer_radius, inner_radius, num_spikes, rotation=0,
                 color=(255, 255, 255), batch=None, group=None) -> None:
        """Create a star.
```
I'm not sure whether we should keep this return value annotation since it is inconsistant with the rest of the classes.

benmoran56 (approved):

PR #432, opened 2021-08-09
Update win32priority.py

syegulalp:
Completely rewritten version of this experimental module. Modernized code; dropped things that no longer applied.


benmoran56 (approved):
Thank you Serdar!

PR #434, opened 2021-08-11
Transparency V1

DougTheDruid:
Adding window transparency as an option in Pyglet. Utilize DWM to blur the background of windows. Include both a non-click-through and overlay style option for users to utilize. 

Also closes https://github.com/pyglet/pyglet/pull/249 as we implement the same adjusted functionality here.


pyglet/window/__init__.py
```diff
@@ -589,7 +594,7 @@ def __init__(self,
             screen = display.get_default_screen()
 
         if not config:
-            for template_config in [gl.Config(double_buffer=True, depth_size=24),
+            for template_config in [gl.Config(double_buffer=True, depth_size=24, alpha_size=8),
```
benmoran56, line 597: There is a catch-all for the alpha size later, so we probably don't want to enforce that here.

pyglet/window/xlib/__init__.py
```diff
@@ -232,6 +232,9 @@ def _create(self):
             root = xlib.XRootWindow(self._x_display, self._x_screen_id)
 
             visual_info = self.config.get_visual_info()
+            depth = 32 if self.style == "transparent" else 24
```
benmoran56, line 235: I think we need to add the new `overlay` here as well. Something like: 
`depth = 32 if self.style in ('transparent', 'overlay') else 24`

benmoran56 (changes requested):
I made some small suggestions, but look good overall. 

After removing `testpy.py`, we could think about adding examples to `examples/windows` in the repo. We can also move the other Window-related examples into there, such as "multiple_windows.py". What do you think?

pyglet/window/__init__.py
DougTheDruid: Hmm... Running locally, if I dont have this here (or specify explicitly in my config) it does _not_ work for me.

DougTheDruid:
Is there any way we could also include some of the other transparency things into the draw event on the backend? If you look at the overlay example, we have to create a class, some globals, and a method, then call that method every loop (along with the glflush()). Would be nice to not have to do that.

benmoran56:
> Is there any way we could also include some of the other transparency things into the draw event on the backend? If you look at the overlay example, we have to create a class, some globals, and a method, then call that method every loop (along with the glflush()). Would be nice to not have to do that.

Windows have a `flip` method that gets called after each call to `on_draw`. That looks like a potentially good place for it. This can be set in `win32/__init__.py`, since it's not necessarily needed for other OSs. 
It might be a good idea to put that behind an `if` flag, if there is any performance impact to it. If so, we don't want to affect the draw performance for non-transparent Window updates.

DougTheDruid:
Thats super awesome! Thank you for the information. I was able to implement that without any issues. It honestly works so good I am wondering if that example is even necessary? Have any thoughts?

benmoran56 (approved):

benmoran56 (approved):
Looks good to me.  Examples are always good to have, but a proper description in the documentation would be better. 

I'll try to test this latest revision on my Windows VM, and if everything is OK I'll see about merging it in. 

Thanks!

DougTheDruid:
Closes #249

DougTheDruid:
I added a couple of example images in the doc & a line regarding the overlay property. Any other things you'd like me to modify?

benmoran56:
Sorry for the delay.  This works well on my Windows VM.
I made a small change to `Window.__init__` to simplify the alpha value setting. 

On my Linux machine the behavior is a bit different, but I can look into that myself later.  
Basically the OVERLAY style is identical to the TRANSPARENCY style. 
I'll merge this in for now, and open an issue for the Linux side.

DougTheDruid:
>Sorry for the delay.

No worries for the delay, I have been busy also

However, the change you implemented in init prevents this from working correctly on my machine. If I pass in a custom config of `config = Config(double_buffer=True, depth_size=24, alpha_size = 8)` to the window creation it works fine, but I guess I don't see that as ideal to have to do

PR #435, opened 2021-08-12
Missing pyglet in argument line

Ansedar:
line 213     window = pyglet.window.Window(style=window.Window.WINDOW_STYLE_DIALOG) results in "NameError: name 'window' is not defined"


benmoran56 (approved):
Thank you very much!

PR #438, opened 2021-08-16
Fixed repetition of "Can", line 27

AnantTiwari-Naman:
"Can" was repeated twice on line 27, I fixed that.


benmoran56 (approved):
Thank you very much!

PR #451, opened 2021-08-30
check version of pip

anukaal:
checking version of pip while installing pyglet


benmoran56:
Thank you for opening the pull request.

What is this trying to accomplish? `make.py` does not rely on pip.

benmoran56:
Closing due to lack of response, and uncertainty as to what this is for.

PR #452, opened 2021-08-31
Fix some blending and alpha issues

caffeinepills:
Fix blit attribute data not having a default color attached causing alpha to be white (since the default shader adds the colors)
Add blending for ImageMouseCursor.


benmoran56 (approved):
Thanks for catching those.

PR #453, opened 2021-09-01
border_color and rotation property for the BorderedRectangle class.

Ross-TheBoss:
This change adds the property "border_color" to the BorderedRectangle class in shapes.py. This allows border colour to be modified after a BorderedRectangle instance is created.
It also adds the property "rotation" so that the BorderedRectangle class can be rotated.


benmoran56 (approved):
Great, thanks!

PR #455, opened 2021-09-10
Fix for shift event on multiple windows.

### LINKED ISSUE #450: Shift modifier not recognized when there are multiple windows
I noticed that when I create two windows and close the second window, the first window no longer recognizes shift modifiers being pressed. Here is some sample code

```
import pyglet
from pyglet.window import key

window = pyglet.window.Window(width=600, height=600)
window2 = pyglet.window.Window(width=400, height=400)

@window.event
def on_key_press(symbol, modifiers):
    print(f"Modifiers with key press on 1: {key.modifiers_string(modifiers)}")

@window2.event
def on_key_press(symbol, modifiers):
    print(f"Modifiers with key press on 2: {key.modifiers_string(modifiers)}")

@window.event
def on_mouse_press(x, y, buttons, modifiers):
    print(f"Modifiers with mouse press on 1: {key.modifiers_string(modifiers)}")
    
@window2.event
def on_mouse_press(x, y, buttons, modifiers):
    print(f"Modifiers with mouse press on 2: {key.modifiers_string(modifiers)}")

pyglet.app.run()
```

After I close `window2`, `window` no longer responds to `MOD_SHIFT`. It recognizes all other modifiers. But if I close `window` first, `window2` still response to `MOD_SHIFT`. Before closing either window, `MOD_SHIFT` works fine. So it is only the first window to be initialized that's facing this issue.

Am I instantiating multiple windows incorrectly? Any help is appreciated!

Thank you for your help!

Edit: I am using pyglet-1.5.19 in Python 3.9.5 on Windows 10

caffeinepills:
Set shift handler to be lost/gained when focusing Windows to fix multiple windows with shift events.

Tested on latest Windows 10.

Should fix #450

### DIFF
--- a/pyglet/window/win32/__init__.py
+++ b/pyglet/window/win32/__init__.py
@@ -235,12 +235,6 @@ class Win32Window(BaseWindow):
                     _user32.ChangeWindowMessageFilterEx(self._hwnd, WM_COPYGLOBALDATA, MSGFLT_ALLOW, None)
 
                 _shell32.DragAcceptFiles(self._hwnd, True)
-                
-            # Register raw input keyboard to allow the window to receive input events.
-            raw_keyboard = RAWINPUTDEVICE(0x01, 0x06, 0, self._view_hwnd)
-            if not _user32.RegisterRawInputDevices(
-                byref(raw_keyboard), 1, sizeof(RAWINPUTDEVICE)):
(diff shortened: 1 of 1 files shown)


caffeinepills:
This was fixed in another pull.

PR #461, opened 2021-09-15
Label and Layout enhancements

caffeinepills:
This adds a couple enhancements to existing Label and layouts to make it similar to Sprites.

Adds a visibility property. This behaves like Sprite visibility, except this deletes the vertices, then re-adds them. This also lets you change text while it's not visible.

Add opacity property for DocumentLabels. - Even though you can adjust this through color property, it gets cumbersome if all you want to do is adjust the opacity.

Add an update function to update both the x and y position simultaneously on layouts. This is a small performance speedup, since both axis can be adjusted in the same loop. In the future can be added with more functionality, for example, adding a Z axis or rotations.

### DIFF
--- a/pyglet/text/__init__.py
+++ b/pyglet/text/__init__.py
@@ -304,6 +304,26 @@ class DocumentLabel(layout.TextLayout):
                                 {'color': color})
 
     @property
+    def opacity(self):
+        """Blend opacity.
+
+        This property sets the alpha component of the colour of the label's
+        vertices.  With the default blend mode, this allows the layout to be
+        drawn with fractional opacity, blending with the background.
+
+        An opacity of 255 (the default) has no effect.  An opacity of 128 will
+        make the sprite appear translucent.
+
+        :type: int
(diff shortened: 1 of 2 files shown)


jason-bowen-zheng:
Can I rotate the Label?

benmoran56 (approved):
Thanks! These were definitely lacking.

benmoran56:
@jason-bowen-zheng Unfortunately not.  The layout class it too complex to add rotation. 
Instead, it would be better to create a Texture out of a Label, and then rotate that. 
Eventually we need to add an example to the Github on how to do that.

jason-bowen-zheng:
@benmoran56 It seems like a good solution!

caffeinepills:
> Can I rotate the Label?

Not with this merge. However, this is supported in the alternate font renderer (Windows only).

See: https://github.com/pyglet/pyglet/blob/pyglet-1.5-maintenance/examples/text/advanced_font.py

PR #466, opened 2021-09-20
Add animation support

jason-bowen-zheng:
Add support for animation, very simple, some functions haven't implemented yet.

Also fix some mistakes.

When I test this new submodule, I found an error when I exit program(just once):
```
Exception ignored in: <function WeakMethod.__new__.<locals>._cb at 0x7f5cf8ad30>
  File "/usr/lib/python3.9/weakref.py", line 61, in _cb
  File "/home/jason/mcpy/pyglet-tutor/pyglet/event.py", line 356, in _remove_handler
  File "/usr/lib/python3.9/weakref.py", line 77, in __eq__
TypeError: isinstance() arg 2 must be a type or tuple of types
Exception ignored in: <function WeakMethod.__new__.<locals>._cb at 0x7f5cf8aaf0>
  File "/usr/lib/python3.9/weakref.py", line 61, in _cb
  File "/home/jason/mcpy/pyglet-tutor/pyglet/event.py", line 356, in _remove_handler
  File "/usr/lib/python3.9/weakref.py", line 77, in __eq__
TypeError: isinstance() arg 2 must be a type or tuple of types
Exception ignored in: <function WeakMethod.__new__.<locals>._cb at 0x7f5cf8ab80>
  File "/usr/lib/python3.9/weakref.py", line 61, in _cb
  File "/home/jason/mcpy/pyglet-tutor/pyglet/event.py", line 356, in _remove_handler
  File "/usr/lib/python3.9/weakref.py", line 77, in __eq__
TypeError: isinstance() arg 2 must be a type or tuple of types
```
If this PR passed, please add detailed document.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -61,6 +61,7 @@ If this is your first time reading about pyglet, we suggest you start at
    :caption: API Reference
 
    modules/pyglet
+   modules/animation
    modules/app
    modules/canvas
    modules/clock
--- a/doc/modules/animation.rst
+++ b/doc/modules/animation.rst
@@ -3,15 +3,10 @@ pyglet.animation
 
 .. automodule:: pyglet.animation
 
-Classes
-=======
-
 .. autoclass:: AnimationClip
-
    :members:
    :undoc-members:
 
 .. autoclass:: Animation
-
    :members:
    :undoc-members:
--- a/doc/modules/animation.rst
+++ b/doc/modules/animation.rst
@@ -1,0 +1,17 @@
+pyglet.animation
+================
+
+.. automodule:: pyglet.animation
+
+Classes
+=======
+
+.. autoclass:: AnimationClip
+
+   :members:
+   :undoc-members:
+
+.. autoclass:: Animation
+
+   :members:
+   :undoc-members:
--- a/pyglet/animation.py
+++ b/pyglet/animation.py
@@ -47,7 +47,7 @@ def out_of_animation_range(num, from_, to):
         if num <= to:
             return True
     else:
-        if num >= from_:
+        if num >= to:
             return True
     return False
 
@@ -118,6 +118,10 @@ class Animation:
 
     def start(self, count=60):
         """Start animation.
+            
+        :Parameters:
+            `count` : int
+                Update objects ``count`` times in 1s.
         """
         self._started = True
         self._count = count
--- a/pyglet/animation.py
+++ b/pyglet/animation.py
@@ -38,8 +38,6 @@
 .. versionadded:: 1.5.22
 """
 
-import math
-
 from pyglet.clock import schedule_once
 
 def out_of_animation_range(num, from_, to):
--- a/pyglet/animation.py
+++ b/pyglet/animation.py
@@ -1,0 +1,134 @@
+# ----------------------------------------------------------------------------
+# pyglet
+# Copyright (c) 2006-2008 Alex Holkner
+# Copyright (c) 2008-2021 pyglet contributors
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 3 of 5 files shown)


jason-bowen-zheng:
Please change this line when new version releases:

https://github.com/Minecraft-in-python/pyglet/blob/9a12f460fb618fdae22be1bf6c19dec0f65b93ef/pyglet/animation.py#L38

caffeinepills:
Animations currently exist in https://github.com/pyglet/pyglet/blob/pyglet-1.5-maintenance/pyglet/image/animation.py

I don't think two different modules for animation should be added. What functionality does this offer over the existing Animation? Adding functionality to the existing one is probably best.

jason-bowen-zheng:
@caffeinepills I think `pyglet.image.animation` is just for images. But **this** can be used on shapes, sprites, and images. I'll create an example soon.

benmoran56:
It's a good point that we could use a generic Animation system. 
Also, as Caffeinpills pointed out, we don't want to have two duplicate systems.
The current animation system is already fairly generic, but all of the control logic is handled by the Sprite class. 

Currently we have: 
`Frame` -  Data container only
`Animation` - Data container only
`Sprite` - Contains the logic for updating the Animation

The Sprite class contains pause/resume and frame_index properties. 
Maybe this can be seperated into a new class, like:
`AnimationController`

This class can contain all of the animation controls. It can then be subclassed by other classes. For example: 
`class Sprite(EventDispatcher, AnimationController):`

What do you think?

jason-bowen-zheng:
@benmoran56 `AnimationController` is a good idea. But how to implement it? What method does it have?
> this PR of `pyglet.animation` just change attributes of **any** object

caffeinepills:
While I do also like the idea of an animation system (and have implemented my own based on cocos) My only issues with this kind of feature are:
1) A generic animation library vs Sprite animations is going to a problem for clarity unless they can be combined, which could further bloat Sprite for all of the uses cases.
2) Libraries using Pyglet already implement their own animation systems. (See Cocos, which does a good job. For instance `sprite.do(Lerp("x", 0, new_x))` ) So does it need to exist in the first place, since every engine is different?
3) It will quickly grow in scope to requested features. (see Shapes or GUI) (I can see, Lerping, Easing, Reversing, Delaying, Chaining, Paralleling, etc) This then becomes at risk of no one maintaining nor expanding it (or wanting to).
4) If it's not powerful enough then people are just going to not use it and implement their own anyways.

jason-bowen-zheng:
I have another idea: jQuery's animate like:
```python
sprite.animate(x=10, y=10, duration=3, callback=lambda: print("done"))
```
But it only can control one object.

> Maybe I can rename `pyglet.animation.Animation` to `pyglet.animation.AnimationGroup`

I think we should choose one:

- Create a `pyglet.animation`
- Let all classes inherit `AnimationController`(or other name)

jason-bowen-zheng:
I think use #466 (this PR) can make complex animations, but @caffeinepills 's suggestion not.

This PR has following advantages:

1. We needn't add a `AnimationController` to all classes
2. It's very easy to use
3. It can make complex animation

benmoran56:
Sorry for leaving this hanging. 
The idea of a generic animation system is good, but it's best if we postpone it until pyglet 2.x.
There are lots of changes taking place already, so it would be difficult if we try to integrate this now.

jason-bowen-zheng:
@benmoran56  I have read the files you changed. In my opinion, it's a very good example for sprite.

But the point is: can we apply the `AnimationController` to a circle, a button or a label?

Actually we need a sequence of images to create animations, but to smoothly move a circle's position from `posA` to `posB` in 5s is also an excellent feature for the game engine.

This PR is controversial, I should close it and create an issue to express my thought.

PR #467, opened 2021-09-20
Add support for PyOgg decoder

caffeinepills:
> PyOgg provides Python bindings for Xiph.org’s Opus, Vorbis and FLAC audio file formats as well as their Ogg container format.

https://github.com/TeamPyOgg/PyOgg

Their bindings are through ctypes, so this makes it easier for us to support.

Each file type divided into it's own source and is found by extension. `PyOggOpusSource`, `PyOggVorbisSource`, `PyOggFLACSource`

Supports loading by filename or file object. 

If anyone can test on different OS's and make sure playback and seeking both work, I would appreciate it. So far all the tests on Windows work with both file objects and filename loaded on Opus, FLAC, and Ogg.

It's possible this can be made independent from PyOgg in the future by integrating ctypes structures and headers through our own bindings. (If say it stops working properly in future versions and goes unmaintained) However, since PyOgg handles installing the dll's for Windows as well, and is already lightweight as is, it's not necessary at this time.

----

This also removes mkv and ogg support for WMF. On many versions this does not work reliably. There is no way to query or tell if it is supported by the system. Furthermore there is no documentation from Microsoft on how these are supported by certain systems.

### DIFF
--- a/pyglet/media/codecs/__init__.py
+++ b/pyglet/media/codecs/__init__.py
@@ -105,6 +105,11 @@ def add_default_media_codecs():
     except ImportError:
         pass
 
+    try:
+        from . import pyogg
+        add_decoders(pyogg)
+    except ImportError:
+        pass
 
 def have_ffmpeg():
     """Check if FFmpeg library is available.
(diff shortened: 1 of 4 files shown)


benmoran56 (approved):
Thanks! I found a small typo, but I'll fix it on my end after merge.

Issue #278, opened 2020-09-10
Sound synthesis tests fail on big-endian architectures


encukou:
`pyglet.media.synthesis` generates an array of `ctypes.c_short`, but it's then compared to bytes read from a file. I'm not very clear on how this works, but somewhere, the byte order is flipped the wrong way on big-endian machines.
For example:

<details>

```
________________________ FMTest.test_sample_rate_11025 _________________________
self = <tests.unit.media.test_synthesis.FMTest testMethod=test_sample_rate_11025>
    def test_sample_rate_11025(self):
        source = self.source_class(1., sample_rate=11025)
        self._test_total_duration(source)
        if self.source_class is not WhiteNoise:
>           self._test_generated_bytes(source, sample_rate=11025)
tests/unit/media/test_synthesis.py:42: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <tests.unit.media.test_synthesis.FMTest testMethod=test_sample_rate_11025>
source = <pyglet.media.synthesis.FM object at 0x3ff5986df40>
sample_rate = 11025, sample_size = 16
    def _test_generated_bytes(self, source, sample_rate=44800, sample_size=16):
        source_name = self.source_class.__name__.lower()
        filename = "synthesis_{0}_{1}_{2}_1ch.wav".format(source_name, sample_size, sample_rate)
    
        with open(get_test_data_file('media', filename), 'rb') as f:
            # discard the wave header:
            loaded_bytes = f.read()[44:]
            source.seek(0)
            generated_data = source.get_audio_data(source._max_offset)
            bytes_buffer = BytesIO(generated_data.data).getvalue()
            # Compare a small chunk, to avoid hanging on mismatch:
>           assert bytes_buffer[:1000] == loaded_bytes[:1000],\
                "Generated bytes do not match sample wave file."
E           AssertionError: Generated bytes do not match sample wave file.
[...]
E             At index 4 diff: b'j' != b'v'
E             Full diff:
E               (
E             -  b'\x00\x00==vj\xd3~,{/g\\L:2\x07\x1d-\x0eS\x051\x01\x0c\x00\xf7\xff'
E             ?               ^    ^ ^ ^  ^ ^      ^^    ^    ^    ^      ^^    ----
[...]

PR #469, opened 2021-09-21
Fix synthesis test on big endian architectures

jamesjer:
I would like to revive issue https://github.com/pyglet/pyglet/issues/278.  I maintain the sympy package for Fedora, and it depends on pyglet, so I would like to fix the endianness issue so we can have sympy on s390x.  The WAV RIFF format is little endian:

- https://stackoverflow.com/questions/1111539/is-the-endianness-of-format-params-guaranteed-in-riff-wav-files
- https://en.wikipedia.org/wiki/Resource_Interchange_File_Format

That means that the test files are correct, and it is the unit tests that need adjusting.  I do not have physical access to an s390x machine, so I cannot test by listening, but I do have shell access to such a machine.  I ran flac over the test WAV files on the s390x.  The flac binary ran without error.  I copied the FLAC output files to an x86_64 machine, then played each original WAV file followed immediately by the corresponding FLAC file.  They sounded the same.  This lends further support to the theory that the WAV files are correct.

Python's WAV encoder swaps bytes if the sample width is greater than 1 and the machine is big endian.  The problem here is that the unit test does not use the WAV encoder, but examines the bytes that would have been fed to it; i.e., they are in host order.  This commit is one way of dealing with the issue.

### DIFF
--- a/tests/unit/media/test_synthesis.py
+++ b/tests/unit/media/test_synthesis.py
@@ -1,6 +1,7 @@
 from ctypes import sizeof
 from io import BytesIO
 import unittest
+import wave
 
 from pyglet.media.synthesis import *
 
@@ -65,9 +66,8 @@ class SynthesisSourceTest:
         source_name = self.source_class.__name__.lower()
         filename = "synthesis_{0}_{1}_{2}_1ch.wav".format(source_name, sample_size, sample_rate)
 
-        with open(get_test_data_file('media', filename), 'rb') as f:
-            # discard the wave header:
-            loaded_bytes = f.read()[44:]
+        with wave.open(get_test_data_file('media', filename)) as f:
+            loaded_bytes = f.readframes(-1)
             source.seek(0)
             generated_data = source.get_audio_data(source._max_offset)
             bytes_buffer = BytesIO(generated_data.data).getvalue()


benmoran56:
Hi James, 

Thanks for looking into this further, and for the PR. I think this is a reasonable idea, especially since you were able to manually check the output. 
Just so future contributors know what's going on in the test, we probably want to add a comment like:
`# for compatibility with big-endian systems, swap the byte order`

I also had a few thoughts on this. If the standard library `wave` module doesn't complain about loading the different endian files, what do you think about using it to simplify things a bit? This could be used in place of the raw file IO. 
```py
with wave.open(filename) as file:
    sampwidth = file.getsampwidth()
    ...
```
We could also then do `loaded_bytes = file.readframes(-1)`, to load in the raw data with the header pre-stripped.

jamesjer:
Good thought.  And since the wave reader swaps bytes as needed, there is no need for the test code to do so.  I have updated the pull request accordingly.

musicinmybrain:
Thanks for working on this—it will also be a big help for other Fedora packages that depend on pyglet, like `python-trimesh`.

benmoran56:
Looks good! I think we can reasonably assume that the `wave` module is doing the right thing, and it's a lot easier to understand at a glance.

benmoran56:
This was merged into `master`, but I'll cherry-pick it for the 1.5-maintenance branch as well. That way it can get out into the next point release for the current stable releases.

PR #472, opened 2021-09-23
Fixing shift and exclusive mouse mode

caffeinepills:
Fixes an issue with shift not being called correctly on multiple windows. It now follows where your keyboard is focused. Also resets shift state when the focus is lost (will be gained again if still being held on return to window)

Fix multiple issues with exclusive mode:
1) If the window were to lose focus and you clicked in the title bar, the ClipCursor no longer was bound to the window, and the cursor was unhidden.
2) Unable to drag and move the window.
3) Unable to maximize/minimize the Window
4) Unable to close the window with the X button.
5) When mouse button is pushed down (not a click) on frame components, it would disappear.
6) Mouse could sometimes go out of the bounds of the view area into the close and minimize button resulting in issues with the window.

Removed the raw input requiring the view_hwnd, it just goes off of where your keyboard has focus.

Added a raise exception to show information on possible NULL issue with the `_view_hwnd`. 

Was tested in Windows 10. Any feedback and additional testing appreciated.

Note:
There is still an issue with random hierarchy issues where the mouse can become visible again in exclusive mode, if the window rapidly loses/gains focus, but it seems pretty rare. It seems some libraries (SDL2) just have a schedule that checks every so often if the ClipCursor area was superseded by another window and then re-applies it. Not sure we need to go that intensive on it.

### DIFF
--- a/pyglet/window/win32/__init__.py
+++ b/pyglet/window/win32/__init__.py
@@ -106,6 +106,8 @@ class Win32Window(BaseWindow):
     _exclusive_mouse_lpos = None
     _exclusive_mouse_buttons = 0
     _mouse_platform_visible = True
+    _pending_click = False
+    _in_title_bar = False
     
     _keyboard_state = {0x02A: False, 0x036: False}  # For shift keys.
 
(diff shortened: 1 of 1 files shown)


benmoran56:
We probably don't need to get go as far as SDL just yet. 

How do you feel about merging this in for now?

caffeinepills:
It's fine merging in as is. This does fix the shift issue, and numerous exclusive mode issues to make it more usable.

It's not a perfect solution but I think it's good enough for most uses.

PR #478, opened 2021-10-01
Fix-shape-color-type

cbrnr:
The `color` property should be a tuple, but the setter currently assigns a list. This makes it hard to make comparisons, because e.g. `shape.color == (255, 255, 255)` won't work after the color has been set to `[255, 255, 255]`.

I think this should also be fixed in `master`.

### DIFF
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -267,7 +267,7 @@ class _ShapeBase:
 
     @color.setter
     def color(self, values):
-        self._rgb = list(map(int, values))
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):
Thank you!

cbrnr:
Should I make another PR for `master` or can/will this be "forward-ported"?

benmoran56:
Syncing the changes forward has to be done semi-manually, so I only do it once in a while. If you have the time to make another PR, that would assure that it gets pulled in right away. Either way is OK!

PR #480, opened 2021-10-02
Fix shape color type (list -> tuple)

cbrnr:
See #478.


benmoran56 (approved):
Thanks again for this one!

PR #482, opened 2021-10-04
new functionality to 2D Vectors

MrWardKKHS:
In Vec2 class
Added a __radd__ dunder method to work with sum([Vec2, Vec2, Vec2...])
new heading method returns angle in radians, 
new set_magnitude method to avoid having to chain methods to do this common task
new magnitude method as an alias of abs(Vec2)
new limit method which has behavior distinct from clamp

In Vec3 and 4
added magnitude, set_magnitude methods for consistency.

### DIFF
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -86,6 +86,31 @@ class Vec2(tuple):
     def __round__(self, ndigits=None):
         return Vec2(*(round(v, ndigits) for v in self))
 
+    def __radd__(self, other):
+        if other == 0:
+            return self
+        else:
+            return self.__add__(other)
+
+    def heading(self):
+        return _math.atan2(self[1], self[0])
+
+    def set_heading(self, angle):
+        mag = self.__abs__()
+        return Vec2(mag * _math.cos(angle), mag * _math.sin(angle))
+
+    def set_magnitude(self, value):
+        return self.normalize().scale(value)
+
+    def magnitude(self):
+        return self.__abs__()
+
+    def limit(self, value):
+        mag = self.__abs__()
+        if mag > value:
+            return self.set_magnitude(value)
+        return self
+            
     def lerp(self, other, alpha):
         return Vec2(self[0] + (alpha * (other[0] - self[0])),
                     self[1] + (alpha * (other[1] - self[1])))
@@ -158,6 +183,18 @@ class Vec3(tuple):
 
     def __round__(self, ndigits=None):
         return Vec3(*(round(v, ndigits) for v in self))
+
+    def __radd__(self, other):
+        if other == 0:
+            return self
+        else:
+            return self.__add__(other)
+
+    def set_magnitude(self, value):
+        return self.normalize().scale(value)
+
+    def magnitude(self):
+        return self.__abs__()
 
     def cross(self, other):
         return Vec3((self[1] * other[2]) - (self[2] * other[1]),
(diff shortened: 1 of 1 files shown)


einarf:
Glad you decided to make a PR 🥳 

Can probably add more docstrings here. I know the module don't have much of that currently, but we might as well include that.

pyglet/math.py
```diff
@@ -86,6 +86,31 @@ def __neg__(self):
...
         return Vec2(*(round(v, ndigits) for v in self))
 
+    def __radd__(self, other):
+        if other == 0:
+            return self
+        else:
+            return self.__add__(other)
+
+    def heading(self):
+        return _math.atan2(self[1], self[0])
```
einarf, line 96: This can probably be a `@property` and it needs some docstrings and return type hints.

pyglet/math.py
```diff
@@ -86,6 +86,31 @@ def __neg__(self):
...
+
+    def set_heading(self, angle):
+        mag = self.__abs__()
+        return Vec2(mag * _math.cos(angle), mag * _math.sin(angle))
```
einarf, line 100: The name here is a bit confusing. `set_*` often implies we are setting some state on the instance. Also a docstring and an indication what this method returns.

pyglet/math.py
```diff
@@ -86,6 +86,31 @@ def __neg__(self):
...
+
+    def set_magnitude(self, value):
+        return self.normalize().scale(value)
```
einarf, line 103: Again, The `set_*` name can be confusing here + docstring + typing

pyglet/math.py
```diff
@@ -86,6 +86,31 @@ def __neg__(self):
...
+
+    def magnitude(self):
+        return self.__abs__()
```
einarf, line 106: Can be a property + docstring + typing

einarf (commented):
Added some initial comments. These applies to other methods in this PR as well.

shenjackyuanjie:
make it use `@property`

pyglet/math.py
MrWardKKHS: Currently Vectors are immutable. Is there an appetite for making this a mutable type? It would allow for convenient in place statements like 

`Vec2.heading = pi / 2` 
`Vec2.heading += angle`

which I think is an elegant way of working with them and a similar thing could be achieved with the magnitude property. Currently to add friction, say, is 

`velocity_vector = velocity_vector.scale(friction_strength)`

which could be replaced with an __imul__ 

`vel *= friction_strength`

...after sorting multiplying between vectors and scalars

einarf: There's been a lot of profiling to make these types fast. All types extend `tuple` for that reason. We only have a few milliseconds per frame, so this is important 😃

MrWardKKHS:
- Added a lot of documentation - plus a .rst in the docs folder. 
- heading and mag are now properties
- A static method for creating a vector from an angle has been created, 
- set_heading set_mag renamed to_heading etc. Open to suggestions for better names. Removed these from 3D, 4D vectors. 
- added rotate method

If I'm on the right track here I can make similar changes to the 3D 4D classes, although my use for these is more limited (teaching coding and vector math to teenagers using video games)

benmoran56:
Your changes look good, and well done on the documentation. 
As for the naming, we are already using `from_xxxx` for the methods in the Matrix classes. Maybe it's good to stick to that for this as well.

PR #488, opened 2021-10-15
Add CameraGroup Example

LennyPhoenix:
Hey, I've been working on a project with Pyglet and ended up writing a new camera system using the graphics group functionality instead of the existing camera example.

I've cleaned up the code and put it into a new example file so others can use it themselves.
The example also includes a "centered" version and a small demo.

I believe this is easier to use and is a bit more minimal compared to the existing example, but I have left the current one as it is right now.

Let me know if there is anything I should change, thanks!


benmoran56:
Thanks Lenny, more examples are always good! 

I have a similar Group-based camera that I use for parallax effects in side-scrolling games. Maybe we should move the camera examples into the `window/` folder, and I can add mine as well?

LennyPhoenix:
Yeah that might be a good idea, I'll move them now :+1:

PR #494, opened 2021-10-24
Fix example events/window_platform_event.py

LennyPhoenix:
The `events/window_platform_event.py` example fails on Linux 1.5-maintenance due to the `xlib` module not being imported implicitly by `from pyglet.window.xlib import *`.

This PR fixes this by explicitly importing the submodule, and does this for the win32 section for uniformity.

Note that the OSX section is not affected by the PR as it is unimplemented anyway.


benmoran56 (approved):
Thanks Lenny!

PR #496, opened 2021-10-25
Implement Mouse 4 and 5 Support

LennyPhoenix:
This PR aims to implement mouse button 4 and mouse button 5 support.
This is my first time really diving into the window code, so let me know if there is anything I should change!

## Status

- [X] win32
  - [X] Button Press
  - [X] Mouse Drag
- [X] xlib
  - [X] Button Press
  - [ ] ~~Mouse Drag~~ Seems to be unsupported by Xorg
- [ ] ~~cocoa (I do not have a MacOS device so I cannot implement this myself)~~
  - [ ] Button Press
  - [ ] Mouse Drag


benmoran56:
Hi @LennyPhoenix,  I had a look at this recently, I think it just needs some additions to the `XlibWindow._event_motionnotify` and `XlibWindow._event_motionnotify_view` methods. Those are the only two places that dispatch the `on_mouse_drag` event, and they're only checking for `Button1MotionMask` to `Button3MotionMask`, so I'm guessing that adding the additional button masks there would help.

benmoran56:
Also, my information could be out of date, but I don't believe OSX has any native support for multi-button mouse. My understanding is that you need to use some kind of third party application to map these buttons to something useful. There may not be an easy way for us to receive these buttons in pyglet.

LennyPhoenix:
Ah ok, I'll just focus on Xlib for now then we can get this merged.

LennyPhoenix:
Ah, I'm seeing what the problem I was encountering before is now. It appears that `ev.xmotion.state` does not set any flags when button 4 or 5 are held down (it is always just 0), so I can't detect if the mouse buttons are held or not.

LennyPhoenix:
Using `xev` shows that Xorg doesn't actually set the state at all when dragging with mouse 4 or mouse 5:
```
MotionNotify event, serial 33, synthetic NO, window 0x6e00001,
    root 0x1de, subw 0x0, time 2181823, (469,172), root:(1473,753),
    state 0x0, is_hint 0, same_screen YES
```
Whereas with the left mouse button:
```
MotionNotify event, serial 33, synthetic NO, window 0x6e00001,
    root 0x1de, subw 0x0, time 2251476, (546,293), root:(1550,874),
    state 0x100, is_hint 0, same_screen YES
```
I have no idea why this is, it might just be that mouse 4 and 5 simply aren't fully supported by Xorg, in which case I don't think there is much I can do.

benmoran56:
That's strange. It could be that _dragging_ in particular is not supported. 
It might be worth while just adding a note to the codebase, and merging in what you have so far. That would also be easier for someone else to give it a try. 

I don't have any many-button mice, but I was thinking I might be able to find an emulator or something like that to test with.

LennyPhoenix:
Sure, if you could give it a try that would be handy but I'll add a note to the motion event for now.

benmoran56 (approved):

PR #501, opened 2021-10-31
docs: correct game example tutorial

HenrYxZ:
- In "Tweaking the game loop": Isn't to_add list declared above, and not bellow the loop previously referred to?
- In "Customizing bullet collisions": TheBullet class -> The Bullet class


benmoran56 (approved):

benmoran56:
Thank you Henry! 
Good catch.

PR #509, opened 2021-12-06
Remove dict subclass from KeyStateHandler and MouseStateHandler

Cleptomania:
This PR makes KeyStateHandler and MouseStateHandler be their own classes which do not subclass `dict`. 

Sub-classing dict causes problems with WeakRef when sub-classing Window, essentially making it impossible to these classes inside of a subclassed Window object.

This also caused problems with function names inherited from dict, like for example `dict.update()` which if you had an event handler for an update function, would try to use this function, which obviously isn't intended for it.

The solution to this is adding a `data` attribute to the classes, and having the functions work on that attribute rather than on `self` attribute directly. This is roughly the same kind of implementation found in the [UserDict](https://docs.python.org/3/library/collections.html#userdict-objects) class in Python. I've tested this somewhat extensively and I haven't been able to find anything within Pyglet that this breaks.

However this could potentially effect user code that is attempting to use this class directly as a dict, however my guess is that no one is doing that, or if so it's very few and far between. At worst this could just only be included 2.0.

This primarily serves to make it easier to use these classes in [Arcade](https://github.com/pythonarcade/arcade), and I've tested extensively from within Arcade as well and none of our existing code/examples/test suite broke with it.

### DIFF
--- a/pyglet/window/key.py
+++ b/pyglet/window/key.py
@@ -64,7 +64,7 @@ Usage::
 from pyglet import compat_platform
 
 
-class KeyStateHandler(dict):
+class KeyStateHandler:
     """Simple handler that tracks the state of keys on the keyboard. If a
     key is pressed then this handler holds a True value for it.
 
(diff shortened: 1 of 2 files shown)


benmoran56 (approved):
This should be fine. The `dict` subclass was just done for convenience, but a simple class works just as well.

PR #514, opened 2021-12-09
probably flaky test

XinyuLiu5566:
When I run pytest --flake-finder, this test gave me an error but if I re-initialize the clock object again in each test, the test passed. I don't know if this is a possible flaky test or not so I opened this PR.


benmoran56:
Thanks for opening the pull request. I can confirm that multiple runs will break this test, as it wasn't accounted for. 

I actually went ahead and cleaned up the clock module a bit, and at the same time I fixed this unit test in another way (using pytest fixtures).

I'll close this PR now, but thank you for raising the issue!

PR #515, opened 2021-12-10
Fix for Xaudio2 bugs

caffeinepills:
Fix issue with voices not being returned properly, causing them to grow out of control and start lagging.

Add sample rate to the voice key to prevent those from being re-used by another frequency. This should fix distortion with multiple audio sources.


caffeinepills:
This can be applied to master branch as well.

benmoran56:
Excellent!

PR #516, opened 2021-12-11
Adds WIC encoder

caffeinepills:
By default other platforms use a default of RGB/A when loading images such as PNG. However, Windows uses BGR/A as the default image format when loading. This can cause slowdowns because the default encoders such as PIL or PNG force RGB/A when encoding images, so it has to go through regex restructuring of the image data to the desired format. 

Here are some example of the speed differences between encoding different sized images:
```
WIC:
480x480 time: 0.04906469999999996s
2048x2048 time: 0.49710109999999985s

PIL:
480x480 time: 1.1027328s
2048x2048 time: 14.9952449s
In many cases running the 2048x2048 conversion would result in a MemoryError exception.
```

This new encoder supports bmp, jpg, tif, wmp, and png (default). It does not require any dependency such as PIL, pypng, or anything else. It will also convert the data to the BGR/A format if it is not correct. Other encoding options could be added in the future to control quality options if enough people request it.

Some other thoughts:
Not sure if other encoders can support BGR/A (Such as PIL), but if they do, perhaps they should be enhanced to support the other pixel formats?

### DIFF
--- a/pyglet/image/codecs/wic.py
+++ b/pyglet/image/codecs/wic.py
@@ -48,14 +48,20 @@ WICBitmapDitherTypeErrorDiffusion = 0x8
 WICBITMAPDITHERTYPE_FORCE_DWORD = 0x7fffffff
 WICBITMAPTRANSFORMOPTIONS_FORCE_DWORD = 0x7fffffff
 
-
 WICDecodeOptions = UINT
 WICDecodeMetadataCacheOnDemand = 0
 WICDecodeMetadataCacheOnLoad = 0x1
 WICMETADATACACHEOPTION_FORCE_DWORD = 0x7fffffff
 
+WICBitmapEncoderCacheOption = UINT
+WICBitmapEncoderCacheInMemory = 0x0
+WICBitmapEncoderCacheTempFile = 0x1
+WICBitmapEncoderNoCache = 0x2
+WICBITMAPENCODERCACHEOPTION_FORCE_DWORD = 0x7fffffff
+
 # Different pixel formats.
 REFWICPixelFormatGUID = com.GUID
+GUID_WICPixelFormatDontCare = com.GUID(0x6fddc324,0x4e03,0x4bfe,0xb1,0x85,0x3d,0x77,0x76,0x8d,0xc9,0x00)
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

larsoner:
FYI it looks like this PR [broke this workflow](https://dev.azure.com/LABSN/expyfun/_build/results?buildId=325&view=logs&jobId=db093384-e8eb-5bcc-f0a9-2a8f5e4220af&j=db093384-e8eb-5bcc-f0a9-2a8f5e4220af&t=b83569e5-86aa-5f40-c5f2-cb2c66a2c238):
```
expyfun\_experiment_controller.py:827: in screenshot
    pyglet.image.get_buffer_manager().get_color_buffer().save(file=fid)
C:\Users\VssAdministrator\AppData\Roaming\Python\Python39\site-packages\pyglet\image\__init__.py:500: in save
    encoder.encode(self, file, filename)
C:\Users\VssAdministrator\AppData\Roaming\Python\Python39\site-packages\pyglet\image\codecs\wic.py:594: in encode
    stream.InitializeFromFilename(filename, GENERIC_WRITE)
C:\Users\VssAdministrator\AppData\Roaming\Python\Python39\site-packages\pyglet\com.py:150: in _call
    ret = self.method.get_field()(self.i, self.name)(obj, *args)
_ctypes/callproc.c:997: in GetResult
    ???
E   OSError: [WinError -2147024809] The parameter is incorrect
```
We can work around it by passing a `filename`, but it doesn't seem like this should be necessary.

larsoner:
... actually, passing `filename` does not fix it:
```
expyfun\_experiment_controller.py:827: in screenshot
    pyglet.image.get_buffer_manager().get_color_buffer().save(
C:\Users\VssAdministrator\AppData\Roaming\Python\Python39\site-packages\pyglet\image\__init__.py:500: in save
    encoder.encode(self, file, filename)
C:\Users\VssAdministrator\AppData\Roaming\Python\Python39\site-packages\pyglet\image\codecs\wic.py:629: in encode
    image_data = image.get_data(fmt, actual_pitch)
E   AttributeError: 'ColorBufferImage' object has no attribute 'get_data'
```

caffeinepills:
https://github.com/pyglet/pyglet/pull/528

Can you see if this fixes the issues? Thanks

Issue #506, opened 2021-11-15
Generalize rotation of shapes


Ross-TheBoss:
I propose each shape having a method or read-only property specifing its vertices before rotation. Then a generalized function can be used to rotate those vertices before setting them to the shape's `vertex_list.vertices`.

For example in the Rectangle class:
```python
from math import cos, sin

def rotate(vertices, angle = math.tau, anchor_x, anchor_y):
  # Adjust all vertices by the anchor.
  vertices = [[x - anchor_x, y - anchor_y] for x, y in vertices]

   # Rotate the around the anchor.
   x, y = anchor_x, anchor_y
   r = -math.radians(self._rotation)
   cr = math.cos(r)
   sr = math.sin(r)

   for i, v in enumerate(vertices):
      v = [v[0] - x, v[1] - y]
      v = [v[0] * cr - v[1] * sr + x, v[0] * sr + v[1] * cr + y]
      vertices[i] = c
   
   return vertices

class Rectangle(_ShapeBase):
    def get_vertices(self) -> tuple[int, int]:
        x1 = self._x - self._anchor_x
        y1 = self._y - self._anchor_y
        x2 = x1 + self._width
        y2 = y1 + self._height

        return (x1, y1), (x2, y1), (x1, y2), (x2, y2)

    def _update_position(self):
        if not self._visible:
            self._vertex_list.position[:] = (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
        elif self._rotation:
            vertices = rotate(self.get_vertices(), angle=self._rotation, self._anchor_x, self._anchor_y)
            self._vertex_list.position[:] = vertices[0] + vertices[1] + vertices[3] + \
                                            vertices[0] + vertices[3] + vertices[2]
```
This code might not actually be functional but hopefully it gets across what I am proposing.

benmoran56:
Hmm, maybe we could just make the `self._vertex_list` public? That way users could just modify it directly if they wanted to. Would that make sense? 
We have to be a little careful with how much functionality goes in here, since the `graphics` module can already be used directly for advanced cases.

Ross-TheBoss:
Yes, I think `self.vertex_list` could be made public. Users could then use it to easily generate vertices for a shape for use in calculations or to create a composite shape (using the pyglet.graphics module to draw it).
E.g. 
```python
vertices = Circle(10, 10, 5).vertex_list.position
```
Or they could colour individual points differently.

[...]

PR #517, opened 2021-12-12
Pyglet 1.5 maintenance

Ross-TheBoss:
Addresses issue: #506

This change adds a general function for rotating points (based upon how the `Ellipse` rotates points), and changes the `Ellipse`, `Rectangle`, `BorderedRectangle` and `Polygon` class to use it. This reduces some code duplication.


benmoran56:
For pyglet 2.0, most of this code will be moved inside of the shader (eventually). 
For that reason, could we make that `rotate` function private (`_rotate`) so that it doesn't get relied on?

Ross-TheBoss:
Yes, I didn't know about the changes that would take place for pyglet 2.0. I think the rotate function should be made private then.

benmoran56 (approved):
Thanks Ross!

PR #518, opened 2021-12-13
Add a way to directly set HTML font size

shenjackyuanjie:
now can use `<font real_size=20>abc</font>` to set text size directly


shenjackyuanjie:
sorry, but my respoincy have more commit

pyglet/font/win32query.py
```diff
@@ -340,7 +340,11 @@ def _enum_font_names(logfont, textmetricex, fonttype, param):
     global FONTDB
 
     lf = logfont.contents
-    name = lf.lfFaceName.decode('utf-8')
+    try:
+        name = lf.lfFaceName.decode('utf-8')
+    except UnicodeDecodeError:
+        name = lf.lfFaceName.decode('ANSI')
```
shenjackyuanjie, line 346: bad windows

shenjackyuanjie:
now it's done!

pyglet/font/win32query.py
shenjackyuanjie: this is may a way on how to fix #524

shenjackyuanjie:
about the decode of `lfFaceName`
here is a `ttf`` file witch need decode in `ANSI` or `gb2312`
http://gofile.me/4yON4/haluBRI2f

pyglet/font/win32query.py
benmoran56: If possible, can you move these changes to a new pull request?

shenjackyuanjie: > If possible, can you move these changes to a new pull request?

hmmmm，i'll try

caffeinepills:
The `UnicodeDecodeError` portion of this was never separated. https://github.com/pyglet/pyglet/pull/584 Should resolve issue with unicode font names.

Also the file you linked is not available.

shenjackyuanjie:
> 

oh，got it

shenjackyuanjie:
aaaa
see #592  for more!

shenjackyuanjie:
close!

PR #522, opened 2021-12-22
Fix for on_resize event

caffeinepills:
Fixes issue with `on_resiz`e not being called due to handling of `WM_WINDOWPOSCHANGED` event.


einarf:
I can confirm this fixes #522 in both master and 1.5. Also in Aracde and moderngl-window.

benmoran56 (approved):

benmoran56:
Thanks Caffeine! And thanks for testing Einarf!

PR #525, opened 2021-12-25
Fix kernel32.SetLastError argtypes

Square789:
Scrolled through the discord in boredom and stumbled over [that month-old question](https://discord.com/channels/438622377094414346/701585557637038152/911679732578414663), I guess there's no harm fixing the problem in pyglet itself: https://docs.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-setlasterror


benmoran56 (approved):

benmoran56:
Thanks for that! Some of these bindings are "just enough" for what's required by pyglet, so some things are not as robust as they could be.

PR #533, opened 2022-01-10
BUG: Fix bug with dropping frames

larsoner:
When querying `get_next_video_timestamp()`, it should not (effectively) discard the given frame. The current code seems to do this unintentionally because it pops from the video queue, but does not put the valid frame back in the queue.

For example, this code with this video:

https://user-images.githubusercontent.com/2365790/148837466-8c8c9d51-b720-4077-9554-51c312b6fff2.mp4

```
import numpy as np
import pyglet
from pyglet import gl
media = pyglet.media.load('example-video.mp4')
player.queue(media)
player.play()


@window.event
def on_draw():
    window.clear()
        tex = player.get_texture()
        for _ in range(10):
            print(player.source.get_next_video_timestamp())
        tex.blit(0, 0)
    else:
        pyglet.app.exit()


pyglet.app.run()
```
Should produce a smooth video with no interruptions, and we should get blocks of the same timestamp printed 10 times in a row. On `pyglet-1.5-maintenance` we get very choppy video with the following increasing time outputs (wrong!):

https://user-images.githubusercontent.com/2365790/148837732-0158421b-198b-42e7-a772-4e49ac312138.mp4

```
0.133333
0.166667
0.233333
0.266667
0.3
0.333333
0.366667
0.4
0.433333
0.466667
```
On this branch, we get correctly-printed blocks of 10 frame numbers and smooth video:

https://user-images.githubusercontent.com/2365790/148838028-fb5fd2af-b505-49e7-ac99-2fb88a7d19f9.mp4

```
0.066667
0.066667
0.066667
0.066667
0.066667
0.066667
0.066667
0.066667
0.066667
0.066667
0.1
0.1
0.1
0.1
0.1
0.1
0.1
0.1
0.1
0.1
```

### DIFF
--- a/pyglet/media/player.py
+++ b/pyglet/media/player.py
@@ -478,6 +478,11 @@ class Player(pyglet.event.EventDispatcher):
 
             pyglet.clock.schedule_once(self._video_finished, 0)
             return
+        elif ts > time:
+            # update_texture called too early (probably manually!)
+            pyglet.clock.schedule_once(self.update_texture, ts - time)
+            return
+
 
         image = source.get_next_video_frame()
         if image is not None:
--- a/pyglet/media/codecs/ffmpeg.py
+++ b/pyglet/media/codecs/ffmpeg.py
@@ -962,6 +962,7 @@ class FFmpegSource(StreamingSource):
                     self._decode_video_packet(video_packet)
                 if video_packet.image is not None:
                     ts = video_packet.timestamp
+                    self.videoq.appendleft(video_packet)  # put it back
                     break
                 self._get_video_packet()
         else:


larsoner:
The second commit fixes a second bug: calling `update_texture` should not always result in an advance in the video by one frame. If you replace the `print` line in the code above with `player.update_texture()`, you get a video that finishes in about 1 second (because `update_texture` *always* moves ahead by at least one frame), which is wrong. With the shortcut in fc474c8, you get a correct video.

larsoner:
FWIW, these fixes restore functionality / code that worked with Pyglet 1.3 but currently fails/behaves badly with 1.5.

pyglet/media/player.py
```diff
@@ -478,6 +478,11 @@ def update_texture(self, dt=None):
 
             pyglet.clock.schedule_once(self._video_finished, 0)
             return
+        elif ts > time:
```
larsoner, line 481: If this seems too risky (in terms of maybe dropping an occasional frame) due to float arithmetic and timer precision, it could easily be adjusted to `ts > time + frame_duration / 2.` or something more tolerant.

benmoran56 (approved):
Thank you for the pull request.

benmoran56:
A lot of this code was rewritten new since 1.3 (when AVbin went unmaintained), so unfortunately it's not as solid as it could be. I think the FFmpeg decoder works pretty well considering, but there are also a few remaining edge cases. The biggest one I can think of is that it doesn't properly parse sources that contain lots of stream types. Things like audio files with embedded images, for example. 

Thanks again for putting the PR together!

PR #534, opened 2022-01-10
BUG: Fix video alpha on Windows

### LINKED ISSUE #531: BUG: Video texture alpha differs by platform
This simple video code on Linux using [this file](https://github.com/LABSN/expyfun-data/blob/master/video/example-video.mp4?raw=true):

<details>

```
import numpy as np
import pyglet
from pyglet import gl
media = pyglet.media.load('example-video.mp4')
player.queue(media)
player.play()


@window.event
def on_draw():
    window.clear()
        tex = player.get_texture()
        global rawimage
        rawimage = tex.get_image_data()
        fmt = 'RGBA'
        assert rawimage.format == fmt
        pitch = rawimage.width * len(fmt)
        pixels = rawimage.get_data(fmt, pitch)
        data = np.frombuffer(pixels, np.uint8).reshape(rawimage.height, rawimage.width, len(fmt))
        print(data[..., -1])
        print(data.shape)
        gl.glEnable(gl.GL_BLEND)
        gl.glBlendFunc(gl.GL_SRC_ALPHA, gl.GL_ONE_MINUS_SRC_ALPHA)
        tex.blit(0, 0)
    else:
        pyglet.app.exit()


pyglet.app.run()
```

</details>

Produces a correct video output:

Screenshot from 2022-01-07 11-17-17

And correct alpha channel in the texture if you pull the data (all 255's):
```
...
[[255 255 255 ... 255 255 255]
 [255 255 255 ... 255 255 255]
 [255 255 255 ... 255 255 255]
 ...
 [255 255 255 ... 255 255 255]
 [255 255 255 ... 255 255 255]
 [255 255 255 ... 255 255 255]]
...
```

But on Windows, it produces an all black image, and the alpha channel is all zeros:
```
[[0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 ...
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]]
```
This was discovered because blitting does not work for this video/texture on Windows when blending is enabled (because alpha is zero).

This suggests that there is some difference in the FFMPEG->texture extraction on Windows compared to Linux.

larsoner:
Closes #531.

One possible fix -- actually request the alpha channel in the data. It seems to fix the issue. Another option would be just to set explicitly the alpha channel to 1 (or 255?) in `ImageData` after creation, but this solution should be faster since it relies on Microsoft's C code to do it for us.

### DIFF
--- a/pyglet/media/codecs/wmf.py
+++ b/pyglet/media/codecs/wmf.py
@@ -616,7 +616,7 @@ class WMFSource(Source):
 
         imfmedia.Release()
 
-        uncompressed_mt.SetGUID(MF_MT_SUBTYPE, MFVideoFormat_RGB32)
+        uncompressed_mt.SetGUID(MF_MT_SUBTYPE, MFVideoFormat_ARGB32)
         uncompressed_mt.SetUINT32(MF_MT_INTERLACE_MODE, MFVideoInterlace_Progressive)
         uncompressed_mt.SetUINT32(MF_MT_ALL_SAMPLES_INDEPENDENT, 1)
 


benmoran56 (approved):

benmoran56:
This should be OK.

The WMF video decoder is not accellerated, so one possible downfall would be if this causes a performance impact.
Generally speaking the WMF decoder is suitable for smaller videos. For larger resolutions, FFmpeg is preferred. 

I don't have a physical Windows machine at the moment, but if possible could you test this change with a larger video? (1080p should do it).

larsoner:
I tested a 1080p clip on a Windows VM and it played smoothly on `pyglet-1.5-maintenance` and on this branch

benmoran56:
Thank you for taking the time to test it.
Merged!

PR #535, opened 2022-01-14
ENH: Prefer ffmpeg

larsoner:
In https://github.com/pyglet/pyglet/pull/534#issuecomment-1010732272 it was pointed out that WMF is not accelerated, and ffmpeg is preferred for larger videos. However, there is no way for the user to actually make this choice AFAIK, because it's set in the order of `add_decoders`. Assuming it's not *awful* for smaller videos (where performance will be less important anyway), it seems like `ffmpeg` should be the preferred decoder on Windows. So this PR is really a proposal to change the priority order, feel free to close if it doesn't make sense.

An alternative would be to allow users to set an option to set the priority / preference using `pyglet.options` somehow, like there is for audio.

The motivation here is that the playback speed difference between these backends [for large videos is drastic](https://github.com/LABSN/expyfun/pull/439#issuecomment-1013295729).

### DIFF
--- a/pyglet/media/codecs/__init__.py
+++ b/pyglet/media/codecs/__init__.py
@@ -53,7 +53,7 @@ class MediaDecoder(Decoder):
 
     def decode(self, file, filename, streaming):
         """Read the given file object and return an instance of `Source`
-        or `StreamingSource`. 
+        or `StreamingSource`.
         Throws MediaDecodeException if there is an error.  `filename`
         can be a file type hint.
         """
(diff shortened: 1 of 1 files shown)


caffeinepills:
You can pass a specific decoder to the media load function.

```py
from pyglet.media.codecs.ffmpeg import FFmpegDecoder
specific_decoder = FFmpegDecoder()
pyglet.media.load(filename, decoder=specific_decoder)
```

larsoner:
Okay, makes sense!

PR #539, opened 2022-01-26
Changed Vec2

janscas:
1) Allow to operate with scalar values
2) Dynamic class return so it's an object easier to subclass
3) Added \_\_floordiv\_\_ method
4) Added as_type method to return the tuple with a different type (int to float or whatever)

Can this somehow be merged?

Thanks

### DIFF
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -52,69 +52,263 @@ def clamp(num, min_val, max_val):
 
 
 class Vec2(tuple):
+    """A two dimensional vector represented as an X Y coordinate pair.
+
+    :parameters:
+        `x` : int or float :
+            The X coordinate of the vector.
+        `y`   : int or float :
+            The Y coordinate of the vector.
+
+    Vectors must be created with either 0 or 2 values. If no arguments are provided a vector with the coordinates 0, 0 is created.
+
+    Vectors are stored as a tuple and therefore immutable and cannot be modified directly
+    """
 
     def __new__(cls, *args):
         assert len(args) in (0, 2), "0 or 2 values are required for Vec2 types."
-        return super().__new__(Vec2, args or (0, 0))
+        return super().__new__(cls, args or (0, 0))
+
+    @staticmethod
+    def from_polar(mag, angle):
+        """Create a new vector from the given polar coodinates.
+
+        :parameters:
+            `mag`   : int or float :
+                The magnitude of the vector.
+            `angle` : int or float :
+                The angle of the vector in radians.
+
+        :returns: A new vector with the given angle and magnitude.
+        :rtype: Vec2
+        """
+        return self.__class__(mag * _math.cos(angle), mag * _math.sin(angle))
 
     @property
     def x(self):
+        """The X coordinate of the vector.
+
+        :type: float
+        """
         return self[0]
 
     @property
     def y(self):
+        """The Y coordinate of the vector.
+
+        :type: float
+        """
         return self[1]
 
+    @property
+    def heading(self):
+        """The angle of the vector in radians.
+
+        :type: float
+        """
+        return _math.atan2(self[1], self[0])
+
+    @property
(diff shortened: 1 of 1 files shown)


pyglet/math.py
```diff
@@ -52,69 +52,263 @@ def clamp(num, min_val, max_val):
...
+        return self.__class__(self[0] + (alpha * (other[0] - self[0])),
+                              self[1] + (alpha * (other[1] - self[1])))
 
     def scale(self, value):
-        return Vec2(self[0] * value, self[1] * value)
+        """Multiply the vector by a scalar value.
+
+        :parameters:
+            `value`  : int or float :
+                The ammount to be scaled by
```
LennyPhoenix, line 229: This looks like a typo here "ammount"

pyglet/math.py
```diff
@@ -52,69 +52,263 @@ def clamp(num, min_val, max_val):
...
+        :rtype: Vec2
+        """
+        return self.__class__(self[0] * value, self[1] * value)
+
+    def rotate(self, angle):
+        """Create a new Vector rotated by the angle. The magnitude remains unchanged.
+
+        :parameters:
+            `angle` : int or float :
+                The angle to rotate by
```
LennyPhoenix, line 241: Is this radians or degrees?

LennyPhoenix (changes requested):
Code looks good to me, just a few little documentation changes imo

pyglet/math.py
janscas: This is not from my code, I just copied from the original Vec2, but I can fix that anyway

pyglet/math.py
janscas: no clue, it's from the original Vec2 code...

LennyPhoenix: I think it will be radians, might be worth adding that to the docstring

janscas: i’ll add that

janscas: @LennyPhoenix Done!

LennyPhoenix: I think "radiants" is a typo, should be all clear once that's fixed!

janscas: Done

LennyPhoenix (approved):
Looks great! Nice work :+1:

benmoran56:
Thanks for the pull request Janscas, and thanks for reviewing it Lenny! 
One small comment: 
It's probably a bit faster to use `Vec2(...` in place of `self.__class__(...` because we can avoid the `self` dot lookup. 
It's a micro-optimization for sure, but it's good if we can keep these classes as fast as possible.

janscas:
> Thanks for the pull request Janscas, and thanks for reviewing it Lenny!
> One small comment:
> It's probably a bit faster to use `Vec2(...` in place of `self.__class__(...` because we can avoid the `self` dot lookup.
> It's a micro-optimization for sure, but it's good if we can keep these classes as fast as possible.

Then keep in mind that the “dot” lookup is extremely fast compared to the call to `isinstance`.
I you want super fast Vec2 then you probably don’t want this feature although I find calling `Vec2(4,5) * 2` ultra useful.

benmoran56:
Sorry for leaving this hanging. 
That's a good point about the `isinstance` cost. Despite this being a useful change, it might not be worth it in that regard.

janscas:
Yep, better close this. Thanks anyway

einarf:
There are still some new methods in this PR that could be useful?

benmoran56:
I believe all of these methods are already existing in the 2.0 branch, so I think we're safe to close it.

janscas:
> I believe all of these methods are already existing in the 2.0 branch, so I think we're safe to close it.

Well, the `__floordiv__` is actually missing from Vec2

benmoran56:
Thanks @janscas, I added that in master.

PR #548, opened 2022-02-24
Fix #542

### LINKED ISSUE #542: [2.0] Window.set_size no longer dispatch on_resize (windows 10)
There seems to be a change in behavior between dev11 and dev12.

If you call `set_size(...)` in the window initializer (subclass) the `on_resize` method is never dispatched. You can however call `set_size()` later and it will work, but not in the window initializer.

The user is doing something like this:

```py
class App(Window):
        super().__init__(*args, **kwargs, resizable=True)
        # Apply configuration from config.yml
        self.set_icon(c.ICON)
        self.set_caption(c.TITLE)
        self.set_size(c.WIDTH, c.HEIGHT)
        self.set_update_rate(1 / c.FPS)
```

This did work in dev11  and earlier.

The image below shows an image with misconfigured viewport. The white area is even part of the bug here. It's persistent might indicate the underlying framebuffer itself was not really resized at all?

Ball-Man:
The actual usable window surface wouldn't get correctly resized.

In order to force an update from the window, after changing the size
with SetWindowPos a rapid toggle of SW_HIDE/SW_SHOW is executed.

### DIFF
--- a/pyglet/window/win32/__init__.py
+++ b/pyglet/window/win32/__init__.py
@@ -410,6 +410,8 @@ class Win32Window(BaseWindow):
         width, height = self._client_to_window_size(width, height)
         _user32.SetWindowPos(self._hwnd, 0, 0, 0, width, height,
                              (SWP_NOZORDER | SWP_NOMOVE | SWP_NOOWNERZORDER))
+        _user32.ShowWindow(self._hwnd, SW_HIDE)
+        _user32.ShowWindow(self._hwnd, SW_SHOW)
         self.dispatch_event('on_resize', width, height)
 
     def get_size(self):


Ball-Man:
I'm sorry if this is on master and not on maintenance, but the issue was here

caffeinepills:
This causes the window to visibly disappear, flicker, and reappear.

Ball-Man:
I see, this is inconvenient, I couldn't notice it at first. I'll look for a different approach then.

caffeinepills:
By the way I was able to replicate the issue you are seeing about it sometimes not updating the window to refresh it's contents. I will look into a resolution. Looks like this was a regression introduced in a commit.

PR #559, opened 2022-03-16
DirectWrite enhancement and fixes

caffeinepills:
* Fix pointer leak on analyzing scripts.
* 0 width/height glyphs are now ignored to prevent errors.
* Add `win32_disable_shaping` as an option to speedup text by removing the shaping process for DW. If your font is simple, monospaced, or you require no advanced OpenType features, this option may be useful.
* Added `colored` attribute for glyphs as a possible future enhancement to exclude them from being rendered with the font color. Fonts are now queried to see if they support color or not.

You can disable color rendering entirely via
```py
from pyglet.font.directwrite import DirectWriteGlyphRenderer
DirectWriteGlyphRenderer.draw_options = 0
```

### DIFF
--- a/pyglet/__init__.py
+++ b/pyglet/__init__.py
@@ -165,6 +165,7 @@ options = {
     'advanced_font_features': False,
     'headless': False,
     'headless_device': 0,
+    'win32_disable_shaping': False,
 }
 
 _option_types = {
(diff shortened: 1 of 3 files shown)


benmoran56 (approved):

benmoran56:
Looks good!

PR #564, opened 2022-03-24
Create buffers with 0 byte data by default

einarf:
There is no guarantee a buffer will contain 0 byte values on creation if no data is supplied. it really depends on the drivers.

Changing this fixes the text issue because the translation buffer data is not always set and translation values will contain random data depending on drivers. AMD and Intel drivers are more likely to have random data on creation. It might be that we want to fix the original issue as well. When the text vertex data is split into a different domain the translation values are probably not copied to the new buffer, so that probably create other issues!

Still, creating 0 byte value buffers is probably a good idea, or it can be optional on creation (init parameter)

### DIFF
--- a/pyglet/graphics/vertexbuffer.py
+++ b/pyglet/graphics/vertexbuffer.py
@@ -188,7 +188,8 @@ class BufferObject(AbstractBuffer):
         self.id = buffer_id.value
 
         glBindBuffer(target, self.id)
-        glBufferData(target, self.size, None, self.usage)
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

benmoran56:
Great fix!
I think in the past we didn't trigger this bug, because we never created an attribute buffer without also providing it with some data. Since 2.0, after moving things like translation into the shader, we are doing that more often.

PR #565, opened 2022-03-25
Glyph optimization

caffeinepills:
Right now the glyph atlas is not optimized well. It seems to be using a pre-TextureBin setup, but not as elegant.  This leads to problems:

The default atlas size is 512x512. If your glyph is below any of those dimensions it will keep generating textures. So if your glyph is large 180x510 (like the timer example), you could end up having only 3 or so characters per texture, causing lots of textures to be generated.  

An alternative: Create a texture atlas big enough to fit as many glyphs as possible. Standard US keyboard contains about 96 different characters, so we should try to fit atleast that many glyphs into an atlas to reduce calls. 

This PR offers the solutions:
1) Add the TextureBin class to manage the Atlases.
2) Added the `optimize_fit` option to get a rough size of how big an atlas needs to be to hold 100 glyphs (configurable by `glyph_fit`. This will start at 512x512 and move up sizes until it gets a size that fits the first glyph created. This will cap based on your video cards texture size (`pyglet.image.get_max_texture_size`). You can set `optimize_fit` to `False` if you wish to force a specific atlas size yourself.

### DIFF
--- a/pyglet/font/base.py
+++ b/pyglet/font/base.py
@@ -47,7 +47,7 @@ from pyglet import image
 
 _other_grapheme_extend = list(map(chr, [0x09be, 0x09d7, 0x0be3, 0x0b57, 0x0bbe, 0x0bd7, 0x0cc2,
                                         0x0cd5, 0x0cd6, 0x0d3e, 0x0d57, 0x0dcf, 0x0ddf, 0x200c,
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):
Very nice reuse the other Atlas code, 
and good idea for the optimal sizing!

PR #571, opened 2022-03-27
DirectWrite fix for 0 space characters.

caffeinepills:
It seems the font renderer still needs a glyph for 0 space characters. Since layouts use the length of the text, and *not* the count of the glyphs returned by the renderer, characters like \n which have no glyph are still counted as length of the original string.

A more permanent fix is needed eventually to use glyph count instead of length of text when determining runs and other information. As some fonts may combine multiple letters into one glyph for ligatures or special fonts.

This PR creates an 'empty' glyph from a space character, when the font renderer is created to use as a dummy character when a glyph is not found.

### DIFF
--- a/pyglet/font/directwrite.py
+++ b/pyglet/font/directwrite.py
@@ -1877,6 +1877,7 @@ class Win32DirectWriteFont(base.Font):
     _default_name = 'Segoe UI'  # Default font for Win7/10.
 
     _glyph_renderer = None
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

benmoran56:
It seems like this is already the default behavior in `base.Font.get_glyphs`, when dealing with newline characters.

Issue #570, opened 2022-03-27
[2.0] Creating labels with RGB color makes glyphs disappear


einarf:
Specyfing RGB instead  of RGBA color using  the standard  timer example:

![bilde](https://user-images.githubusercontent.com/5418180/160273743-da5d501b-330c-4af5-ad69-83ae3eabbcd9.png)
![bilde](https://user-images.githubusercontent.com/5418180/160273745-a158c4f0-5a57-4a06-9832-99ab45446e9f.png)

Seems like the supplied color is just blindly multiplied and assigned to the color buffer.

Pyglet 1.5 will trigger this error unless you use an RGBA  color
```py
2022-03-27 00:47:19-0500 [-]   File "C:\client\pyglet\graphics\__init__.py", line 371, in add
2022-03-27 00:47:19-0500 [-]     vlist._set_attribute_data(i, array)
2022-03-27 00:47:19-0500 [-]   File "C:\client\pyglet\graphics\vertexdomain.py", line 447, in _set_attribute_data
2022-03-27 00:47:19-0500 [-]     region.array[:] = data
2022-03-27 00:47:19-0500 [-] ValueError: Can only assign sequence of same size
```

Probably affects all layouts.

benmoran56:
Should be fixed now!

Issue #568, opened 2022-03-26
Rendering a layout with attributed text randomly miscolors text


DigiDuncan:
In pyglet 2.0a2 (the version installed with Arcade 2.6.13), rendering this string to a layout, like this:

```py
import pyglet

class Window(pyglet.window.Window):

    def __init__(self):
        super().__init__()
[...]

PR #572, opened 2022-03-27
Background color fixes and color length checks

caffeinepills:
Fix for background color text.

1) Using `i` for creating the indice list was overshadowing the `i` passed into the function.

2) Wasn't using the correct amount of indices for background vertices.

3) Added checks for length of tuples for colors to ensure RGBA. This prevents blank data from occurring in the alpha channel causing issues. (As no error would be produced to give you an idea of why your text was not displaying.)

Should fix issues with: https://github.com/pyglet/pyglet/issues/568 and https://github.com/pyglet/pyglet/issues/570


benmoran56 (approved):
Thanks for the fix!

PR #578, opened 2022-04-03
Fix memory leak in DirectInputDevice

Ball-Man:
Creation of new DirectInputDevices through get_devices() (and other
functions that rely on it) would result in a memory leak since new
devices are allocated each time from the dinput API, but never released.

By releasing the device at destruction time, the leak is easily solved.


benmoran56 (approved):

benmoran56:
Thanks for this!

PR #583, opened 2022-04-09
Fix resize issue and move event

caffeinepills:
This should fix two issues: One window was not updating image on resize: 
Example:
image

`on_move` event was not being called.

(Notes: WM_WINDOWPOSCHANGED is newer and supposed to be faster, however lParam returns garbage data at the beginning of Window creation making it unreliable. Sizes are also based on total window size, not viewable areas. So not as suitable.)

### DIFF
--- a/pyglet/window/win32/__init__.py
+++ b/pyglet/window/win32/__init__.py
@@ -832,11 +832,6 @@ class Win32Window(BaseWindow):
         else:
             return None
 
-    @Win32EventHandler(WM_WINDOWPOSCHANGED)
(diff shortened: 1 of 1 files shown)


matanster:
Would this be tied to resizing a loaded image, or just a window resizing bug fix? The loaded image object is a little opaque and we need to resize it to fit the window on occasion.

matanster:
Okay, that resizing can be simply done at the blit of the image. (https://stackoverflow.com/a/20809459/1509695). Please ignore the question.

Issue #540, opened 2022-01-27
Thai Language can not be used in windows when I use pyglet


zhangxiansheng:
When I use Thai Language in MacOS, it works.
But in Windows, Thai Language can not be supported.

e.g.

```
st = '<type some Thai Language here>'
pyglet.text.Label( st ) # error here
```

error message:
```
Traceback (most recent call last):
  File "<pyshell#14>", line 1, in <module>
    pyglet.text.Label(es.native_str(a))
  File "C:\Users\zyw\AppData\Local\Programs\Python\Python310\lib\site-packages\pyglet\text\__init__.py", line 469, in __init__
    super(Label, self).__init__(document, x, y, width, height,
  File "C:\Users\zyw\AppData\Local\Programs\Python\Python310\lib\site-packages\pyglet\text\__init__.py", line 268, in __init__
    super(DocumentLabel, self).__init__(document,
  File "C:\Users\zyw\AppData\Local\Programs\Python\Python310\lib\site-packages\pyglet\text\layout.py", line 858, in __init__
    self.document = document
  File "C:\Users\zyw\AppData\Local\Programs\Python\Python310\lib\site-packages\pyglet\text\layout.py", line 887, in document
    self._init_document()
  File "C:\Users\zyw\AppData\Local\Programs\Python\Python310\lib\site-packages\pyglet\text\layout.py", line 1336, in _init_document
    self._update()
  File "C:\Users\zyw\AppData\Local\Programs\Python\Python310\lib\site-packages\pyglet\text\layout.py", line 1247, in _update
    lines = self._get_lines()
  File "C:\Users\zyw\AppData\Local\Programs\Python\Python310\lib\site-packages\pyglet\text\layout.py", line 1224, in _get_lines
    glyphs = self._get_glyphs()
  File "C:\Users\zyw\AppData\Local\Programs\Python\Python310\lib\site-packages\pyglet\text\layout.py", line 1384, in _get_glyphs
    glyphs.extend(font.get_glyphs(text[start:end]))
  File "C:\Users\zyw\AppData\Local\Programs\Python\Python310\lib\site-packages\pyglet\font\base.py", line 398, in get_glyphs
    self.glyphs[c] = glyph_renderer.render(c)
  File "C:\Users\zyw\AppData\Local\Programs\Python\Python310\lib\site-packages\pyglet\font\win32.py", line 427, in render
    ord(text), ord(text), byref(abc)):
TypeError: ord() expected a character, but string of length 2 found
```

[...]

PR #585, opened 2022-04-12
GDI crash fix for larger sized unicode characters

caffeinepills:
Fix for issue mentioned here: https://github.com/pyglet/pyglet/issues/540

Stale thread but, I was able to reproduce the error using this test string: টেস্টিং 

We assume the text passed will always be 1 length character. Change to iterate over the text passed and get each codepoint and use that to determine lsb and width.


caffeinepills:
Cleaned it up a bit more. Now only uses width of the bounding box if the GDI call fails as a backup. (Haven't found an example that does)

PR #586, opened 2022-04-12
Simplify layout shaders

einarf:
* Creating a mat4 to translate is overkill. Just add the translation to the position
* The translation matrix was also moving the  text one unit closer to the camera. Not sure if this was intended.
* Fixed a  few literals that could  potentially cause some compilers to complain


benmoran56 (approved):

benmoran56:
Thanks. That single unit translation towards the camera looks like a mistake.

PR #595, opened 2022-04-17
Some Caret and Layout adjustments.

caffeinepills:
Fix selection
Fix background.

Still some issue with the first Layout created. In the text_input example, if you create the rectangle before the textbox, then it will show up. If you create after, then the backgrounds do not show up.

### DIFF
--- a/pyglet/font/directwrite.py
+++ b/pyglet/font/directwrite.py
@@ -35,6 +35,7 @@ except OSError as err:
 
 _debug_font = debug_print('debug_font')
 
+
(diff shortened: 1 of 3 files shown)


eschan145 (approved):

benmoran56:
Are there any dangers with merging this in?

benmoran56:
OK thanks, I'll merge it in and then give it a look locally.
I'll have a look at the scissor_area as well. 
We should probably rename that to `clip_area` since we're not using glScissor anymore...

PR #596, opened 2022-04-18
Fix matrix math in model example

einarf:
Fix matrix math in the 3d model example. The models had the wrong origin and we are better off calculating the rotation from time instead of delta time to.


benmoran56 (approved):

benmoran56:
Thanks!

PR #599, opened 2022-04-22
Sprite enhancements

caffeinepills:
Add Z axis to Sprite
Created AdvancedSprite that let's people set a program if they wish to modify the default shader.
Added sprite examples.


benmoran56 (approved):
This was definitely needed, thanks!

PR #600, opened 2022-04-24
Fix attempting to load multiple libraries on win32

einarf:
When trying to load multiple library names on win32 `load_library` will just exit the loop on the first library that fails. Let's instead print some debug info instead.


benmoran56 (approved):

benmoran56:
That's more useful.

PR #580, opened 2022-04-05
FFmpeg 5 support

caffeinepills:
FFmpeg 5 added and deprecated a lot of things. This branch attempts to support both 4.x and 5.

However, will need testing on both on various platforms.

PR #604, opened 2022-05-03
Pin ffmpeg4 libs on linux too

xantares:
This avoids picking up ffmpeg 5.x libs in the stable version (#580 is only in dev version)


benmoran56 (approved):

benmoran56:
Thanks!

PR #615, opened 2022-05-24
Fix typing issue in evdev on Py < 3.10

### LINKED ISSUE #614: Bug: Subscripting list instead of typing.List breaks on Py < 3.10
The README states support for 3.6+. However, the signature on this function uses a feature not supported until later Python versions.
https://github.com/pyglet/pyglet/blob/c48c45ff8bc391b4fdab892e5b0026c0a98a6f6a/pyglet/input/evdev.py#L583-L584

This causes errors on earlier python versions, such as https://github.com/pythonarcade/arcade/issues/1225 .

To my understanding, the fix is reverting to using `typing.List` instead of subscripting the `list` type directly.

pushfoo:
Closes #614

### DIFF
--- a/pyglet/input/evdev.py
+++ b/pyglet/input/evdev.py
@@ -45,6 +45,8 @@ from ctypes import c_int32 as _s32
 from ctypes import c_int64 as _s64
 from concurrent.futures import ThreadPoolExecutor
 
+from typing import List
+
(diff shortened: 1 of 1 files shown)


caffeinepills (approved):
Looks good to me.

PR #633, opened 2022-06-15
[Win32] Stop setting `HWND_TOPMOST` for fullscreen

### LINKED ISSUE #566: Option to not force fullscreen to be topmost (alt+tab issue)
Right now, if you create a fullscreen window in pyglet, it sets the HWND_TOPMOST flag for it, which means it is always on top of everything else. This means that you cannot use alt+tab to switch between the pyglet window and other programs, which is bad if you want to do *anything* else (chatting, looking up things on the web etc) while a pyglet game is running.

My proposed solution would be to add an argument to enable fullscreen without setting the HWND_TOPMOST flag, maybe something like window(topmost=...), where the value can be set to True, False, or auto (current behaviour).

As a temporary workaround, monkey patching can be used:
```py
import pyglet.libs.win32.constants
pyglet.libs.win32.constants.HWND_TOPMOST = pyglet.libs.win32.constants.HWND_NOTOPMOST

# import only after patching
import pyglet
```

Note that this post applies to Windows ([pyglet.window.win32](https://github.com/pyglet/pyglet/blob/pyglet-1.5-maintenance/pyglet/window/win32/__init__.py)), I have not tested this behaviour on other platforms.

LennyPhoenix:
This patch removes the code that sets the `HWND_TOPMOST` flag when the window is in fullscreen mode.
If the `overlay` window style is used then the `HWND_TOPMOST` flag will still be used, so the original effect can still be achieved if intended.

This resolves #566

### DIFF
--- a/pyglet/window/win32/__init__.py
+++ b/pyglet/window/win32/__init__.py
@@ -267,18 +267,13 @@ class Win32Window(BaseWindow):
                                    GWL_EXSTYLE,
                                    self._ex_ws_style)
 
-        if self._fullscreen:
-            hwnd_after = HWND_TOPMOST
-        else:
-            hwnd_after = HWND_NOTOPMOST
-
         # Position and size window
         if self._fullscreen:
-            _user32.SetWindowPos(self._hwnd, hwnd_after,
+            _user32.SetWindowPos(self._hwnd, HWND_NOTOPMOST,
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

benmoran56:
Thanks Lenny, sorry for leaving this hanging.

LennyPhoenix:
No worries 🙂

PR #636, opened 2022-06-17
Cleanup of UniformBlocks and UniformBufferObjects

Square789:
Copypaste of the first commit message follows:
```
- Fix a bug where indices local to a loop instead of the actual block member indices were written to a uniform dict,
  which caused KeyErrors in the creation of the UniformBlock views.
- Shaders are now set up to read each uniform block from the uniform buffer binding point with the same index as the block.
  - This should not affect any existing shaders as none of them use more than one block (the WindowBlock), which is still
    sourced from 0 and whose UBO binds to 0.
  - Shaders do not have their block bindings configured by UniformBufferObjects anymore.
- Views are now created when required for each WindowBlock and not by each UBO when it is created. 
```
I've ran the sprite examples before and after and couldn't tell of any problems there, but this will ruin third-party code using the library if it relied on some of the exotic binding behavior of the UBOs or of their coupling to ShaderPrograms.
If the decoupling of UBOs to a specific ShaderProgram is undesired, feel free to drop the second commit.

### DIFF
--- a/pyglet/graphics/shader.py
+++ b/pyglet/graphics/shader.py
@@ -664,7 +664,7 @@ class UniformBlock:
         """
         if self.view_cls is None:
             self.view_cls = self._introspect_uniforms()
-        return UniformBufferObject(self, index)
+        return UniformBufferObject(self.view_cls, self.size, index)
 
     def _introspect_uniforms(self):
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):
Nice work, thanks!

PR #637, opened 2022-06-18
Cache list/tuple of font names

caffeinepills:
Cache list/tuple of font names to which font was found, rather than constantly checking font database if the font exists.

Passing a tuple of fonts, will cause a `have_font` lookup on the font renderer. Not all font renderers cache font names.

This PR will cache which font that was chosen from a given tuple or list, and use that instead to prevent excessive lookups.


einarf:
I can confirm this works great!

benmoran56 (approved):

PR #654, opened 2022-07-21
Type-annotate pyglet.math

TomFryers:
This does include a few non-type-annotation changes, namely:
* Make some `AssertionError`s `TypeError`s or `ValueError`s.
* Allow `__eq__`s and `__ne__`s to work on arbitrary types.
* Slightly change some formatting
* `raise AttributeError(...) from None`
* Fix some typo errors (`max` → `maximum`, `cross_product` → `cross`, `.negate()` → `-`).
* Some functions are more or less strict with the types they take. The matrix type constructors no longer require `Sized` , but their `__add__`, `__sub__` and `__matmul__` require `MatN`/`VecN`s.

Mypy still has some complaints. Two caused by [this issue](https://github.com/python/mypy/issues/8541), and the others caused by the matrices inheriting from `tuple`, which violates the Liskov substitution principle in several ways.

### DIFF
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -712,7 +712,7 @@ class Mat3(tuple[float, float, float, float, float, float, float, float, float])
 
     def __sub__(self, other: Mat3) -> Mat3:
         if not isinstance(other, Mat3):
-            raise TypeError("Can only add to other Mat3 types")
+            raise TypeError("Can only subtract from other Mat3 types")
         return Mat3(s - o for s, o in zip(self, other))
 
     def __pos__(self) -> Mat3:
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -673,7 +673,9 @@ class Mat3(tuple[float, float, float, float, float, float, float, float, float])
     the "@" operator.
     """
 
-    def __new__(cls, values: Iterable[float] | None = None) -> Mat3:
+    def __new__(
+        cls, values: Iterable[float] = (1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0)
+    ) -> Mat3:
         """Create a 3x3 Matrix
 
         A Mat3 can be created with a list or tuple of 9 values.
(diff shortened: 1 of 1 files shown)


benmoran56:
Thank you for working on this. Most of this looks good, but I do have some small concerns.

Mainly, the performance impact of the added `if` checks in the `__eq__` and `__new__` methods. 
These classes need to be as fast as possible and, since they're pure Python, there have been a lot of micro-optimizations. 
The `assert` checks in the `__new__` for instance become a NOP when running frozen or with Python -O, which is why it was left in. 
 
Performance is also the main reason the Matrix types are subclassed from `tuple`. 
The second reason that Mat3/4 subclass from `tuple` is that they can be directly used by ctypes in place of a ctypes array.
Using a generic Python object would add another step of first converting the values to a tuple (or ctypes array), before passing to OpenGL.  Because of that it seems unlikely that we can solve the Liskov complaints :thinking:

TomFryers:
I can remove the checks if necessary. I'd caution against it in the case of `__eq__` and `__ne__`, though, since it would break things like `Vec2 in list[Vec2 | None]`. The performance cost in this case is about 15 % from my tests.

benmoran56:
Ah, I see what you mean. I don't know how common those operations will be by users, but it's a good argument adding the check. 
How about keeping that in, but rolling back the checks in `__new__`?

TomFryers:
I've made the length check back into an assertion. Creating a matrix from a tuple is now about 5 % faster than before this PR. Using `Mat3()` to get an identity matrix is about 15 % slower, but this is probably much less common. With `-O`, both are now about 2 % faster.

pyglet/math.py
```diff
@@ -640,49 +687,56 @@ def __new__(cls, values=None) -> 'Mat3':
...
+        if not isinstance(other, Mat3):
+            raise TypeError("Can only add to other Mat3 types")
+        return Mat3(s + o for s, o in zip(self, other))
 
-    def __sub__(self, other) -> 'Mat3':
-        assert len(other) == 9, "Can only subtract from other Mat3 types"
-        return Mat3(tuple(s - o for s, o in zip(self, other)))
+    def __sub__(self, other: Mat3) -> Mat3:
+        if not isinstance(other, Mat3):
+            raise TypeError("Can only add to other Mat3 types")
```
benmoran56, line 715: Looks like it should be "subtract" here instead of "add".

TomFryers: Fixed.

benmoran56:
Thanks for the cleanups. I'm going to give this another look as soon as I can, and get it merged in.

benmoran56 (dismissed):

benmoran56:
Looks like using `tuple` directly isn't supported until Python 3.8, so might need to replace with `typing.Tuple` for now?
```py
class Mat3(tuple[float, float, float, float, float, float, float, float, float]):
```

benmoran56 (approved):

benmoran56:
Thank you for all of the extra changes Tom!

PR #664, opened 2022-08-01
Shapes refactor

benmoran56:
For better performance, move Shape positioning and rotation inside of the Shader.


benmoran56:
@pushfoo Since you've been looking at the shapes module recently, would you mind giving this a quick look?

pushfoo:
> Since you've been looking at the shapes module recently, would you mind giving this a quick look?

At first glance, nothing looks unreasonable, but I don't understand shaders in depth. I'll try to give this a more in-depth look in the next few days if nobody beats me to it.

Issue #659, opened 2022-07-28
Enhancement: Provide frozen equivalents of Vec2 and Vec3


pushfoo:
## What should be added?

`Vec2` and other classes were once immutable. They have since been made mutable for convenience and efficiency when changing attributes on objects.

It would be useful to add the following:
1. A frozen version of these classes that supports hashing in addition to the normal addition and subtraction operations
2. Convenience methods to make frozen copies of mutable originals

## Why

It would be useful for defining immutable constants interoperable with the mutable versions. For example, immutability would prevent people from accidentally changing directions if a constant is passed as an argument to a function (ie `game.constants.NORTH`, etc):
```py
# will raise errors if any attempt at in-place modification is made with += or similar
NORTH = FrozenVec2(0.0, 1.0)
WEST = FrozenVec(-1.0, 0.0)
NORTHWEST = NORTH + WEST
```

It would also be useful for defining immutable hashable versions that can be added to sets, dicts, or other containers and re-used later. This would be great for quickly prototyping path finding, but it would also be useful in other cases. For example, although the current version of Vec2 can be converted to tuples for storage, tuples cannot be added to mutable Vec2 instances:
```
>>> destinations.append(tuple(a_vec2_instance))
...
>>> position = Vec2(0,0)
>>> position += destinations.popleft()
Traceback (most recent call last):
  File "/usr/lib/python3.9/code.py", line 90, in runcode
```

pushfoo:
I'm working on this with two goals:
1. Implementing all current functionality for FrozenVec* classes
2. Unit tests for current & new classes so we can make changes safely

This is my current plan for type interactions:
```py
# binary operations always return a new instance
frozen  +  frozen   # creates a new frozen
frozen  +  mutable  # creates a new frozen
mutable +  frozen   # creates a new mutable
mutable +  mutable  # creates a new mutable

# in-place operations get more complicated
frozen  += frozen   # raises AttributeError
frozen  += mutable  # raises AttributeError
mutable += mutable  # modifies in-place
mutable += frozen   # modifies in-place
```
I'm not sure if I'll add the in-place operation behavior in the first PR.

benmoran56:
That looks like a good plan.
Your logic looks sound.

olehermanse:
@pushfoo 

```
frozen  +  mutable  # creates a new frozen
mutable +  frozen   # creates a new mutable
```

Why should these 2 be different?

TomFryers:
[...]

PR #666, opened 2022-08-05
Fix small repeated typo in math.py

pushfoo:
While reading through source to better understand #659, I noticed that "swizzed" was used instead of "swizzled".

Changes: `%s/swizzed/swizzled/g` in `pyglet/math.py`


benmoran56 (dismissed):
Thanks!

benmoran56 (approved):

PR #667, opened 2022-08-05
Fix unit test errors caused by importing mock directly

pushfoo:
For unclear reasons, the Windows test environment is ok with `mock` being imported directly rather than from `unittest` as normal. bbace77 broke tests for platforms other than windows by importing mock directly.

This PR should restore tests to passing.


benmoran56 (dismissed):

pushfoo:
@benmoran56 I think we might need to make some changes in conftest to do this more properly, but the most recent changes I just pushed seem to make all platforms pass.

benmoran56 (approved):

benmoran56:
Thanks for that.
I guess I also had the standalone mock package on my local machine.

pushfoo:
I don't think it's just your system. The Windows tests passed on GitHub's CI as well.

PR #668, opened 2022-08-06
Centralize min py version in doc & remove 3.6 support

pushfoo:
This PR centralizes the minimum python version in execution and doc from a single source of truth: a tuple constant in `pyglet/__init__.py`. It also bumps the minimum version to 3.7. It makes sense do this now because:
* #654 broke unit tests & compatibility for 3.6 in `master` branch
* Python 3.7 reaches EOL in 10 months, so we may need to do this again soon

There may still be legacy backward-compatibility code buried in dependencies or submodules. Also, README.md currently still has to be updated manually.

Changes:
* Bump minimum version in `pyglet/__init__.py` to 3.7 as a tuple constant
* Use [sphinx substitutions](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#substitutions) to provide auto-updating version numbers across doc
* Bump CI to use 3.7 as the minimum version
* Remove 3.6 from setup.py tags
* Add 3.10 to setup.py tags
* Update mentions of 3.6 throughout the doc using sphinx substitutions


doc/internal/testing.rst
```diff
@@ -50,7 +50,7 @@ The pyglet test suite is based on the `pytest framework <http://pytest.org>`_.
 It is preferred to use a virtual environment to run the tests.
 For instructions to set up virtual environments see :doc:`virtualenv`.
 Make sure the virtual environment for the Python version you want to
-test is active. It is preferred to run the tests on 3.6+
+test is active. It is preferred to run the tests on 3.7+
```
benmoran56, line 53: To future proof this, perhaps we should just say something like: 
"To make sure changes are compatible with all supported Python versions, tests should ideally be run on all versions".

doc/internal/virtualenv.rst
```diff
@@ -104,7 +104,7 @@ Setting up
 
 Make sure you download and install:
 
-    - `Python 3.6+  <http://www.python.org/downloads/windows/>`_
+    - `Python 3.7+  <http://www.python.org/downloads/windows/>`_
```
benmoran56, line 107: Maybe we can omitt the version here, so we don't have to keep updating it in the future.

benmoran56 (changes requested):

benmoran56 (approved):

benmoran56 (approved):

doc/internal/testing.rst
pushfoo: It's future-proofed using a central source of truth on minimum versions. See new changes.

pushfoo:
I think this might break testing against 3.6 in its current state for the 1.5 branch. This probably needs a closer look if 1.5.X is still supposed to support that.

benmoran56 (dismissed):
Looks good to me.

benmoran56 (approved):

PR #670, opened 2022-08-09
Fix typos in Vec scale/lerp docstrings & add scale/lerp docstrings for Vec4

pushfoo:
This PR only contains doc fixes for things I noticed while working on #659.


benmoran56 (approved):
Looks good, thanks!

PR #675, opened 2022-08-17
Initial Text rotation implementation

Cleptomania:
This is an initial implementation of Text rotation. This works by adding a rotation attribute to `TextLayout` and adding `rotation` and `anchor` inputs to the layout/decoration shaders.

The anchor point is the X/Y position of the entire TextLayout object, and is the point that the rotation is performed around. This enables rotation of the entire layout around a point.

The main thing that hasn't been well tested yet are scrollable text layout and ensuring that the scissor area works with this change if you rotate a scrollable area(though perhaps there are other problems with rotating a scrollable area, I'm not that familiar with it's usage). 

To support using this the `Label` and `DocumentLabel` have had a `rotation` parameter added to set this. It can be updated after the fact using the `rotation` property on the base TextLayout class. I'm very unfamiliar with HTMLLabel so I have not added the parameter to this, however it could probably just be added and passed down to the base `DocumentLabel` class.

### DIFF
--- a/pyglet/text/__init__.py
+++ b/pyglet/text/__init__.py
@@ -236,7 +236,7 @@ class DocumentLabel(layout.TextLayout):
     def __init__(self, document=None,
                  x=0, y=0, width=None, height=None,
                  anchor_x='left', anchor_y='baseline',
-                 multiline=False, dpi=None, batch=None, group=None):
+                 multiline=False, dpi=None, batch=None, group=None, rotation=0):
         """Create a label for a given document.
 
         :Parameters:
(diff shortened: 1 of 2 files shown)


benmoran56 (approved):

benmoran56:
Thanks for this.

The basic TextLayout and ScrollableTextLayouts works fine, but the IncrementalTextLayout needs a tiny bit of work. I'll see if I can address that myself. 

For the Labels, the HTMLLabel will work the same as the others. The Labels are all essentially the same, but they format the underlying document object in differently.

Issue #652, opened 2022-07-18
shapes.Polygon class appears to have a broken position property


pushfoo:
## Versions affected
* 2.0.dev20 from pip
* tip of master branch
* 1.5.26 from pip

## System info
Running over Gnome/Wayland on Debian 11
```
vendor: Intel
renderer: Mesa Intel(R) UHD Graphics 620 (KBL GT2)
version: (4, 6)
python: 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110]
platform: linux
```

## Expected behavior
Setting the position property moves all vertices of a polygon

## Actual behavior
Before setting position:
![image](https://user-images.githubusercontent.com/36696816/179612727-427b22f6-02ca-43c9-b941-ed6039e3396c.png)
After setting position to center of screen:
![image](https://user-images.githubusercontent.com/36696816/179612804-98f7edd1-d15a-47b2-9910-0986264aca20.png)

## Example code
```py
import pyglet                                                                                           
from pyglet import shapes

class PolygonIsBroken(pyglet.window.Window):
    def __init__(self, width, height):
        super().__init__(width, height, "Broken Polygon Example")

        self.batch = pyglet.graphics.Batch()

        self.polygon = shapes.Polygon(
            [0, 0], [-100, 100], [100, 100],
            batch=self.batch
        )

    def on_draw(self):
        self.clear()
        self.batch.draw()

    def on_key_press(self, key, modifiers):
        self.polygon.position = self.width / 2, self.height / 2 

if __name__ == "__main__":
    demo = PolygonIsBroken(800, 600)
    pyglet.app.run()

```

Ross-TheBoss:
I have created a pull request to fix this issue.

benmoran56:
Should now be fixed in 1.5 and master.

PR #676, opened 2022-08-19
Pyglet 1.5 maintenance

Ross-TheBoss:
Fixes the broken position property by ensuring all coordinates are updated when the first coordinate (the position) is changed.

Addresses issue: #652


benmoran56 (approved):
Looks good, thank you!

PR #682, opened 2022-09-07
Fix "sticky buttons"

### LINKED ISSUE #681: add on_activate event to KeyStateHandler
Please add on_activate event to KeyStateHandler, with reset to data dict(setting the data dict to a new empty one), because when you leave the window and you holded down a key then when you release it its still holded down, and the only way to stop it is to press the key once

LennyPhoenix:
This PR makes 2 changes:

- Adds the `on_deactivate` event to the mouse and key state handlers, meaning that when the window loses focus it will treat it like all keys and mouse buttons had just been released. This prevents issues with "sticky buttons" such as when alt+tabbing when holding a key down.
- Imports the `pyglet.window.mouse` module by default.

Resolves #681

### DIFF
--- a/pyglet/window/__init__.py
+++ b/pyglet/window/__init__.py
@@ -129,6 +129,7 @@ import warnings
 
 import pyglet
 import pyglet.window.key
+import pyglet.window.mouse
 import pyglet.window.event
 
 from pyglet import gl
--- a/pyglet/window/key.py
+++ b/pyglet/window/key.py
@@ -67,6 +67,8 @@ from pyglet import compat_platform
 class KeyStateHandler(dict):
     """Simple handler that tracks the state of keys on the keyboard. If a
     key is pressed then this handler holds a True value for it.
+    If the window loses focus, all keys will be reset to False to avoid a
+    "sticky" key state.
 
     For example::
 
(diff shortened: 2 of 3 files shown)


caffeinepills:
Doesn't this just create an issue with the reverse? If I hold a key down, alt tab, then tab back, it's no longer treated as being held down?

csd4ni3l:
but already if someone holds down a key in that window and tab to another one why would they want it to be holded down in that window too?

LennyPhoenix:
> Doesn't this just create an issue with the reverse? If I hold a key down, have the window lose focus, then tab back, it's no longer treated as being held down even though the key is?

I'd personally rather have the key release itself than it appear stuck, in most games having a key stuck is going to lead to more annoyance than just having it no-longer pressed.

I'm just not sure how we would fix an issue like this, is there a way to detect keyboard events even when the window is not focused? The only way I can see a full solution to this is keeping track of key presses even when unfocused, then checking if the window is active in `__getitem__` (probably with some flag set by `on_activate` and `on_deactivate`)

benmoran56 (approved):

benmoran56:
Thank you Lenny!

PR #691, opened 2022-10-06
Make OpenGL the default API

einarf:
The current change will do the following:

* Ensure the EGL backend works without explicitly passing `opengl_api="gl"` in the config.
* Report back to user when the selected opengl api is not supported (or misconfigured)

Things broke when we added GLES. In some instances the context would be created even if `opengl_api` is `None` since opengl is assumed to be default by some backends. This is not always the case. Making sure the opengl api is explicitly set means less trouble.

### DIFF
--- a/pyglet/gl/base.py
+++ b/pyglet/gl/base.py
@@ -120,7 +120,7 @@ class Config:
     major_version = None
     minor_version = None
     forward_compatible = None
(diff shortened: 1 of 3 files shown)


benmoran56 (approved):
Great! And those extra exceptions will be useful.

PR #697, opened 2022-10-30
Fixes for Layouts

caffeinepills:
* Fix for Layout position for expanded vertices.

* Fix for IncrementalTextLayout when it starts with no text.
- The current fix only updates if the count changes, if 1 group_cache is removed and 1 is added, this will not update. I haven't experienced this occurring as the group_cache is not cleared in anyway unlike the regular TextLayout. (It is cleared every update in a TextLayout, but not for other layouts, unsure this needs to be done still). If group_cache needs to be cleared for layouts with scissor areas, the layouts will need to update the scissor area every update call.

### DIFF
--- a/pyglet/text/layout.py
+++ b/pyglet/text/layout.py
@@ -1131,8 +1131,8 @@ class TextLayout:
             dz = z - self._z
             for vertex_list in self._vertex_lists:
                 vertices = vertex_list.position[:]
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):
Nice find with the _update_scissor_area bug.

PR #714, opened 2022-11-29
Fix ShaderProgram.vertex_list

alex-khod:
`ShaderProgram.vertex_list` passes `self._id` as `program` argument instead of `self` to batch.get_domain.
This leads to `ShaderGroup` being created with `int` type argument instead of `ShaderProgram` when no `group` argument is passed to `vertex_list`.
The following snippet results in `self.program.use()... AttributeError: 'int' object has no attribute 'use'` before the change.
```py
import pyglet

program = pyglet.graphics.get_default_shader()

batch = pyglet.graphics.Batch()
vlist = program.vertex_list(1, pyglet.gl.GL_POINTS, batch, group=None, position=('f', [100, 100]), colors=('B', [255, 255, 255,255]))
@window.event
def on_draw():
    window.clear()
    batch.draw()

pyglet.app.run()
```

### DIFF
--- a/pyglet/graphics/shader.py
+++ b/pyglet/graphics/shader.py
@@ -580,7 +580,7 @@ class ShaderProgram:
                 raise ShaderException(f"\nThe attribute `{name}` doesn't exist. Valid names: \n{list(attributes)}")
 
         batch = batch or pyglet.graphics.get_default_batch()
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

benmoran56:
Thanks for this!

PR #717, opened 2022-12-02
Fix for get_default_screen() 

caffeinepills:
`display.get_default_screen()` doesn't get the primary monitor. It just gets the first monitor in the list, which is not guaranteed to be the primary.

From what I have found in testing, the display at position 0, 0 is considered the primary in almost all cases.

I have tested in a 3 monitor setup and my primary is on "Display 2", but my default prior to the fix was always considered Display 1.  After testing, (on Windows) now, even if I switch to Display 3 being primary, Display 3 becomes 0, 0 and is properly recognized.

If the primary is not found for whatever reason, it will fallback to default behavior of choosing first in the list. 

This just needs confirmation on Linux and Mac that the behavior is the same.

This can be backported to 1.5

### DIFF
--- a/pyglet/canvas/base.py
+++ b/pyglet/canvas/base.py
@@ -98,12 +98,18 @@ class Display:
         raise NotImplementedError('abstract')
 
     def get_default_screen(self):
-        """Get the default screen as specified by the user's operating system
+        """Get the default (primary) screen as specified by the user's operating system
         preferences.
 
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

benmoran56:
Thanks. I though we fixed this in the past, but maybe that was Linux or macOS. I should probably check on Linux to confirm.

PR #718, opened 2022-12-02
Fix class links

xantares:


benmoran56 (approved):

benmoran56:
Thank you very much! Documentation improvements are always welcome.

PR #737, opened 2022-12-22
DW Enhancements

caffeinepills:
Enhancements to DirectWrite. Feel free to test and let me know if we need to make any adjustments.

- Legacy font naming parser. 
  Requires `pyglet.options["dw_legacy_naming"] = True`
  Attempt to parse files by the passed name, to best match legacy RBIZ naming. For instance asking for things like "Arial Narrow" rather than "Arial" with a condensed stretch or "Arial Black" instead of "Arial" with a weight of black. This may enhance naming compatibility cross-platform for select fonts as older font renderers went by this naming scheme. 
  
  Starts by parsing the string for any known style names (which we can expand), and searches all font collections for a matching RBIZ name. If found, will go through the faces attempting to find the closest match to the passed styles. Will find the second closest match if a perfect match is not found.

  Due to the high variation of styles and limited capability of some fonts, there is no guarantee the second closest match will be exactly what the user wants which is why this is an option. Any suggestions on improvement are welcome. Enabling `debug_font` option can provide information on what settings are being selected.

- Filename property to a font
Will return 'Not Available' if the path cannot be queried (such as from custom fonts).

- Also expands the render width to account for italics.
There is no way to get the actual width of an italic from glyph properties. So we use an estimate at the cost of additional texture memory. The only solution I have found online was to render it to a larger texture, then scan every pixel trying to find the furthest pixel to the right to get the width, which would be a lot slower.


benmoran56 (approved):

benmoran56 (approved):

benmoran56 (approved):

benmoran56:
Looks good to me. We could deprecate that pyglet option at some point in the future if it proves unnecessary.

PR #738, opened 2022-12-25
Change math module to behave like GLSL

### LINKED ISSUE #733: Matrix multiplication ordering is confusing
Perhaps bug, perhaps just my confusion about conventions:

The transformations provided by the pyglet.math matrix classes right-multiply with other matrices. That is, as seen in the 3dmodel example:

```python
model_logo.matrix = rot_x @ rot_y @ rot_z @ trans
```

The "composition" happens left-to-right.

This should imply matrices right-multiply (row) vectors too, so that we would have e.g. `vec @ rot @ trans`. However the Vec classes do not implement `@`, and instead the matrices left-multiply with vectors as if they were right-multiplying (note the sums down the columns):

```python
    def __matmul__(self, other):
        if isinstance(other, Vec4):
            # Columns:
            c0 = self[0::4]
            c1 = self[1::4]
            c2 = self[2::4]
            c3 = self[3::4]
            return Vec4(sum(map(_mul, c0, other)),
                        sum(map(_mul, c1, other)),
                        sum(map(_mul, c2, other)),
                        sum(map(_mul, c3, other)))
```

So one has to write `(rot @ trans) @ vec` to get what is really vec * rot * trans.

In OpenGL, GLSL vectors are actually column vectors, i.e. matrices **left**-multiply. (But it's a bit confusing since by default matrices must be fed into uniforms in column-major order, so in memory they are stored transposed.)

I feel that pyglet should try to be consistent here, at least internally, but ideally the same as GLSL. At least it would be good to have `@` be associative, which it isn't in `(rot @ trans) @ vec`.

Demo:

```
>>> from pyglet.math import Mat3, Vec3
>>> swap_xy = Mat3((0,1,0, 1,0,0, 0,0,1))
>>> scale_x = Mat3((2,0,0, 0,1,0, 0,0,1))
>>> swap_xy @ scale_x
Mat3(0, 1, 0)
    (2, 0, 0)
    (0, 0, 1)
>>> (swap_xy @ scale_x) @ Vec3(0,1,0)
Vec3(2, 0, 0)
```

Thanks 🙇‍♂️

tmewett:
Note this is API-breaking.

Closes #733

### DIFF
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -861,17 +861,6 @@ class Mat4(tuple):
                     vector[0], vector[1], vector[2], 1.0))
 
     @classmethod
-    def look_at_direction(cls: type[Mat4T], direction: Vec3, up: Vec3) -> Mat4T:
-        vec_z = direction.normalize()
-        vec_x = direction.cross(up).normalize()
-        vec_y = direction.cross(vec_z).normalize()
-
-        return cls((vec_x.x, vec_y.x, vec_z.x, 0.0,
-                    vec_x.y, vec_y.y, vec_z.y, 0.0,
-                    vec_x.z, vec_z.z, vec_z.z, 0.0,
-                    0.0, 0.0, 0.0, 1.0))
-
-    @classmethod
     def look_at(cls: type[Mat4T], position: Vec3, target: Vec3, up: Vec3):
         f = (target - position).normalize()
         u = up.normalize()
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -696,30 +696,30 @@ class Mat3(tuple):
 
     def __matmul__(self, other):
         if isinstance(other, Vec3):
-            # Columns:
-            c0 = self[0::3]
-            c1 = self[1::3]
-            c2 = self[2::3]
-            return Vec3(sum(map(_mul, c0, other)),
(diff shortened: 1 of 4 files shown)


pyglet/math.py
```diff
@@ -768,7 +768,11 @@ def orthogonal_projection(
         z_near: float,
         z_far: float
     ) -> Mat4T:
-        """Create a Mat4 orthographic projection matrix."""
+        """Create a Mat4 orthographic projection matrix for use with OpenGL.
```
benmoran56, line 771: I wonder if the detailed explanation is necessary here, as the context and use case should be understood already. 
Perhaps it would be enough to just describe it as a transformation matrix? Something like: 
```Create a 4x4 transformation matrix for orthographic projection.```
or 
```Create a 4x4 orthographic projection transformation matrix.```

What do you think?

tmewett: A projection is a transformation, so I don't think that removes any ambiguity

I don't think it's bad to have an extra sentence to be precise, but I can remove that commit if you would like

benmoran56 (approved):

pyglet/math.py
benmoran56: No worries Tom, I can make some tweaks on my end later, but lets merge this in for now.

PR #742, opened 2023-01-04
Remove left-over print statement

Cleptomania:
There was a `print()` statement that seems to have been left-over from development work in loading freetype fonts. This causes some unwanted output for end users.


benmoran56 (approved):

benmoran56:
Oops! Thanks for the pull request.

PR #744, opened 2023-01-05
more operations for bezier curve

jason-bowen-zheng:
As the title showed, more operations were added to the curve.

First, we are allowed to change the control point.

The second thing, which I think it's unnecessary, is a `t` argument . It tells the program to draw a **part** of the curve. Maybe it helps us to create animations?

### DIFF
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -659,11 +659,15 @@ class BezierCurve(ShapeBase):
             # Calculate the points of the curve:
             points = [(x + self._make_curve(t / self._segments)[0],
                        y + self._make_curve(t / self._segments)[1]) for t in range(self._segments + 1)]
+            trans_x, trans_y = points[0]
+            trans_x += self._anchor_x
+            trans_y += self._anchor_y
+            coords = [[x - trans_x, y - trans_y] for x, y in points]
 
             # Create a list of doubled-up points from the points:
             vertices = []
-            for i in range(len(points) - 1):
-                line_points = *points[i], *points[i + 1]
+            for i in range(len(coords) - 1):
+                line_points = *coords[i], *coords[i + 1]
                 vertices.extend(line_points)
 
         self._vertex_list.vertices[:] = vertices
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -112,7 +112,7 @@ vertex_source = """#version 150 core
     {
         m_translate[3][0] = translation.x;
         m_translate[3][1] = translation.y;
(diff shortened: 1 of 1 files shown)


pyglet/shapes.py
```diff
@@ -602,10 +602,10 @@ def draw(self):
 class BezierCurve(ShapeBase):
     _draw_mode = GL_LINES
 
-    def __init__(self, *points, segments=None, color=(255, 255, 255, 255), batch=None, group=None):
```
benmoran56, line 605: Maybe we can just do `segments=100` here, instead of None?

benmoran56:
These are reasonable changes, thank you.

I'm not very familiar with this myself, but do you think users will understand the meaning of `t`?
Would it be better to give it a more descriptive name, such as "position"?

jason-bowen-zheng:
I want to use `lambda_` as the name, which is used in the midpoint formula(or the similar one, but I don't know the name).

the `λ` symbol usually used to extend or shorten a line segment by a specific length.

benmoran56:
It might be confusing because of the `_` in the name. Maybe just leaving it as `t` is fine. 
We can just explain what it means in the docstring.

jason-bowen-zheng:
I have created a demo for the using of `t`:
```python
import pyglet
from pyglet import shapes
from pyglet.clock import schedule_interval

curve = shapes.BezierCurve((50, 50), (100, 150), (150, 100), (200, 200), t=0.01)
lines = [
    shapes.Line(50, 50, 100, 150, color=(255, 0, 0)),
    shapes.Line(100, 150, 150, 100, color=(255, 0, 0)),
    shapes.Line(150, 100, 200, 200, color=(255, 0, 0)),
]

def update(dt):
    if curve.t >= 1:
        curve.t = 0.01
    else:
        curve.t += 0.001

@window.event
def on_draw():
    window.clear()
    schedule_interval(update, 0.001)
    for obj in lines:
        obj.draw()
    curve.draw()

pyglet.app.run()
```

Can you give me some advice on the documents of `t`?

benmoran56:
Sorry for the delay. The changes you have made look good. The docstrings look easily understandable.
This looks OK to me. I can merge it in if you are ready. 

By the way, it's not a good idea to put `schedule_interval(...)` inside of `on_draw()`. It will keep re-scheduling it each time the window draws.
Instead, place it somewhere before `pyglet.app.run()`:

```py
schedule_interval(update, 0.001)
pyglet.app.run()
```

jason-bowen-zheng:
Thank for your advice, please merge it as soon as possible.

benmoran56:
Done!

Thanks again.

PR #747, opened 2023-01-08
TextEntry on_mouse_press (minor change)

onuelito:
Made it so when you click elsewhere on the window it deactivates the TextEntry. This prevents the weird effect of two TextEntry  being active at the same time _(for instance: clicking a TextEntry while another is active allowing you to write in both at the same time)_.


benmoran56 (approved):

benmoran56:
Thanks, looks good!

PR #753, opened 2023-01-16
Update year

einarf:
Update all headers, licence and docs to 2023

Some recently added moduels were missing headers. Those were added. Older files with missing header was not changed.


einarf:
We decided to purge these headers and rely on the project-wide license.

PR #782, opened 2023-02-27
Various fixes for DirectInput event handling

caffeinepills:
Fixes various issues including fallback behavior in certain input scenarios:

1. If you were to disable the shadow window, and import `pyglet.input` before you created a normal Window this would cause errors.

Now behavior is:
Attempt to find a Window to register events to. If no window found, skip.
When attempting to iterate `_di_manager` devices. Such as (`get_joysticks`, `get_controllers`, or `get_controllers` on `ControllerManager`), and Window event is not registered:
   a) Look for Window and attempt to register the event again.
   b) If failure, produce a warning.
   c) Refresh current list of devices in the DI Device Manager.

2. This also fixes issue with Connection Manager not having a list of prepopulated devices if above scenario played out.

3. Added missing `UnregisterDeviceNotification` so we can register events to a new window if needed.
 
4. Also now adds itself as a window handler to determine if the window it's attached to is closed. In a multi-window scenario it will attempt to register itself to another open window if found. (May need a better way to do this, since `on_close` even can be completely overridden)

5. Now that COM thread mode is STA by default, we cannot run xinput's COM checks inside the handled event without an error. Created a dummy dispatch event so we can run in the main thread outside of the send event to fix this.

### DIFF
--- a/pyglet/image/__init__.py
+++ b/pyglet/image/__init__.py
@@ -818,10 +818,15 @@ class ImageData(AbstractImage):
 
         # Unset GL_UNPACK_ROW_LENGTH:
         glPixelStorei(GL_UNPACK_ROW_LENGTH, 0)
+        self._default_region_unpack()
+
         # Flush image upload before data get GC'd:
         glFlush()
 
     def _apply_region_unpack(self):
+        pass
+
+    def _default_region_unpack(self):
         pass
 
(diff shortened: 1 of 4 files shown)


benmoran56 (approved):

benmoran56:
Looks good to me!

PR #784, opened 2023-03-01
TextEntry Patches (1.5.x)

onuelito:
- added `_active_entry` property to `Frame` class for the purpose of updating `TextEntry` regardless of the mouse being far from it
 - removed and added lines to `TextEntry` class for the same purpose as above

### DIFF
--- a/pyglet/gui/frame.py
+++ b/pyglet/gui/frame.py
@@ -63,6 +63,7 @@ class Frame:
         self._cell_size = cell_size
         self._cells = {}
         self._active_widgets = set()
+        self._active_entry = None
         self._order = order
         self._mouse_pos = 0, 0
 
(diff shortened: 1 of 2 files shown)


pyglet/gui/frame.py
```diff
@@ -87,49 +88,84 @@ def remove_widget(self, widget):
...
         """Pass the event to any widgets within range of the mouse"""
+
+        if self._active_entry: 
+            self._active_entry.on_mouse_press(x, y, buttons, modifiers)
+            self._active_entry = self._active_entry if self._active_entry._focus else None
+        
+
         for widget in self._cells.get(self._hash(x, y), set()):
-            widget.on_mouse_press(x, y, buttons, modifiers)
+            if type(widget).__name__ == 'TextEntry':
```
Lattay, line 98: This should either be `if type(widget) == TextEntry` or `if isinstance(widget, TextEntry)` (more likely the latter).

pyglet/gui/frame.py
```diff
@@ -87,49 +88,84 @@ def remove_widget(self, widget):
...
         for widget in self._active_widgets:
             widget.on_mouse_release(x, y, buttons, modifiers)
         self._active_widgets.clear()
 
     def on_mouse_drag(self, x, y, dx, dy, buttons, modifiers):
         """Pass the event to any widgets that are currently active"""
+        if self._active_entry: self._active_entry.on_mouse_drag(x, y, dx, dy, buttons, modifiers)
+
         for widget in self._active_widgets:
+            if widget == self._active_entry: continue
```
Lattay, line 120: Can you explain this special case in `mouse_drag` but not in `mouse_release`?

pyglet/gui/frame.py
```diff
@@ -87,49 +88,84 @@ def remove_widget(self, widget):
...
+
+        if self._active_entry: 
+            self._active_entry.on_mouse_press(x, y, buttons, modifiers)
+            self._active_entry = self._active_entry if self._active_entry._focus else None
+        
+
         for widget in self._cells.get(self._hash(x, y), set()):
-            widget.on_mouse_press(x, y, buttons, modifiers)
+            if type(widget).__name__ == 'TextEntry':
+                if self._active_entry == widget: continue
```
Lattay, line 99: Single line if blocks are very unusual in Python, and they don't seem to be used anywhere in pyglet. Maybe use black for automatic formatting.

Lattay (commented):
Hi,
I am not (yet?) a contributor of Pyglet, but I am also interested in the GUI module and planning on contributing to it, so I read your commit. I have a few suggestions that may improve your contribution.

pyglet/gui/frame.py
onuelito: I forgot to add the line

PR #792, opened 2023-03-11
Uniforms: Add missing sampler types

einarf:
* Cube samplers
* INT and UNIT versions of each sampler
* Multisample versions
* All array versions


caffeinepills (approved):

caffeinepills:
Looks good to me.

PR #796, opened 2023-03-19
fix args when calling super class

drakbar:
[`SpriteGroup`](https://github.com/pyglet/pyglet/blob/e374e81c7b2128b24da52707e4f233b19749c40e/pyglet/sprite.py#L177)

SpriteGroup class constructor only has 5 parameters in its definition. The call to the super class constructor provides 6 arguments.  Removing that argument makes the up-stream arguments superfluous, thus removed them as well.


benmoran56 (approved):

benmoran56:
Thanks!

drakbar:
Not a problem!

PR #797, opened 2023-03-19
fix is_open in x11_input

drakbar:
[`Device`](https://github.com/pyglet/pyglet/blob/e374e81c7b2128b24da52707e4f233b19749c40e/pyglet/input/base.py#L43)

`XInputDevice` is a sub-class of `Device`.  The property (**is_open**) is inherited, however the parent class defines this property as **_is_open**. These changes correct this difference.


benmoran56 (approved):
Thanks, good catch.

PR #799, opened 2023-03-21
window: on_draw doesn't take "dt"

kivikakk:
Shrimple as that 🦐


benmoran56:
Hi @kivikakk, 
Thank you for your pull request. We did not add `dt` to on_draw because it would break existing code.
Instead, the `on_refresh` method exists, which _does_ pass a `dt` value.

kivikakk:
My pleasure! :heart:

PR #804, opened 2023-04-02
Update make.py

gran4:
no need for empty print. fixed Dictionary(line 64)

### DIFF
--- a/make.py
+++ b/make.py
@@ -61,7 +61,7 @@ def _print_usage():
 
 
 if __name__ == '__main__':
-    avail_cmds = dict(clean=clean, dist=dist, docs=docs)
+    avail_cmds = {clean=clean, dist=dist, docs=docs}
     try:
         command = avail_cmds[sys.argv[1]]
     except IndexError:
@@ -69,7 +69,6 @@ if __name__ == '__main__':
         _print_usage()
     except KeyError:
         print('Unknown command:', sys.argv[1])
-        print()
         _print_usage()
     else:
         command()
--- a/make.py
+++ b/make.py
@@ -68,7 +68,7 @@ if __name__ == '__main__':
         # Invalid number of arguments, just print help
         _print_usage()
     except KeyError:
-        print('Unknown command:', sys.argv[1])
+        print(f"Unknown command: {sys.argv[1]}\n")
         _print_usage()
     else:
         command()
--- a/make.py
+++ b/make.py
@@ -54,8 +54,7 @@ def dist():
 
 def _print_usage():
     print('Usage:', op.basename(sys.argv[0]), '<command>')
-    print('  where commands are:', ', '.join(avail_cmds))
-    print()
+    print('  where commands are:', ', '.join(avail_cmds), "\n")
     for name, cmd in avail_cmds.items():
         print(name, '\t', cmd.__doc__)
 
--- a/make.py
+++ b/make.py
@@ -61,7 +61,7 @@ def _print_usage():
 
 
 if __name__ == '__main__':
-    avail_cmds = {clean=clean, dist=dist, docs=docs}
+    avail_cmds = dict(clean=clean, dist=dist, docs=docs)
     try:
         command = avail_cmds[sys.argv[1]]
     except IndexError:


make.py
```diff
@@ -61,15 +61,14 @@ def _print_usage():
...
 if __name__ == '__main__':
-    avail_cmds = dict(clean=clean, dist=dist, docs=docs)
+    avail_cmds = {clean=clean, dist=dist, docs=docs}
     try:
         command = avail_cmds[sys.argv[1]]
     except IndexError:
         # Invalid number of arguments, just print help
         _print_usage()
     except KeyError:
         print('Unknown command:', sys.argv[1])
```
pushfoo, line 71: Removing the extra line hurts the readability of the output since the newline separates the error message from the usage information:
```console
$ python make.py badcommand
Unknown command: badcommand

Usage: make.py <command>
  where commands are: clean, dist, docs

clean    Clean up all build & test artifacts, and generated documentation.
dist     Create all files to distribute Pyglet
docs     Generate documentation
```
```suggestion
        print(f"Unknown command: {sys.argv[1]}\n")
```

pushfoo (changes requested):
tl;dr you can remove the print statement, but the extra newline is necessary for readability

make.py
```diff
@@ -61,15 +61,14 @@ def _print_usage():
 
 
 if __name__ == '__main__':
-    avail_cmds = dict(clean=clean, dist=dist, docs=docs)
+    avail_cmds = {clean=clean, dist=dist, docs=docs}
```
pushfoo, line 64: My apologies for not flagging this in the previous review.  To my knowledge, this is not valid syntax in any current version of Python.

gran4: I don't think the tiny performance benefit is worth it. It's just not pythonic.

benmoran56: As pushfoo pointed out, your proposed change is not valid Python syntax. 

I don't agree that the original code is non-Pythonic.

pushfoo, line 64: Since you requested a re-review...

**tl;dr Please revert this. The line as originally written is probably the most readable & maintainable option.**

#### 1. The line, as written in the current PR revision, is invalid grammar

Try importing the file:
```py
>>> from make import *
  File "/usr/lib/python3.9/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "/opt/pycharm-community-2022.1.1/plugins/python-ce/helpers/pydev/_pydev_bundle/pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
  File "/home/user/src/pyglet/make.py", line 64
    avail_cmds = {clean=clean, dist=dist, docs=docs}
                       ^
SyntaxError: invalid syntax
```
The problem, isolated from pyglet:
```py
>>> {"a"="b"}
  File "/usr/lib/python3.9/code.py", line 63, in runsource
    code = self.compile(source, filename, symbol)
  File "/usr/lib/python3.9/codeop.py", line 178, in __call__
    return _maybe_compile(self.compiler, source, filename, symbol)
  File "/usr/lib/python3.9/codeop.py", line 106, in _maybe_compile
    raise err1
  File "/usr/lib/python3.9/codeop.py", line 93, in _maybe_compile
    code1 = compiler(source + "\n", filename, symbol)
  File "/usr/lib/python3.9/codeop.py", line 143, in __call__
    codeob = compile(source, filename, symbol, self.flags, True)
  File "<input>", line 1
    {"a"="b"}
        ^
SyntaxError: invalid syntax
```

#### 2. Using syntax of the form `{clean: clean}` likely isn't what you want

tl;dr it  breaks `make.py`

If you meant to use `:`, you may be remembering syntax from another language. In Python, it will not do what you expect since `{name: name}` uses the value of `name` as the key to its own value. Since functions are hashable, using it in this case will produce a dictionary mapping functions to functions. 

This produces two issues which break `make.py`:
1. String lookup of the commands will fail since the hash of a function will never match the hash of a string
2. Command listing will also break since it iterates over the keys, but expects them to be strings

This is the output when you use `:` in place of `=` in this line:
```console
$ python make.py docs
Unknown command: docs

Usage: make.py <command>
  File "/home/user/src/pyglet/make.py", line 66, in <module>
    command = avail_cmds[sys.argv[1]]
KeyError: 'docs'

During handling of the above exception, another exception occurred:

  File "/home/user/src/pyglet/make.py", line 73, in <module>
    _print_usage()
  File "/home/user/src/pyglet/make.py", line 57, in _print_usage
    print('  where commands are:', ', '.join(avail_cmds))
TypeError: sequence item 0: expected str instance, function found
```

#### 3. String key syntax is less readable & less maintainable

If you wanted to use something other than the `dict(...)` syntax, this is the other option:
```py
avail_cmds = {'clean': clean, 'dist': dist, 'docs': docs}
```
It's longer, and the quotes make it harder to read and edit. I would also argue that it's less pythonic.

pushfoo (changes requested):
Although you've requested a re-review, the issue is not resolved. I've written a more thorough explanation of why in a new comment. Please revert the `avail_cmds` change.

benmoran56 (approved):

PR #808, opened 2023-04-17
Attempts to add docs strings and Type hints

gran4:


gran4:
I'm confused about the types of _DefaultLoader.

benmoran56:
Thanks for working on this. I have a few suggestions, but I'll add them later as in-line comments. 

There is an issue, however. Merging this is blocked because automated tests are failing. There is a circular import leading back to `pyglet/tests/conftest.py`. This file does import several fixtures, including importing the event_loop. 

I'm not positive, but the `from __future__ import annotations` might be the culprit here. Strange that it only happens on OSX and Linux.

benmoran56:
> I'm confused about the types of _DefaultLoader.

`resource.path` is just a list of str (directory paths).

gran4:
Thanks

gran4:
> I'm not positive, but the `from __future__ import annotations` might be the culprit here. Strange that it only happens on OSX and Linux.

I think I broke it more becuase I added `from __future__ import annotations` to more files.

benmoran56:
> Thanks
> 
> > I'm not positive, but the `from __future__ import annotations` might be the culprit here. Strange that it only happens on OSX and Linux.
> 
> What should I do?

Why not make a new branch, and then start a pull request with only one file changed?  The automated tests will then run on that, and we can verify if the future import is indeed the cause.

gran4:
If that's the problem, should I just remove that?

gran4:
In the arcade python library, they use TYPE_CHECKING bool from typing. I think I'll use that.

benmoran56:
I'm afraid that after much consideration, it's not going to be possible to merge such as large pull request.
There are many mistakes that I can find on a quick glance, and it's too time consuming and error prone to review them all. 

Please break these changes into smaller individual pull requests. 
It's better if you stick to a single module, or even a single class (depending on the number of changes).

PR #810, opened 2023-04-18
Add context manager wrapping GDI acquisition & release

pushfoo:
tl;dr trial run for context managers to help cleanup font loader code as discussed on discord

This PR adds a context manager which does the following to altered code:
1. Increase durability due to no longer needing to manually close the resource
2. Increase readability by adding structure to code

This is a trial step for cleaning up font loading rather than optimizing pyglet performance. Performance-critical code is not intended to use context managers unless benchmarking deems their overhead acceptable.


benmoran56 (approved):

shenjackyuanjie:
image

looks good
go for merge!

benmoran56 (approved):

PR #814, opened 2023-04-22
TEST

gran4:
Testing `from __future__ import annotations` as @benmoran56  said in https://github.com/pyglet/pyglet/pull/808


gran4:
Wow, that was the problem.

Issue #787, opened 2023-03-06
cannot blit_into image 


emrum:
**Describe the bug**
for some reason , image . blit_into( source , ...)  calls an abstract method and throws exception. 

**System Information:**
this is for pyglet version 2.0.5 

**How To Reproduce**
here is the code:

import pyglet
from pyglet.gl import *

window = pyglet.window.Window()

tex_big = pyglet.image.load('texture_RGBA.png').get_texture()

tex_small = pyglet.image.load('white_pixel.png').get_texture()

@window.event
def on_draw():

    glClearColor(255,255,255,255)
    # Clear the window
    window.clear()

    tex_big .blit_into (source=tex_small, x=0, y=0, z=0)

    tex_big .blit (0,0,0)
    
#
pyglet.app.run()

caffeinepills:
The source has to be `ImageData`, not a `Texture`.

emrum:
> The source has to be `ImageData`, not a `Texture`.

ah , this.  then why it is not in the documentation ?  I had tried in vain, with image as well.

emrum:
true,  it works with get_imge_data() ,  seems this time, the bug is more in the documentation and error-msg, wich do not tell anything about it.

PR #818, opened 2023-04-30
Add Type Hints to image/__init__.py

gran4:
I wanted to fix https://github.com/pyglet/pyglet/issues/787 so I type hinted it.

### DIFF
--- a/pyglet/image/__init__.py
+++ b/pyglet/image/__init__.py
@@ -226,8 +226,7 @@ def _color_as_bytes(color: Tuple[int, int, int, int]) -> bytes:
     :Parameter:
         `color`: Tuple[int, int, int, int]
             Color in RGBA standard
-    
-    :return: color as a bytes object
+
     """
     if len(color) != 4:
         raise TypeError("color is expected to have 4 components")
@@ -573,7 +572,7 @@ class ImageData(AbstractImage):
     """An image represented as a string of unsigned bytes.
 
     :Parameters:
-        `data` : str
+        `data` : bytes
             Pixel data, encoded according to `format` and `pitch`.
         `format` : str
             The format string to use when reading or writing `data`.
@@ -590,7 +589,7 @@ class ImageData(AbstractImage):
     _current_texture = None
     _current_mipmap_texture = None
 
-    def __init__(self, width: int, height: int, fmt: str, data: Sequence, pitch: Optional[int]=None) -> None:
+    def __init__(self, width: int, height: int, fmt: str, data: bytes, pitch: Optional[int]=None) -> None:
         """Initialise image data.
 
         :Parameters:
@@ -600,8 +599,8 @@ class ImageData(AbstractImage):
                 Height of image data
             `fmt` : str
                 A valid format string, such as 'RGB', 'RGBA', 'ARGB', etc.
-            `data` : sequence
-                String or array/list of bytes giving the decoded data.
+            `data` : bytes
+                array/list of bytes giving the decoded data.
             `pitch` : int or None
                 If specified, the number of bytes per row.  Negative values
                 indicate a top-to-bottom arrangement.  Defaults to
@@ -665,7 +664,7 @@ class ImageData(AbstractImage):
             return self._current_data
         return self._convert(fmt, pitch)
 
-    def set_data(self, fmt: str, pitch: int, data: Union[str, Sequence[bytes]]) -> None:
+    def set_data(self, fmt: str, pitch: int, data: bytes) -> None:
         """Set the byte data of the image.
 
         :Parameters:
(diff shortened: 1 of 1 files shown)


gran4:
NOT done

pyglet/image/__init__.py
```diff
@@ -1431,10 +1524,13 @@ def __repr__(self):
 
 
 class TextureRegion(Texture):
-    """A rectangular region of a texture, presented as if it were a separate texture.
+    """
```
benmoran56, line 1527: Please leave this one as it was, with the leading """ on the same line.

pyglet/image/__init__.py
```diff
@@ -1786,21 +1893,21 @@ def __init__(self, image, rows, columns, item_width=None, item_height=None, row_
         self.row_padding = row_padding
         self.column_padding = column_padding
 
-    def get_texture(self, rectangle=False):
+    def get_texture(self, rectangle=False) -> Texture:
         return self.image.get_texture(rectangle)
 
-    def get_image_data(self):
+    def get_image_data(self) -> TextureRegion:
```
benmoran56, line 1899: This return type is wrong. Should be ImageData.

pyglet/image/__init__.py
```diff
@@ -1811,7 +1918,7 @@ def _update_items(self):
                     x += self.item_width + self.column_padding
                 y += self.item_height + self.row_padding
 
-    def __getitem__(self, index):
+    def __getitem__(self, index: int) -> None:
```
benmoran56, line 1921: The return type here is certainly not None. It will be either a single `AbstractImage` or a list of `AbstractImage`.

benmoran56 (changes requested):
Most of this looks good, but there are a large amount of inaccuracies. I have made comments for those that I have found, but there could be more.

pyglet/image/__init__.py
```diff
@@ -362,7 +386,10 @@ def get_region(self, x, y, width, height):
         """
         raise ImageException('Cannot get region for %r' % self)
 
-    def save(self, filename=None, file=None, encoder=None):
+    def save(self, filename: Optional[str]=None,
```
benmoran56, line 389: Two things: 
`filename` should not default to `None`. This is a bug. Filename is required. 
This function does not return anything, so return type should be `-> None`.

pyglet/image/__init__.py
```diff
@@ -376,6 +403,8 @@ def save(self, filename=None, file=None, encoder=None):
                 are tried.  If all fail, the exception from the first one
                 attempted is raised.
 
+        :rtype: None or bytes
```
benmoran56, line 406: :rtype: and :return: are not needed, because they are implied by the typing definitions.

pyglet/image/__init__.py
```diff
@@ -413,6 +442,16 @@ def blit_into(self, source, x, y, z):
         Note that if `source` is larger than this image (or the positioning
         would cause the copy to go out of bounds) then you must pass a
         region of `source` to this method, typically using get_region().
+
+        :Parameters:
```
benmoran56, line 446: I would remove this section as it's unneeded.
It's also incorrect, as (x, y, z) is the destination, not the anchor point.

pyglet/image/__init__.py
```diff
@@ -1003,20 +1053,20 @@ def get_data(self, fmt=None, pitch=None):
         pitch = pitch or self._current_pitch
         return super().get_data(fmt, pitch)
 
-    def set_data(self, fmt, pitch, data):
+    def set_data(self, fmt: str, pitch: int, data: Sequence[Union[bytes, str]]):
```
benmoran56, line 1056: `data` should just be `bytes` here, not bytes/str.

pyglet/image/__init__.py
```diff
@@ -1203,17 +1259,19 @@ class Texture(AbstractImage):
 
     """
 
-    region_class = None  # Set to TextureRegion after it's defined
```
benmoran56, line 1206: Lets leave these local variables as they are. I don't see much benefit from typing information here, as they're internal defaults. It just adds more clutter, IMO.

pyglet/image/__init__.py
```diff
@@ -1359,14 +1417,35 @@ def blit(self, x, y, z=0, width=None, height=None):
...
+
+        `source` will be copied into this image such that its anchor point
+        is aligned with the `x` and `y` parameters.  If this image is a 3D
+        texture, the `z` coordinate gives the image slice to copy into.
+
+        Note that if `source` is larger than this image (or the positioning
+        would cause the copy to go out of bounds) then you must pass a
+        region of `source` to this method, typically using get_region().
+
+        :Parameters:
```
benmoran56, line 1432: This can be removed. It's also incorrect - (x, y, z) do not mean the anchor points.

benmoran56 (changes requested):

gran4:
Ok. Are you the only one reviewing pyglet PRs.

PR #819, opened 2023-05-02
fix broken packs/unpacks in ttf/png

kilimnik:
I notices, that the loading of tiled maps was not working, when the image was using a color map. The exception was an unpack error.
After some digging, I noticed that the format had the wrong syntax. This was introduced in [this](https://github.com/pyglet/pyglet/commit/561069c110a14e9b388d69d2cf23274252e86933) commit.

I looked through the commit and fixed the same issue in the ttf parser.


caffeinepills (approved):
Looks good, thanks for catching that.

Issue #821, opened 2023-05-03
Auto repotition "widget mask"


shenjackyuanjie:
you can see, in a resizeable window, when you resize it, the widget/shape/line will be effected
but their position won't be tranfer to the new position,
like this

normally, we will write a bunch of code in on_resize()
but what if we just write a cover to the widget/shape
with it, you can just call `cover.on_resize` to resize all the shape/widget attach on it
and you only need to give a function takes in `height` `width` and give out `x` and `y` (or even `rotate`)

like this
import random
```python
class WidgetCover(EventDispatcher):
  def __init__(self, **args):
    self.widget_dict: Dict[Union[str, int], Tuple[Union[Shape, WidgetBase], Callable[[wdiget, x, y], None]]] = {}
    self.random_size = [-1_0000, 1_0000]

  def add_widget(self, widget: Union[Shape, Widget], func: Callable[[widget, x, y], None], index: Optional[Union[str, int]] = None) -> Union[str, int]:
    if index is None:
      index = random.randomint(*self.random_size) # or some other size
    while self.widget_dict.get(index) is not None:
      index = randome.randomint(*self.random_size)
    self.widget_dict[index] = (widget, func)
    return index
```

shenjackyuanjie:
as the mention in https://github.com/pyglet/pyglet/pull/743#issuecomment-1531772525

caffeinepills:
There is now an on_resize event for widgets to re-order. Example provided in pyglet 3.0 now.

shenjackyuanjie:
\o/

PR #822, opened 2023-05-03
[WIP] Auto Reposition Frame for #821

shenjackyuanjie:
the Pull Request for #821


shenjackyuanjie:
hello?
anyone here?

benmoran56:
Hi @shenjackyuanjie,  sorry for the delay. 

How about implementing this as a subclass of Frame?  That way, the spacial hash can be used.

shenjackyuanjie:
> How about implementing this as a subclass of Frame? That way, the spacial hash can be used.

Hmmmmm, I add the `hash` here just for indexing the widget/shape so that you can add/remove widget/shape with a `hash`, witch spacial hash are you talking about? if it is the hash used by `Frame`. I do not think it will have a good performance ( after all it will remove and add all the widget/shape when resizing)

I do think make this a subclass of `Frame` is a good idea, but it will require many modify in `Frame` class. sence the `Frame` have some unuse function for the reposition frame

(BTW, @jason-bowen-zheng and I is think about make this function just like the tkinker's `grid` function, but it will be a hard time to do it

shenjackyuanjie:
So I change it to draft with WIP
let's see if this PR can become better

shenjackyuanjie:
I still think this PR is just an WIP one
just that I'm not sure how to implment it

PR #833, opened 2023-05-10
Testing whether a point is inside a shape

jason-bowen-zheng:
Some functions were added to several shapes:

1. Circle
2. Ellipse
3. Sector
4. Line
5. Rectangle
6. BoderedRectangle
7. Triangle
8. Star
9. Polygon

The following code
```python
from pyglet import shapes

rect = Rectangle(...)
# Do something else
if (100, 100) in rect:
    rect.color = (255, 0, 0)
```
will turn `rect`'s color to red when the point `(100, 100)` is in the shape.

It's also available for rotated and anchored shapes.

### DIFF
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -8,6 +8,8 @@ Convenience methods are provided for positioning, changing color
 and opacity, and rotation (where applicable). To create more
 complex shapes than what is provided here, the lower level
 graphics API is more appropriate.
+You can also use the ``in`` operator to check whether a point is
+inside a shape.
 See the :ref:`guide_graphics` for more details.
 
 A simple example of drawing shapes::
@@ -132,6 +134,7 @@ def _sat(vertices, point):
         if point_proj < min(projections) or point_proj > max(projections):
             return False
     return True
+
 
 class _ShapeGroup(Group):
     """Shared Shape rendering Group.
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -960,6 +960,7 @@ class Sector(ShapeBase):
         if angle < 0: angle += 2 * math.pi
         if self._start_angle < angle < self._start_angle + self._angle:
             return math.dist((self._x - self._anchor_x, self._y - self._anchor_y), point) < self._radius
+        return False
 
     def _create_vertex_list(self):
         self._vertex_list = self._group.program.vertex_list(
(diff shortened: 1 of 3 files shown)


pyglet/shapes.py
```diff
@@ -719,6 +731,10 @@ def __init__(self, x, y, radius, segments=None, color=(255, 255, 255, 255),
         self._create_vertex_list()
         self._update_vertices()
 
+    def __contains__(self, point):
+        assert len(point) == 2
+        return math.dist((self._x - self._anchor_x, self._y - self._anchor_y), point) < self._radius
```
gran4, line 736: 1: You have __contains__ twice
2: You don't actually check if the point is in any of the shapes

jason-bowen-zheng: Can you explain it in detail?

benmoran56:
@jason-bowen-zheng 
I can see you're still working on this. Please let me know once you feel ready.

pyglet/shapes.py
gran4: I am dumb never mind

jason-bowen-zheng:
@benmoran56, I think it's nearly finished. But there are still some problems:

1. only regular polygons were tested
2. the SAT implementation may have problem: [triangle's](https://github.com/pyglet/pyglet/blob/41f4f40a375576fa82240da50a64549a9f431c33/pyglet/shapes.py#L1554) and [polygon's](https://github.com/pyglet/pyglet/blob/41f4f40a375576fa82240da50a64549a9f431c33/pyglet/shapes.py#L1791) codes that gets the correct result is different.

benmoran56:
I haven't checked the *sat* function yet on my own PC, but could it be an issue with convex v.s. concave polygons?

Perhaps a different algorithm might be better, that can handle both types of polygons. This one might be good to try: https://www.algorithms-and-technologies.com/point_in_polygon/python

jason-bowen-zheng:
First, SAT only supports convex polygons(since pyglet only support convex polygons, too).

Then, `point_in_polygon` still has some problems.

So my opinion is that we can merge this PR first. Then make pyglet support concave polygons and `point_in_polygon`.

benmoran56:
Perhaps it's enough to treat the Star as a circle :thinking:

jason-bowen-zheng:
This is indeed a compromise. The radius of a circle can be considered as `(inner_radius+outter_radius)/2`.

benmoran56:
Great! 
That's all of the shapes then, isn't it?

jason-bowen-zheng:
Yes, except for those drawn using `GL_LINES`.

benmoran56 (approved):

jason-bowen-zheng:
I think now this PR is perfect!

Besides, I made the docs of shapes short and brief.

benmoran56 (approved):
Looks good!

PR #835, opened 2023-05-14
Added support for FFmpeg 6.0.

caffeinepills:
Added support for FFmpeg 6.0. Tested with versions 4.4, 5.1, and 6.0; both audio and video on Windows.

Needs testing on Mac and Linux.

Also:
Added a warning if the loaded FFmpeg versions are not the expected versions we have tested with.

FFmpeg debug message now shows some extra information when FFmpeg is not available.

Added a missing print for OSError in the load_library to prevent hidden errors when loaded.

Removed mp4/mov extension from WMF. (Experimental in the first place, and had no acceleration. Just use FFmpeg for smoothest experience)

### DIFF
--- a/pyglet/lib.py
+++ b/pyglet/lib.py
@@ -125,8 +125,9 @@ class LibraryLoader:
                         if _debug_trace:
                             lib = _TraceLibrary(lib)
                         return lib
-                    except OSError:
-                        pass
+                    except OSError as e:
+                        if _debug_lib:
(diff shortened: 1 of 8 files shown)


benmoran56 (approved):

benmoran56:
Manually tested it on my Arch machine, with ffmpeg 6.0-6. 
Works well with the files I tried.

PR #838, opened 2023-05-16
fixed typo in pyglet/media/drivers to the base.py signature

kitscribe:
`PulseAudioPlayer` class of the `pyglet/media/drivers/pulse/adaption.py` file has an incorrect signature within the `_process_events` method.

Fix typo according to the `MediaEvent` class of the `pyglet/media/drivers/base.py` file

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,3 +1,17 @@
+pyglet 2.0.7
+
+Changes and Improvements
+------------------------
+- gui: Add alpha value to the Caret's default color.
+- docs: Some minor rewording of sections, and type fixes.
+
+Bugfixes
+--------
+- Fix broken struct format strings in png and ttf decoders.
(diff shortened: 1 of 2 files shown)


caffeinepills:
Good catch! This must have been changed recently with some of the media event adjustments.

It also looks like other drivers also call `_sync_dispatch_to_player` in their `adaptation.py` files. Would you mind doing a replace on those files as well? Otherwise I can close this and open my own. Thanks.

kitscribe:
> Good catch! This must have been changed recently with some of the media event adjustments.
> 
> It also looks like other drivers also call `_sync_dispatch_to_player` in their `adaptation.py` files. Would you mind doing a replace on those files as well? Otherwise I can close this and open my own. Thanks.

Sure, I will amend it, thank's for your notice!

caffeinepills (approved):
Looks good to me.

Issue #840, opened 2023-05-18
Shapes rotation API inconsistency


jej:
Hi,
I am new to pyglet. I have a question about shape rotations.

Shapes do not support rotation equally. For example, `Triangle` does not support it at all. Some shapes have a property/setter (`self.rotation`). The class `Star` has an argument in its constructor (in addition to property/setter).

I tried to add the rotation to triangle with the code found in other shapes:
```
    @property
    def rotation(self):
        return self._rotation

    @rotation.setter
    def rotation(self, rotation):
        self._rotation = rotation
        self._vertex_list.rotation[:] = (rotation,) * self._num_verts
```
It works partially, seems to rotate around the wrong anchor position.

I wonder why this code is not on `ShapeBase` to provide rotation to all shapes.

Thanks anyway for this nice library, reminding me the great epoc of Amiga demos :)

jej:
Well, bug I guess. See my PR in #841

PR #841, opened 2023-05-18
Bug in triangles and rotation feature

jej:
I had problems with triangle positioning (see #840). I guess it's a bug. Here is how I solved.

By the way, this PR adds rotation to triangle.

Test case:
```
import math
import pyglet
from pyglet import shapes


class ShapesDemo(pyglet.window.Window):

    def __init__(self, width, height):
        super().__init__(width, height, "Shapes")
        self.time = 0

        self.triangle = shapes.Triangle(
            0, -50, 50, 0, 0, 50, color=(55, 255, 255, 175))
        self.triangle.anchor_x = 25
        self.triangle.anchor_y = 0
        self.triangle.position = (100, 100)

    def on_draw(self):
        """Clear the screen and draw shapes"""
        self.clear()
        self.triangle.draw()

    def update(self, delta_time):
        """Animate the shapes"""
        self.time += delta_time

if __name__ == "__main__":
    demo = ShapesDemo(200, 200)
    pyglet.clock.schedule_interval(demo.update, 1/30)
    pyglet.app.run()
```

Without the patch it displays:
screenshot_20230518-131040

The triangle should be centred. Adding a rotation shows the centre is not well positioned.

With the patch:
screenshot_20230518-131123

To see the rotation in action around the anchor point, try:
```
    def update(self, delta_time):
        """Animate the shapes"""
        self.time += delta_time
        self.triangle.rotation = self.time * 30
```

Hope it helps,
J.

### DIFF
--- a/examples/graphics/shapes.py
+++ b/examples/graphics/shapes.py
@@ -26,6 +26,8 @@ class ShapesDemo(pyglet.window.Window):
         self.line = shapes.Line(0, 0, 0, 480, width=4, color=(200, 20, 20), batch=self.batch)
 
         self.triangle = shapes.Triangle(10, 10, 190, 10, 100, 150, color=(55, 255, 255, 175), batch=self.batch)
+        self.triangle.anchor_position = 55, 80
+        self.triangle.position = 120, 100
 
         septagon_step = math.pi * 2 / 7
         self.fading_septagon = shapes.Polygon(
@@ -57,6 +59,8 @@ class ShapesDemo(pyglet.window.Window):
 
         self.line.x = 360 + math.sin(self.time * 0.81) * 360
         self.line.x2 = 360 + math.sin(self.time * 1.34) * 360
+
+        self.triangle.rotation = self.time * 25
 
         self.arc.rotation = self.time * 30
 
--- a/examples/graphics/shapes.py
+++ b/examples/graphics/shapes.py
@@ -25,9 +25,11 @@ class ShapesDemo(pyglet.window.Window):
 
         self.line = shapes.Line(0, 0, 0, 480, width=4, color=(200, 20, 20), batch=self.batch)
 
-        self.triangle = shapes.Triangle(10, 10, 190, 10, 100, 150, color=(55, 255, 255, 175), batch=self.batch)
-        self.triangle.anchor_position = 55, 80
-        self.triangle.position = 120, 100
+        size = 200
+        self.triangle = shapes.Triangle(-size/2, 0, size/2, 0, 0, math.sqrt(3)/2*size, color=(55, 255, 255, 175), batch=self.batch)
+        self.triangle.anchor_position = 0, math.sqrt(3)/6*size
+        self.triangle.position = 100, 100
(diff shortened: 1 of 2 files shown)


jej:
Could be related to #664

benmoran56:
I had a look at this yesterday, and it seems that just adding the `rotation` getter and setter were enough to make it function correctly.

jej:
Sorry, I persist...

Look at this branch (https://github.com/jej/pyglet/tree/triangle-rotation) and run shapes.py example.

It draws a math generated centred equilateral triangle with a red dot as its center. Result:

screenshot_20230524-191203

Then checkout branch: https://github.com/jej/pyglet/tree/triangle-anchor-bug 

It reverts the code that you committed [here](https://github.com/pyglet/pyglet/commit/f5bbeac689e56cc9db66a696093ee9e4985980f0#diff-d58a3d97408215fcb68d071cf3cbbf68dcd3f1af711498f44d01cc1a7e7032c5L1364-R1369) (lines 1364-1375): 

Result as expected:
screenshot_20230524-193344

I may be wrong understanding how anchors & position work, but I am confused anyway :)

benmoran56:
Removing all of the other shapes, the first example looks as expected. 
It's setting the y axis anchor down ~57 pixels:
image

The default anchor point is the first vertex point, or the bottom left point in this triangle. Most shapes will default to the first vertex as the default anchor, which matches the behavior of Images/sprites. The only shapes that don't adhere to this are circular ones where it makes sense to use the center point.

jej:
The 3 triangles vertex coordinates are:
(-100, 0)
(0, 173)
(100, 0)

Then the anchor is set at the middle of the triangle:
(0,57)

So the rotation should be around this point, not (-100, 0).

I don't understand why you removed the coordinates of the first vertex in you august 2022 commit:
screenshot_20230525-082429

benmoran56:
> The 3 triangles vertex coordinates are: (-100, 0) (0, 173) (100, 0)

Yes, but then position is set to 100, 100:

        self.triangle = shapes.Triangle(-size/2, 0, size/2, 0, 0, math.sqrt(3)/2*size, color=(55, 255, 255, 175), batch=self.batch)
        self.triangle.anchor_position = 0, math.sqrt(3)/6*size
        self.triangle.position = 100, 100

The initial coordinates are:
(-100.0, 0) (100.0, 0) (0, 173.20508075688772)

But after changing triangle.position, they are changed to:
(100, 100) (200.0, 100) (100, 273.2050807568877)


> Then the anchor is set at the middle of the triangle: (0,57)

That is not the middle of the triangle - anchors are in relation to the first vertex point. Or in other words:
```py
triangle.x1 - 0
triangle.y1 - 57
```

This is what is shown in my screen shot, and it is what's expected. If you want to set the anchor to the center of the triangle, you need to set both X and Y in relation to the Triangle's first vertex.

jej:
How do you manage a rotation around the middle? Could you rewrite the example according to the real usage?

To me the anchor was just a point relative to all others, so that we can use it like a handle...

PR #848, opened 2023-05-20
Add duration for schedule interval

jej:
This PR adds a duration for a schedule. It was one of my needs.

```
The function can be scheduled for a given period of time by adding
the keyword argument `_duration=` (in seconds). After this delay,
the function is unscheduled automagically.
```
Example:
```
import time
x = 0

def myfunc(dt):
    global x
    x += 1
    print(x)
    if x == 5:
        print(time.time() - start)

start = time.time()
pyglet.clock.schedule_interval(myfunc, 1, _duration=5)
```
Result:
```
1
2
3
4
5
5.000953674316406
```

It's seems reliable, but experimental and need to be tested before merging.

Hope it helps,
J.

### DIFF
--- a/pyglet/clock.py
+++ b/pyglet/clock.py
@@ -469,7 +469,7 @@ class Clock:
                 The function to call when the timer lapses.
             `interval` : float
                 The number of seconds to wait between each call.
-            `duration=` : float
+            `duration` : float
                 The number of seconds for which the function is scheduled.
 
         """
--- a/pyglet/clock.py
+++ b/pyglet/clock.py
@@ -444,9 +444,23 @@ class Clock:
         Specifying an interval of 0 prevents the function from being
         called again (see `schedule` to call a function as often as possible).
 
-        The function can be scheduled for a given period of time by adding
-        the keyword argument `_duration=` (in seconds). After this delay,
-        the function is unscheduled automagically.
+        The callback function prototype is the same as for `schedule`.
+
+        :Parameters:
+            `func` : callable
+                The function to call when the timer lapses.
+            `interval` : float
+                The number of seconds to wait between each call.
+
+        """
+        last_ts = self._get_nearest_ts()
+        next_ts = last_ts + interval
+        item = _ScheduledIntervalItem(func, interval, last_ts, next_ts, args, kwargs)
+        _heappush(self._schedule_interval_items, item)
+
+    def schedule_interval_for_duration(self, func, interval, duration, *args, **kwargs):
+        """Schedule a function to be called every `interval` seconds
+        (see `schedule_interval`) and unschedule it after `duration` seconds.
 
         The callback function prototype is the same as for `schedule`.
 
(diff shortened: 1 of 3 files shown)


benmoran56:
This seems functionally fine, but I have some concerns.
The additionally checking inside of the clock each tick isn't ideal. This code needs to be as an fast as possible, and even tiny things like this can add up at 120Hz.
Another minor point is that the resolution/accuracy of the duration would be limited by the schedule frequency.

Instead of this, how about just using another schedule? 

```py
clock.schedule_interval(myfunc, 1.0)
clock.schedule_once(clock.unschedule, 5.0, myfunc)
```

jej:
Oh sorry, I agree! Bad point to load more the scheduler at this stage... :|

Before that I unscheduled inside the scheduled function, by checking time at this level. Working well but not so DRY when many functions needs that kind of schedule. So I thought it was nice to have it in pyglet core without having to tweak the scheduled functions.

Trying your solution I get this:
```
Exception has occurred: TypeError
unschedule() takes 1 positional argument but 2 were given
  File ".../pyglet/clock.py", line 227, in call_scheduled_functions
    item.func(now - item.last_ts, *item.args, **item.kwargs)
  File ".../pyglet/app/base.py", line 212, in idle
    self.clock.call_scheduled_functions(dt)
  File ".../pyglet/app/base.py", line 151, in run
    timeout = self.idle()
              ^^^^^^^^^^^
  File ".../pyglet/app/__init__.py", line 76, in run
    event_loop.run(interval)
  File ".../application.py", line 57, in run
    pyglet.app.run()
  File ".../myapp.py", line 25, in <module>
    Application().run()
TypeError: unschedule() takes 1 positional argument but 2 were given
```

```
x = 0
def myfunc(dt):
    global x
    x += 1
    print(x)
    if x == 5:
        print(time.time() - start)

start = time.time()
pyglet.clock.schedule_interval(myfunc, 1)
pyglet.clock.schedule_once(pyglet.clock.unschedule, 5.0, myfunc)
```

jej:
Ok, I manage, don't worry...

jej:
Ok, I managed by the way you mention, quite better of course! Needed some tweak to accept the dt argument.

Now, what about the `_duration` args to achieve this? It's not possible the have a default nulled arguments, because of the kwargs. Another solution could be a new method like `schedule_interval_for_duration(myfunc, 1, 5)`. But it we cannot create a new method for each functionality...

jej:
And feel free to decline the request if the contribution is not useful for pyglet.

It's just a pleasure for me to make things with and look into it!

benmoran56:
I think unfortunately the extra checks involved make this a no-go for the moment. I'm also not really sure how people will feel about the kwargs hack either.

jej:
There is no more extra check after my last commit. Did you notice the change?

I agree the _duration is not the best way, because not a practice in pyglet.

But the feature itself is quite useful. Launching a regular task for a given period of time, without having to manage it manually...

jej:
Yep, according to the current pyglet design pattern, a new method would do the job.

Ok then, I will commit something later.

jej:
PS: the function is called `schedule_interval_/for/_duration` to avoid confusion with `interval_duration`

pyglet/clock.py
```diff
@@ -458,6 +458,28 @@ def schedule_interval(self, func, interval, *args, **kwargs):
...
+        (see `schedule_interval`) and unschedule it after `duration` seconds.
+
+        The callback function prototype is the same as for `schedule`.
+
+        :Parameters:
+            `func` : callable
+                The function to call when the timer lapses.
+            `interval` : float
+                The number of seconds to wait between each call.
+            `duration=` : float
```
benmoran56, line 472: Just a tiny typo here (=).

benmoran56 (changes requested):
Looks good to me. Just a tiny typo, but after that it should be good to merge.

pyglet/clock.py
jej: ooops sorry, a remain of the keyword argument documentation...

thx

benmoran56 (approved):

PR #851, opened 2023-05-21
Sprite height/width @setters

jej:
In the app I am coding, I need to work with pixel sizes, not with scale. Seems to work now like this:

```
sprite = pyglet.sprite.Sprite(img=image)
sprite.width = 1920
sprite.height = 1440
```
Currently width/height are read-only.

### DIFF
--- a/pyglet/sprite.py
+++ b/pyglet/sprite.py
@@ -655,23 +655,31 @@ class Sprite(event.EventDispatcher):
     def width(self):
         """Scaled width of the sprite.
 
-        Read-only.  Invariant under rotation.
+        Invariant under rotation.
 
         :type: int
         """
         w = self._texture.width * abs(self._scale_x) * abs(self._scale)
         return w if self._subpixel else int(w)
 
+    @width.setter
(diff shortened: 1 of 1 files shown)


gran4:
I don't think there is a case where _scale is negitive, is there?

pushfoo:
> I don't think there is a case where _scale is negitive, is there?

There is. Negative scale is a quick way of mirroring along the X and Y axes, or only one axis if you only alter scale for one dimension.

The downside is that it produces negative apparent widths if handled naively. Using `abs` ensures meaningful return values which can be used for layout and packing.

jej:
I just used the formula to get width from scale:

```
width = self._texture.width * abs(self._scale_x) * abs(self._scale)
```
to get the scale from width (or height). As said by @pushfoo, a negative scale/width (or height) mirror the sprite.

benmoran56 (approved):

benmoran56:
This looks like a reasonably useful change, and doesn't break anything. Thanks for the PR.

PR #853, opened 2023-05-26
add _get_right to label

gran4:
@benmoran56 can you review this quickly, I need this for a different PR in arcade.

Also, does the functions like `_get_right` need _? Since arcade uses things like `_get_left`, is arcade using it wrong?


benmoran56:
This is unnecessary code duplication.
It's just `_get_left + content_width`.

PR #856, opened 2023-05-28
Add type hints to resources.py

gran4:
Why is there a circlular error?

EDIT: nvmd, don't know what happened

### DIFF
--- a/pyglet/resource.py
+++ b/pyglet/resource.py
@@ -50,6 +50,8 @@ The default path is ``['.']``.  If you modify the path, you must call
 .. versionadded:: 1.1
 """
 
+from __future__ import annotations
+
 import os
 import sys
 import zipfile
--- a/pyglet/resource.py
+++ b/pyglet/resource.py
@@ -351,7 +351,7 @@ class Loader:
             application script.
 
     """
-    def __init__(self, path: [str]=None, script_home: str=None) -> None:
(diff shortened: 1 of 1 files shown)


pyglet/resource.py
```diff
@@ -50,20 +50,32 @@
...
 import sys
 import zipfile
 import weakref
 
 from io import BytesIO
+from typing import Union, Optional, List, IO, TYPE_CHECKING
 
 import pyglet
 
+if TYPE_CHECKING:
```
benmoran56, line 65: This breaks normal operation.

pyglet/resource.py
```diff
@@ -240,15 +255,26 @@ def __init__(self, filepath):
...
+        """
         return open(os.path.join(self.path, filename), mode)
 
 
 class ZIPLocation(Location):
     """Location within a ZIP file.
     """
 
-    def __init__(self, zip, dir):
+    def __init__(self, zip: zipfile.ZipFile, dir: str) -> None:
```
benmoran56, line 277: The zip argument type is not correct.

pyglet/resource.py
```diff
@@ -240,15 +255,26 @@ def __init__(self, filepath):
         """
         self.path = filepath
 
-    def open(self, filename, mode='rb'):
+    def open(self, filename: str, mode: str='rb') -> IO:
```
benmoran56, line 258: The parameters are not correct.

PR #857, opened 2023-05-28
Add type hints to shapes

gran4:


pyglet/shapes.py
```diff
@@ -42,18 +42,28 @@ def on_draw():
...
 from pyglet.gl import glBlendFunc, glEnable, glDisable
 from pyglet.graphics import Batch, Group
+from pyglet.graphics.shader import ShaderProgram, ComputeShaderProgram
+
+from typing import (
+    Union,
+    Optional,
+    Tuple,
+    List
+)
```
shenjackyuanjie, line 63: how about just `from typing import Union, Optional, Tuple, List`

pyglet/shapes.py
```diff
@@ -116,7 +126,7 @@ class _ShapeGroup(Group):
     sharing the same parent group and blend parameters.
     """
 
-    def __init__(self, blend_src, blend_dest, program, parent=None):
+    def __init__(self, blend_src: int, blend_dest: int, program: ShaderProgram, parent: Optional[Group]=None) -> None:
```
shenjackyuanjie, line 129: why not add a gap between `Optional[Group]` with `=None`
like `parent: Optional[Group] = None`

pyglet/shapes.py
```diff
@@ -431,16 +449,18 @@ def batch(self, batch):
 class Arc(ShapeBase):
     _draw_mode = GL_LINES
 
-    def __init__(self, x, y, radius, segments=None, angle=math.tau, start_angle=0,
-                 closed=False, color=(255, 255, 255, 255), batch=None, group=None):
+    def __init__(self, x: number, y: number, radius: float, segments: Optional[int]=None, 
+                 angle: float=math.tau, start_angle: float=0, closed: bool=False, 
+                 color: Tuple[int, int, int, int]=(255, 255, 255, 255), 
+                 batch: Optional[Batch]=None, group: Optional[Group]=None) -> None:
```
shenjackyuanjie, line 455: add gap between type hint and default value

pyglet/shapes.py
```diff
@@ -1395,7 +1424,7 @@ def border_color(self, values):
         self._update_color()
 
     @property
-    def color(self):
+    def color(self) -> Tuple[int, int, int, int]:
```
shenjackyuanjie, line 1427: how about add a type named `color rgba`

shenjackyuanjie (commented):
add some gap?

pyglet/shapes.py
gran4: As imports get longer, it is better to put them on separate line. I do not actually know how many are needed to start putting them on separate lines though.

pyglet/shapes.py
gran4: Just makes the line longer and does not add anything

pyglet/shapes.py
gran4: Maybe

gran4:
In the ellipse class, segments is not in the docstrings. Is this a bug @benmoran56 or is this something that just needs to be updated?

pyglet/shapes.py
gran4: I think that will be a waste of time. It will make the lines longer for no reason.

PR #859, opened 2023-05-30
Improvements to win32 font docstrings, formatting, and annotations 

pushfoo:
### Changes

* Adds annotations to win32 font modules
* Adds docstrings
* Improves whitespace

### How to test

Run unit tests to make sure it still works

### Follow-up work

I'll be working my way through the other font code as time allows.

PR #860, opened 2023-05-30
Handle null return edge case in Win32Font.add_font_data

pushfoo:
Docstrings and annotations will be handled in #859

### Changes

Raise a WinError when a null return edge case [described in MSDN's AddFontMemResourceEx page](https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-addfontmemresourceex#return-value) is detected:
> If the function succeeds, the return value specifies the handle to the font added. This handle uniquely identifies the fonts that were installed on the system. If the function fails, the return value is zero. No extended error information is available.

### DIFF
--- a/pyglet/font/win32.py
+++ b/pyglet/font/win32.py
@@ -236,7 +236,12 @@ class Win32Font(base.Font):
     @classmethod
     def add_font_data(cls, data):
         numfonts = c_uint32()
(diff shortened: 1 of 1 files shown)


caffeinepills (approved):
This seems fine. Shouldn't occur with normal usage, but might help prevent some hidden errors from improper usage.

Issue #858, opened 2023-05-29
Erratic error: pyglet.gl.lib.GLException: (0x1281): Invalid value. A numeric argument is out of range.


jej:
I am struggling with an erratic error. I get it when playing a video. I tried different format, mp4, m4v but seems not related to the video itself. The video plays well, following the doc example. But after some restart of the app, it fails. If I reboot the computer or unplug/replug the external monitor, the error generally goes away. But it appears again later, even if I modify other parts of the app not related to the player. I got some opengl errors during the early dev of my app but it was always my fault. Here I don't have any idea why this error appears randomly.

The trace:
```
Traceback (most recent call last):
  File "/home/user/src/app/./myapp.py", line 262, in <module>
    Application().run()
  File "/home/user/src/app/./myapp.py", line 115, in run
    pyglet.app.run()
  File "/home/user/src/venv/lib/python3.10/site-packages/pyglet/app/__init__.py", line 76, in run
    event_loop.run(interval)
  File "/home/user/src/venv/lib/python3.10/site-packages/pyglet/app/base.py", line 151, in run
    timeout = self.idle()
  File "/home/user/src/venv/lib/python3.10/site-packages/pyglet/app/base.py", line 212, in idle
    self.clock.call_scheduled_functions(dt)
  File "/home/user/src/venv/lib/python3.10/site-packages/pyglet/clock.py", line 227, in call_scheduled_functions
    item.func(now - item.last_ts, *item.args, **item.kwargs)
  File "/home/user/src/venv/lib/python3.10/site-packages/pyglet/media/player.py", line 460, in update_texture
    self._texture.blit_into(image, 0, 0, 0)
  File "/home/user/src/venv/lib/python3.10/site-packages/pyglet/image/__init__.py", line 1469, in blit_into
    self.owner.blit_into(source, x + self.x, y + self.y, z + self.z)
  File "/home/user/src/venv/lib/python3.10/site-packages/pyglet/image/__init__.py", line 1364, in blit_into
    source.blit_to_texture(self.target, self.level, x, y, z)
  File "/home/user/src/venv/lib/python3.10/site-packages/pyglet/image/__init__.py", line 823, in blit_to_texture
    glTexSubImage2D(target, level,
  File "/home/user/src/venv/lib/python3.10/site-packages/pyglet/gl/lib.py", line 79, in errcheck
    raise GLException(f'(0x{error}): {msg}')
[...]

PR #861, opened 2023-06-02
Fix video player context

jej:
This is the result of a 30-hours bug hunt... This PR follows on from my bug report #858. The context of the bug is the use of the video player in an application managing several windows with interactions between them.

My investigations into this report were a bit hazy, as the bug encountered was slippery as a fish. Sometimes it appeared rarely, sometimes systematically, with no real connection to my debugging attempts. Outside vscode-ium, the bug appeared but rarely. Strangely, by overloading my code with log points, it slowed down and the bug became more systematic. Investigating the player code, I had a hunch that it was some kind of race hazard. I barded the player source with debugger logpoints and I noticed that at times the opengl context changed during the player workflow, and no longer corresponded to that of the window that supported the player. This was the cause of the opengl exception. Not some bad `switch_to` on my side...

So I created this python `@contextmanager` in order to execute graphics operations in the context of the player's creator. All my problems are gone in 1 second... Now it's up to whoever created this Player to tell me if the way of solving the problem is the right one. I can make an explanatory video to show the problem, if necessary.

I am happy because after several hours of bug hunt, I was really thinking about moving away from pyglet, because I had no chance on this problem. My code didn't use much specificity of pyglet so it was doable to move to another graphic backend. Problem: there no much alternatives in python world to manage graphics, sound and video together... so much the better!

Hope it helps.

### DIFF
--- a/pyglet/media/player.py
+++ b/pyglet/media/player.py
@@ -2,6 +2,7 @@
 
 import time
 from collections import deque
+from contextlib import contextmanager
 
 import pyglet
 from pyglet.gl import GL_TEXTURE_2D
@@ -105,6 +106,7 @@ class Player(pyglet.event.EventDispatcher):
 
     def __init__(self):
         """Initialize the Player with a MasterClock."""
+        self._context = pyglet.gl.current_context
         self._source = None
         self._playlists = deque()
         self._audio_player = None
@@ -125,6 +127,16 @@ class Player(pyglet.event.EventDispatcher):
     def __del__(self):
         """Release the Player resources."""
         self.delete()
+
+    @contextmanager
+    def context(self):
+        """ This context manager preserves the creator's context."""
+        previous_context = pyglet.gl.current_context
+        self._context.set_current()
+        try:
+            yield
+        finally:
+            previous_context.set_current()
 
     def queue(self, source):
         """
@@ -359,8 +371,9 @@ class Player(pyglet.event.EventDispatcher):
 
     def _create_texture(self):
         video_format = self.source.video_format
-        self._texture = pyglet.image.Texture.create(video_format.width, video_format.height, GL_TEXTURE_2D)
-        self._texture = self._texture.get_transform(flip_y=True)
+        with self.context():
+            self._texture = pyglet.image.Texture.create(video_format.width, video_format.height, GL_TEXTURE_2D)
+            self._texture = self._texture.get_transform(flip_y=True)
         # After flipping the texture along the y axis, the anchor_y is set
         # to the top of the image. We want to keep it at the bottom.
         self._texture.anchor_y = 0
@@ -457,7 +470,8 @@ class Player(pyglet.event.EventDispatcher):
         if image is not None:
             if self._texture is None:
                 self._create_texture()
-            self._texture.blit_into(image, 0, 0, 0)
+            with self.context():
+                self._texture.blit_into(image, 0, 0, 0)
         elif bl.logger is not None:
             bl.logger.log("p.P.ut.1.8")
 


pushfoo:
I don't have enough familiarity with the video player to say whether this implementation is appropriate, but it does raise an interesting question: does it make sense to put this on a mix-in class? If it still seems like a good idea after additional consideration, it could be explored as a separate issue or PR.

jej:
> mix-in class

What mix-in class?

I guess the _context manager_ is the Python pattern to address this kind of situation...

jej:
Ah ok sorry. My first attempt to solve the issue was just 2 functions (`switch_to()`/`switch_back()`), called before/after the graphic operations, with a temporary self.property to save the context. It's more elegant here but as you say it's better to check the side effects. For now I can continue my app safely :)

jej:
Same error.

Can you reproduce the bug according to my message here? https://github.com/pyglet/pyglet/issues/858#issuecomment-1575300473

benmoran56:
@jej

Sorry for leaving this hanging.  I don't think your solution is wrong, but I'm not sure if it's a good idea to add this complexity yet. We still need to do more research to determine the best way to handle multi-window (context) applications in pyglet. 

With that said, it does seem like a useful idea to add (Python) context methods (`__enter__` & `__exit__`) to `pyglet.gl.base.Context`. This gets us half-way there.  The `media_player.py` example itself is also not written in a way that can support multiple-windows (it's using various `item.draw()` calls, and also using `pyglet.graphics.draw` which is deprecated).

With that addition to pyglet's gl.Context, we can then add some safety barriers in `Player`, which should be safe and non-intrusive. Ie:
```py
self._context = pyglet.gl.current_context
...

with self._context:
  # texture updates
...
```

To summarize, I think a safe first steps forward would be to:
1. Allow pyglet.gl.Context to be used as a Python context manager.
2. Rewrite the media_player example so it's not using legacy code paths.

What do you think?

jej:
> What do you think?

I still don't know if someone could reproduce this problem. I provided the steps to reproduce (https://github.com/pyglet/pyglet/issues/858#issuecomment-1575300473) outside the context of my own application. I don't know if it's a pyglet problem or something related to my setup.

Concerning the way to solve it, it's up to you. When I discovered the source of the problem (after 1 week of investigation), I just solved it by saving the current context in a temp var before the blit, and restoring the context after. It worked the same way.

I don't want to investigate more to understand why the context changes and why my hack solves this. But sure it solves it at 100%, in my application/setup at least.

benmoran56:
I was able to reproduce it, and added a small addition to the Player class. This is as far as we want to go with this at the moment, but it should help in simple cases. I'll close this for now, but that you for your trouble shooting. It was helpful to pinpoint the source of the crash. 

The media_player.py example is still going to crash occasionally when pushing the buttons, because the use of `graphics.draw` is not not multi-context safe. This example needs to get re-written to remove those in any case.

jej:
My app is working now with _pyglet:master_.

Thanks too.

matanox:
Can anyone help me driving this home about why (or rather how) the opengl context would be switching inside of the media player? I'm using a `pyglet.media.Player()` for just audio playing, and though a slippery bug that doesn't always reproduce as in the original thread, it may be the same case as I would crash ocassionaly over the same error text mentioned from the original issue description leading here:

> pyglet.gl.lib.GLException: (0x1282): Invalid operation.

benmoran56:
@matanox the Player switches back to it's creation context when playing videos, and it creates a texture. That shouldn't be triggered when only playing audio. 

I would recommend upgrading to the latest point release regardless of issues, because there have been many bug fixes since 2.0.5.

matanox:
Thanks! Excellent!

matanox:
> > What mix-in class?
> 
> I meant adding something like this which would get used with multiple inheritance, but I'm not yet sure if it's a good idea. Multiple inheritance can have downsides.
> 
> ```python
> class ContextStackMixin:
> 
>     @contextmanager
>     def context(self):
>         """ This context manager preserves the creator's context."""
>         previous_context = pyglet.gl.current_context
>         self._context.set_current()
>         try:
>             yield
>         finally:
>             previous_context.set_current()
> ```

I think, from looking at the video player current code, that the original issue has been solved in it, albeit not via the context manager suggested here, which was also mentioned on https://github.com/pyglet/pyglet/issues/858#issuecomment-1575300473.

benmoran56:
@matanox what you are saying is possible, but would lead to a potentially high performance impact. OpenGL Context switching can be a costly operation on some platforms/drivers. It may be enough to enhance the VertexDomain classes to store/switch_to their contexts when drawing, as deletion is already handled by the context deletion pools. 

Window.switch_to would still be necessary to call manually to ensure that objects are created in the desired context. 

By the way, in case it can work for your application, you _can_ use the exact same Context for multiple windows. (Simply pass the first Window's context into the second Window on creation):
```py
>>> import pyglet
>>> 
>>> win1 = pyglet.window.Window()
>>> win2 = pyglet.window.Window(context=win1.context)
```
This will have side-effects on some (all?) platforms, such as vsync draw rate being divided among all of the Windows, and other potential anomalies, but it does seem to work.

Square789:
@matanox You said you are experiencing a

> crash over the same error text as the original issue

which i'm assuming refers to #858. The fact that a player is raising this exception in `update_texture` is bad, since it likely means the texture actually lives in a different context than the player is switching to, which could only have happened if https://github.com/pyglet/pyglet/blob/9fe927d1322f2e845288c9b10a54f1ded12b4d08/pyglet/media/player.py#L384-L387 was somehow called from another context than what's in the player's `_context`. That's plausible considering the method does not switch to the player's context and is in one place called from `_set_playing`:
https://github.com/pyglet/pyglet/blob/9fe927d1322f2e845288c9b10a54f1ded12b4d08/pyglet/media/player.py#L190-L192

Could you maybe try adding a `with self._context` block here and see if your program still throws the error?

matanox:
@Square789 If I'm not terribly mistaken the lines of code which you point at are not reached in case the player is only used for audio playing as in my case (which I invoke from the context of a key press handler). I have wrapped it with this context now, and some debug printing and it seems it's not being called indeed.

matanox:
@benmoran56 Thanks as usual. I have instrumented my application with various more informational loggings, awaiting my transient pyglet.gl.lib.GLException: (0x1282): Invalid operation to happen again as I have failed miserably in managing to reproduce it by demand. I'll be a bit wiser on this when it happens again over coming weeks, and then address it outside this issue, which seems resolved.

> It may be enough to enhance the VertexDomain classes to store/switch_to their contexts when drawing, as deletion is already handled by the context deletion pools.

Maybe the above is technically a worthy trajectory for making pyglet users write multi-window code more fluently, but since I've not pinpointed whether/how/why my application (as a case in point) manages to coerce pyglet into that exception even while not using more than one window at all, I'll keep waiting for more diagnostics next time I am lucky to get it happen again. 

I might find that avoiding users having to `switch_to` carries great benefit for all those who write multi-window, by abstracting over this facet for them within the library itself, either way. In case it's low and runtime cost and logically tractable. Or for example, provide idioms for checking/enumerating contexts through pyglet api. That said the complexities and limitations of abstracting over OpenGL are as you know a little beyond me.

PR #862, opened 2023-06-02
Allows to pause the player at the end of the video

jej:
It was one of my needs. To stall the video at the last frame, instead of rewinding...

`player.pause_at_end = True` and the video stays on the last frame.

In the spirit of `loop`. Needs review.

### DIFF
--- a/pyglet/media/player.py
+++ b/pyglet/media/player.py
@@ -122,6 +122,12 @@ class Player(pyglet.event.EventDispatcher):
         #: .. versionadded:: 1.4
         self.loop = False
 
+        #: Pause the current source on the last frame.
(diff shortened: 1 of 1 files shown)


benmoran56:
This can be accomplished by setting a `on_eos` handler.

In your case you can probably just replace the default handler directly:
```py
player.on_eos = player.pause
```

jej:
Working very well! Thanks.

So I can close this PR.

Maybe I could update the doc to explain this trick?

benmoran56:
I was already working on the docs, so I added a note to the end of the "controlling playback" section in the docs:
https://pyglet.readthedocs.io/en/latest/programming_guide/media.html#controlling-playback

PR #864, opened 2023-06-03
Improve Atlas Type Hints

gran4:
The fact that the tests are failing seems to be a problem with the linter settings.

### DIFF
--- a/pyglet/image/atlas.py
+++ b/pyglet/image/atlas.py
@@ -23,7 +23,7 @@ the application's responsibility to keep track of the regions returned by the
 
 .. versionadded:: 1.1
 """
-from typing import TYPE_CHECKING, Tuple
+from typing import TYPE_CHECKING, Tuple, Optional
 
 import pyglet
 
@@ -162,7 +162,7 @@ class TextureAtlas:
         self.texture = pyglet.image.Texture.create(width, height)
         self.allocator = Allocator(width, height)
 
-    def add(self, img: AbstractImage, border: int = 0) -> TextureRegion:
+    def add(self, img: 'AbstractImage', border: int = 0) -> 'TextureRegion':
         """Add an image to the atlas.
 
         This method will fail if the given image cannot be transferred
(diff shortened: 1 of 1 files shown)


pyglet/image/atlas.py
```diff
@@ -40,13 +40,13 @@ class AllocatorException(Exception):
...
 
-    def __init__(self, y: int, max_height: int):
+    def __init__(self, y: int, max_height: int) -> None:
         self.x = 0
         self.y = y
         self.max_height = max_height
         self.y2 = y
 
-    def add(self, width: int, height: int):
+    def add(self, width: int, height: int) -> Tuple[float, float]:
```
benmoran56, line 49: Return types are int.

benmoran56 (changes requested):

pyglet/image/atlas.py
```diff
@@ -252,7 +252,7 @@ class TextureArrayBin:
     ones as necessary as the depth is exceeded.
     """
 
-    def __init__(self, texture_width=2048, texture_height=2048, max_depth=None):
+    def __init__(self, texture_width: int=2048, texture_height: int=2048, max_depth: Optional[int]=None):
```
shenjackyuanjie, line 255: how about `texture_width: int = 2048` ?

pyglet/image/atlas.py
```diff
@@ -145,7 +145,7 @@ def get_fragmentation(self) -> float:
 class TextureAtlas:
     """Collection of images within a texture."""
 
-    def __init__(self, width: int = 2048, height: int = 2048):
+    def __init__(self, width: int = 2048, height: int = 2048) -> None:
```
shenjackyuanjie, line 148: over here, there is a gap between `int = 2048`

pyglet/image/atlas.py
gran4: It should be without spaces

shenjackyuanjie: > It should be without spaces

well, my pycharm say's it should have spaces?

shenjackyuanjie:
image

watch this,  PEP8 E252

gran4:
Pyglet does it without spaces. @benmoran56 is this an issue?

benmoran56 (approved):

PR #867, opened 2023-06-06
Adds super() call in on_resize() method

jej:


benmoran56:
This isn't really appropriate, because that section of the code is specifically describing _replacing_ default event handlers.

I can see the need for some clarity in the documentation however, so I'll update this myself soon with some more examples.

jej:
Up to you.

benmoran56:
The docs have been updated with a bit more information on this behavior.

PR #872, opened 2023-06-08
Fix MacOS Text issues

caffeinepills:
Fix issue with TextView not able to receive events for text input purposes.

Fix issue with some zero length characters causing an IndexError.

Add some estimates for fallback metric detection for fonts for Mac OS (Labels and Documents). This is incase the font being used does not have the glyphs requested and the system falls back to another font (such as Chinese).  It is probably best you use the font that directly supports the language for the most accurate measurements, but this should help prevent text from breaking down into a jumbling of glyphs.

### DIFF
--- a/pyglet/font/quartz.py
+++ b/pyglet/font/quartz.py
@@ -7,7 +7,7 @@ from ctypes import c_void_p, c_int32, byref, c_byte
 from pyglet.font import base
 import pyglet.image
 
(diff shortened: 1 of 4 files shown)


shenjackyuanjie:
image
`你好！现在好了`
works well!

benmoran56 (approved):

PR #879, opened 2023-06-30
Improvements to pyglet.gui

jason-bowen-zheng:
This PR didn't add any features to pyglet.

Instead, `Frame` can change its cell info when widgets change their position, which wasn't implemented before.

Also, users can choose to enable or disable a frame.

### DIFF
--- a/pyglet/gui/frame.py
+++ b/pyglet/gui/frame.py
@@ -77,39 +77,29 @@ class Frame:
 
     def on_mouse_press(self, x, y, buttons, modifiers):
         """Pass the event to any widgets within range of the mouse"""
-        if not self._enable:
-            return
         for widget in self._cells.get(self._hash(x, y), set()):
             widget.on_mouse_press(x, y, buttons, modifiers)
             self._active_widgets.add(widget)
 
     def on_mouse_release(self, x, y, buttons, modifiers):
         """Pass the event to any widgets that are currently active"""
-        if not self._enable:
-            return
         for widget in self._active_widgets:
             widget.on_mouse_release(x, y, buttons, modifiers)
         self._active_widgets.clear()
 
     def on_mouse_drag(self, x, y, dx, dy, buttons, modifiers):
         """Pass the event to any widgets that are currently active"""
-        if not self._enable:
-            return
         for widget in self._active_widgets:
             widget.on_mouse_drag(x, y, dx, dy, buttons, modifiers)
         self._mouse_pos = x, y
 
     def on_mouse_scroll(self, x, y, index, direction):
         """Pass the event to any widgets within range of the mouse"""
-        if not self._enable:
-            return
         for widget in self._cells.get(self._hash(x, y), set()):
             widget.on_mouse_scroll(x, y, index, direction)
 
     def on_mouse_motion(self, x, y, dx, dy):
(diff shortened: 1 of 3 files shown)


benmoran56:
I had some ideas about this. Rather than coupling frame and widgets, how about using events? 
BaseWidget could dispatch a new event, like `on_reposition`.  It can dispatch this event whenever you change the x, y, or position. When you add a Widget to a Frame, the Frame can register this event.

For example, in the Widget:
```py
# WidgetBase
    @position.setter
    def position(self, values):
        self._x, self._y = values
        self._update_position()
        self.dispatch_event('on_reposition', self)
```
And in the Frame:
```py
    def _on_reposition_handler(self, widget):
        ...

    def add_widget(self, widget):
        """Add a Widget to the spatial hash."""
        min_vec, max_vec = self._hash(*widget.aabb[0:2]), self._hash(*widget.aabb[2:4])
        for i in range(min_vec[0], max_vec[0] + 1):
            for j in range(min_vec[1], max_vec[1] + 1):
                self._cells.setdefault((i, j), set()).add(widget)
        widget.update_groups(self._order)
        widget.set_handler('on_reposition', self._on_reposition_handler)
```

Frame disable is a good idea. Actually, I thought about separating `Window` and `Frame` entirely. 
Currently, you have to pass a Window instance to Frame. I think it might be better to make it explicit. For example:
```py
my_frame = Frame()
window.push_handlers(my_frame)
```
That way, you can just set or remove the events at any time. Extra code doesn't need to be added to the Frame class.

jason-bowen-zheng:
About the first idea, unluckly, it won't work **now** because of the strange hehavior of `Frame.remove_widget`. But, I can rewrite the code of `remove_widget`.

In my opinion, I refuse to accept your second idea for:

> (Assume we have clear docs and a beginner read it) `Frame.enable` is meaningful, which shows this property control the enable or disable state of frame.

Using `Frame.enable` is clear, direct and easy to understand!

benmoran56:
Widget changes look good. Thanks! 

One thing I noticed. Is it necessary to do: 
```py
        if not self._enable:
            return
            ....
```
When setting enable = False,  we have already called `window.remove_handlers`. The events should not reach the handlers anymore.

benmoran56:
Looks good! 

There is one final consideration.  Adding `enable` property/setter is OK, because it's backwards compatible. We can merge this directly into the master branch. 

If we add `enable=True` to `__init__`, then it changes the API. If we do that, we should target pyglet 2.1 (the `development` branch). 

We have two options: 
1. Change target from master -> development.  The code is OK as-is. 
2. Remove the `enable=True` from `__init__`, and merge into master. (You can also make a separate PR for `development`, which includes the `enable=True`). 

What do you think?

jason-bowen-zheng:
Thanks for your advice, I have changed it.

jason-bowen-zheng:
I choose option 1, but I don't know how to do it.

Should I reopen a PR?

benmoran56 (approved):
Looks good to go.

benmoran56:
Merged!  I was able to do it through Github by changing the target. 

One small change:  For consistency, I will rename `enable` to `enabled`. This grammatical change will match other pyglet objects, such as `Sprite.paused`.

PR #881, opened 2023-07-01
WIC encoding fix.

caffeinepills:
Use the correct calculated pitch, image.pitch is unreliable for ImageRegions due to a bug as of 2.0.8.

### DIFF
--- a/pyglet/image/codecs/wic.py
+++ b/pyglet/image/codecs/wic.py
@@ -1,5 +1,3 @@
-import warnings
-
 from pyglet.image import *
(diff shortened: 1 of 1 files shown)


caffeinepills:
Not sure why the OSX test failed on an audio driver test here, but only for 3.10.

benmoran56:
We can ignore that failure. It's just due to load in the CI machine. We have some pytest checks to detect CI and skip certain tests, but we need to update for GitHub Actions.

benmoran56 (approved):

Issue #890, opened 2023-07-08
Screen resolution changes on Retina Mac running windowed Arcade app


wronkiew:
**Describe the bug**
I posted an issue for this at pythonarcade/arcade#1830 but the problem is in pyglet.

Running a windowed Arcade app changes the resolution of the display. This behavior started after plugging my laptop into an external monitor. Quitting the app restores the display to the original resolution.

It seems likely that the problem is related to [an old bug with CGDisplaySetDisplayMode](https://lists.apple.com/archives/cocoa-dev/2014/May/msg00295.html). The laptop is running on integrated graphics before running the app, and switches to high-performance during. Also, if I run the Chess app first it starts in high-performance graphics mode, and the problem goes away.

**System Information:**
```
Platform
------------------------------------------------------------------------------
platform:  macOS-10.16-x86_64-i386-64bit
release:   21.6.0
machine:   x86_64

Python
------------------------------------------------------------------------------
implementation: CPython
sys.version: 3.9.10 (main, Jan 15 2022, 12:08:01) 
[Clang 11.0.3 (clang-1103.0.32.62)]
sys.maxint: 9223372036854775807
PyObjC not available
os.getcwd(): /Users/zzz/zzz

pyglet
------------------------------------------------------------------------------
pyglet.version: 2.0.dev23
pyglet.compat_platform: darwin
pyglet.__file__: /Users/zzz/zzz/venv/lib/python3.9/site-packages/pyglet/__init__.py
pyglet.options['audio'] = ('xaudio2', 'directsound', 'openal', 'pulse', 'silent')
pyglet.options['debug_font'] = False
pyglet.options['debug_gl'] = True
pyglet.options['debug_gl_trace'] = False
pyglet.options['debug_gl_trace_args'] = False
pyglet.options['debug_gl_shaders'] = False
pyglet.options['debug_graphics_batch'] = False
pyglet.options['debug_lib'] = False
pyglet.options['debug_media'] = False
pyglet.options['debug_texture'] = False
pyglet.options['debug_trace'] = False
pyglet.options['debug_trace_args'] = False
pyglet.options['debug_trace_depth'] = 1
pyglet.options['debug_trace_flush'] = True
pyglet.options['debug_win32'] = False
pyglet.options['debug_input'] = False
pyglet.options['debug_x11'] = False
pyglet.options['shadow_window'] = True
pyglet.options['vsync'] = None
pyglet.options['xsync'] = True
[...]

PR #891, opened 2023-07-09
Fix screen resolution changes on Retina Mac running windowed Arcade app

wronkiew:
This fixes the easy part of #890 where starting a windowed arcade app changes the screen resolution on Mac. When a Window is created it calls set_fullscreen. If the Window is not fullscreen then this calls Screen.restore_mode. restore_mode calls CGDisplaySetDisplayMode with the mode saved when CocoaScreen was created. This is not necessary because the screen mode was never changed. In the meantime setting up OpenGL may have changed the GPU from integrated to discrete. This exposes a bug in CGDisplaySetDisplayMode where it sets the screen to a random resolution. You can fix it by calling it again but this takes a long time. If the default mode is the same as the current mode, skip calling CGDisplaySetDisplayMode. There may still be cases where the current screen mode is not the same as the default mode, this does not fix those cases. But it fixes the problem I ran into.

### DIFF
--- a/pyglet/canvas/cocoa.py
+++ b/pyglet/canvas/cocoa.py
@@ -78,6 +78,12 @@ class CocoaScreen(Screen):
         self.height = mode.height
 
     def restore_mode(self):
+        match_attrs = ['width', 'height', 'depth', 'rate']
+        current_mode = self.get_mode()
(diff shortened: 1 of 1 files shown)


benmoran56:
Thank you for the pull request!

PR #894, opened 2023-07-14
Use degrees for angles in Arc and Sector

### LINKED ISSUE #868: API inconsistency: degrees vs radians in shapes.Arc
### What's Wrong

The `rotation` property of shapes takes degrees. However, `pyglet.shapes.Arc` appears to take angles in radians:
https://github.com/pyglet/pyglet/blob/204badc2cf23f236e726b22220f067984a9073db/pyglet/shapes.py#L431-L468

### Question: How should we address this?

Options I see include:

1. Add a `rotation_radians` property to shapes
2. Make `Arc` take degrees by default
3. Add degrees versions of the angle keyword arguments to arc

Which combination of options seems best?

pushfoo:
### Changes

Closes #868 by switching `angle` and `angle_start` to use degrees per discussion in that issue's comments.

### Follow-up Work

It may be a good idea to add radians versions of these properties to prevent having to convert between degrees and radians.

### DIFF
--- a/examples/graphics/shapes.py
+++ b/examples/graphics/shapes.py
@@ -33,7 +33,7 @@ class ShapesDemo(pyglet.window.Window):
             batch=self.batch
         )
 
-        self.arc = shapes.Arc(50, 300, radius=40, segments=25, angle=4, color=(255, 255, 255), batch=self.batch)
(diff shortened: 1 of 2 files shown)


benmoran56 (approved):

benmoran56:
This looks good. I'm thinking we might want to target the 2.1 release to avoid breakage.

benmoran56 (approved):

PR #905, opened 2023-07-28
Optimizations to experimental geoshader Sprite

cspotcode:
After learning about pyglet's Sprite, and learning that pyglet's drawables are generic, and shapes and sprites can be combined into the same batch, I wondered how it compares against arcade's Sprite.  My benchmarks showed pyglet is slower, so I dug into the code and made some optimizations.  These optimizations make pyglet's sprite faster than arcade's in my specific benchmark, creating thousands of sprites and moving each one every frame.

I made a variety of changes in separate commits, then ran the same benchmark against each commit.  This table was generated by hyperfine and shows the relative performance of each.  The most recent (fastest) commit is at the top, and the bottom commit is the vanilla geoshader Sprite currently on master.

By looking at the relative timings, you can see which code changes had a positive/negative/negligible impact on performance.

This PR probably should not be merged as-is.  Rather, the most impactful performance improvements can be extracted to another PR.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
[9bdf59dc1 Sprite uses VertexList optimized setter for position updates](https://github.com/pyglet/pyglet/commit/9bdf59dc15d616d5658cae7bf4278ced896f3741) | 58.694 ± 0.850 | 57.771 | 59.443 | 1.00 |
[6da2b35dc Add to VertexDomain and VertexList attribute.buffer_array and optimized setters to write values into the buffers](https://github.com/pyglet/pyglet/commit/6da2b35dc6c2462b0457f45cf98c0220d6a4aab5) | 103.905 ± 0.867 | 103.351 | 104.903 | 1.77 ± 0.03 |
[751d79e2a Refactor sprite from _x, _y, _z attributes to _position tuple attribute](https://github.com/pyglet/pyglet/commit/751d79e2ad9e4a9203ed5c8e9b37c1e1b7dd1958) | 105.724 ± 1.881 | 104.177 | 107.818 | 1.80 ± 0.04 |
[dab20d922 Make Sprite slotted](https://github.com/pyglet/pyglet/commit/dab20d922c85a109c7dfd4751e7ebfe3a2e16814) | 111.933 ± 1.746 | 110.405 | 113.837 | 1.91 ± 0.04 |
[7b9d3a019 Optimize VertexList region cache](https://github.com/pyglet/pyglet/commit/7b9d3a019e9b30a0132fbe11b1ce3b501360c215) | 108.820 ± 0.793 | 107.953 | 109.507 | 1.85 ± 0.03 |
[fcdc49d1c optimize BufferObjectRegion.invalidate](https://github.com/pyglet/pyglet/commit/fcdc49d1cb3ba1b55e0d42e8c6bb3872231b4ed7) | 109.228 ± 1.010 | 108.533 | 110.386 | 1.86 ± 0.03 |
[42841c287 Move geoshader_sprite so it's accessible from editable installs of pyglet](https://github.com/pyglet/pyglet/commit/42841c287a64fbcbd097ce31f314d9ef503cf89d) | 127.479 ± 4.137 | 123.112 | 131.339 | 2.17 ± 0.08 |


pyglet/graphics/vertexbuffer.py
```diff
@@ -309,5 +309,7 @@ def invalidate(self):
         until the next time the buffer is used, for efficiency).
         """
         buffer = self.buffer
-        buffer._dirty_min = min(buffer._dirty_min, self.start)
-        buffer._dirty_max = max(buffer._dirty_max, self.end)
+        if self.start < buffer._dirty_min:
```
pushfoo, line 312: Explaining why we aren't using the more concise syntax will save everyone time by preventing reversion attempts.
```suggestion
        # Intentionally avoid min and max functions to increase performance as of 3.8 - 3.12
        if self.start < buffer._dirty_min:
 ```

File: pyglet/graphics/vertexdomain.py [hunk 2]
```diff
@@ -103,6 +104,18 @@ def __init__(self, program, attribute_meta):
...
         for attribute in self.attributes:
             self.attribute_names[attribute.name] = attribute
+        
+        self._create_attribute_buffer_arrays()
+
+    def _create_attribute_buffer_arrays(self):
+        """
+        For each attribute, create a ctypes array pointing to the entire buffer
+        which can be used to write values into the buffer using slice assignment
+        """
```
pushfoo, line 114: Minor style nitpick: pyglet's docstring conventions place the summary on the same line as the opening quotes.
```suggestion
        """For each attribute, create a ctypes array pointing to to its entire buffer.

        This allows using slice assignment syntax to write values into the buffer.
        """
```

pyglet/graphics/vertexdomain.py
```diff
@@ -292,9 +306,29 @@ def migrate(self, domain):
...
+        self._cache_version = None
+
+    def get_setter(self, name):
+        """
+        Returns a function that, when called, sets the values for a given attribute.
+
+        The setter becomes invalid when the VertexList migrates to a different
+        domain, and it's the caller's responsibility to stop using the old setter
+        and create a new one.
+        """
```
pushfoo, line 318: 1. Same docstring style nitpick as before
2. Phrasing tweaks
3. We can make debugging easier by explaining what happens when you fail to migrate to the new setter so the issue can be identified
```suggestion
        """Return a function which sets the value(s) of the named attribute.

        The setter is invalidated whenever the VertexList migrates to a different
        domain. It's the caller's responsibility to stop using the old setter and
        create a new one when this happens.
        """
```

pushfoo (changes requested):
I've suggested changes which will save contributors and users time in the future, as well as some style nitpicks.

caffeinepills:
Slots may have to be omitted as this will break existing code or pushed further into a release where we can break compatibility.

Personally I am not a fan of restricting sprites to this degree in general. Although it's probably fine for a new class such as `GeoSprite` that is not already being used.

cspotcode:
Yeah, I talked a little bit about the slots change in the PR description, though didn't call it out by name.  It's the bit saying "Some of these changes are bad and should not be merged! ..."

cspotcode:
@benmoran56 I cherry-picked your code that uses a custom VertexList subclass per VertexDomain.

Benchmark results are below.  The fastest is to *combine* our two optimizations, which is weird considering the closure/partial-bound `_position_setter` does not use the generated VertexList property.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
[631c70bff Sprite sets position with `Sprite._position_setter(position)` which is bound closure.  Custom `VertexList` subclass instead of `VertexList.__getattr__`/`__setattr__`](https://github.com/pyglet/pyglet/commit/631c70bfff39b04b6f0b666888116e4ec5cc7799) | 69.797 ± 7.609 | 65.220 | 78.581 | 1.05 ± 0.12 |
[6411c53f2 Sprite sets position with `Sprite._vertext_list.position[:] = position`  Custom `VertexList` subclass instead of `VertexList.__getattr__`/`__setattr__`](https://github.com/pyglet/pyglet/commit/6411c53f2b14db7758d1561ada90bff70d14d49e) | 80.551 ± 1.871 | 78.423 | 81.941 | 1.21 ± 0.03 |
[944be0026 Sprite sets position with `Sprite._position_setter(position)` which is bound closure](https://github.com/pyglet/pyglet/commit/944be002663303c420f44f98201f54a2c9c04202) | 66.445 ± 0.779 | 65.574 | 67.076 | 1.00 |
[3111f9267 Sprite sets position with `Sprite._position_setter(position)` which is bound `functools.partial()`](https://github.com/pyglet/pyglet/commit/3111f92679eafe2b589b40f17f92d915bdf4d931) | 68.395 ± 2.051 | 66.372 | 70.472 | 1.03 ± 0.03 |
[23aa26e4a Refactor sprite from _x, _y, _z attributes to _position tuple attribute](https://github.com/pyglet/pyglet/commit/23aa26e4a26b1e4c146a6b30767ca2d67844a7f1) | 112.209 ± 0.927 | 111.213 | 113.047 | 1.69 ± 0.02 |
[4029cba7e Make Sprite slotted](https://github.com/pyglet/pyglet/commit/4029cba7e43500db3b9b689c86d770ea0c488fc8) | 115.144 ± 1.829 | 113.902 | 117.244 | 1.73 ± 0.03 |
[7d721bf14 Optimize VertexList region cache](https://github.com/pyglet/pyglet/commit/7d721bf1484e636098f5d7cd89367df8e6317f0c) | 115.094 ± 5.124 | 110.604 | 120.676 | 1.73 ± 0.08 |
[c8ee652d4 optimize BufferObjectRegion.invalidate](https://github.com/pyglet/pyglet/commit/c8ee652d476dcdaf83e472a7005abe69d7032b20) | 112.646 ± 0.818 | 111.924 | 113.535 | 1.70 ± 0.02 |
[cd7e06225 Move geoshader_sprite so it's accessible from editable installs of pyglet](https://github.com/pyglet/pyglet/commit/cd7e062257f24fbd1e095545e14ac47901a76953) | 130.280 ± 2.801 | 127.517 | 133.117 | 1.96 ± 0.05 |

cspotcode:
I made one final optimization, this one focused on memory.  Ran a final benchmark.

In short, it's now 2.3 times as fast.

For fairness, this table shows the same 4 commits ran *twice* to make sure the first ran again on a hotter CPU.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
[28b2790d1 further optimize position setter to avoid closures and functools.partial, reduce memory usage](https://github.com/pyglet/pyglet/commit/28b2790d11b55eaeaf03a8a31ac279b240fec8bc) | 61.748 ± 0.290 | 61.568 | 62.083 | 1.00 |
[631c70bff Sprite sets position with `Sprite._position_setter(position)` which is bound closure.  Custom `VertexList` subclass instead of `VertexList.__getattr__`/`__setattr__`](https://github.com/pyglet/pyglet/commit/631c70bfff39b04b6f0b666888116e4ec5cc7799) | 68.715 ± 1.132 | 67.955 | 70.016 | 1.11 ± 0.02 |
[6411c53f2 Sprite sets position with `Sprite._vertext_list.position[:] = position`  Custom `VertexList` subclass instead of `VertexList.__getattr__`/`__setattr__`](https://github.com/pyglet/pyglet/commit/6411c53f2b14db7758d1561ada90bff70d14d49e) | 84.336 ± 3.822 | 81.029 | 88.520 | 1.37 ± 0.06 |
[cd7e06225 Move geoshader_sprite so it's accessible from editable installs of pyglet](https://github.com/pyglet/pyglet/commit/cd7e062257f24fbd1e095545e14ac47901a76953) | 136.987 ± 3.303 | 133.767 | 140.368 | 2.22 ± 0.05 |
[28b2790d1 further optimize position setter to avoid closures and functools.partial, reduce memory usage](https://github.com/pyglet/pyglet/commit/28b2790d11b55eaeaf03a8a31ac279b240fec8bc) | 63.626 ± 0.626 | 62.986 | 64.238 | 1.03 ± 0.01 |
[631c70bff Sprite sets position with `Sprite._position_setter(position)` which is bound closure.  Custom `VertexList` subclass instead of `VertexList.__getattr__`/`__setattr__`](https://github.com/pyglet/pyglet/commit/631c70bfff39b04b6f0b666888116e4ec5cc7799) | 70.843 ± 2.088 | 69.538 | 73.251 | 1.15 ± 0.03 |
[6411c53f2 Sprite sets position with `Sprite._vertext_list.position[:] = position`  Custom `VertexList` subclass instead of `VertexList.__getattr__`/`__setattr__`](https://github.com/pyglet/pyglet/commit/6411c53f2b14db7758d1561ada90bff70d14d49e) | 85.504 ± 3.634 | 81.669 | 88.897 | 1.38 ± 0.06 |
[cd7e06225 Move geoshader_sprite so it's accessible from editable installs of pyglet](https://github.com/pyglet/pyglet/commit/cd7e062257f24fbd1e095545e14ac47901a76953) | 140.134 ± 4.354 | 135.209 | 143.470 | 2.27 ± 0.07 |

pyglet/graphics/vertexdomain.py
```diff
@@ -292,33 +336,60 @@ def migrate(self, domain):
...
 
-        for version in self._cache_versions:
-            self._cache_versions[version] = None
+    def get_setter_params(self, name):
+        """
+        Returns a tuple of params which can be passed to
+        `vertex_list_set_attribute_values` along with a data tuple to set values
+        for the attribute.
+
+        These params become invalid when the VertexList migrates to a different
```
benmoran56, line 347: This might be a blocker, unfortunately. For the default Batch, users should be able to add/remove/migrate things between Batches transparently.

cspotcode: This is safe for `Sprite` to use internally, right?

`Sprite._batch` and `Sprite._vertex_list` are private.  For a user to migrate a Sprite between batches, they have to assign to `Sprite.batch`, `Sprite.program`, etc.  And all those properties call `VertexList.get_setter_params`.

If a user is using `VertexDomain` / `VertexList` directly, they can safely avoid `get_setter_params` and instead use the same API that sprites do today: get a `BufferObjectRegion` via `VertexList.__getattr__` or the generated properties on `_vertexlist_class`.

benmoran56:
Thanks for this work. It helped a lot to improve many points. With pyglet v3.0 things are being redesigned a bit internally, but many of the ideas here were useful.

PR #909, opened 2023-07-31
Update examples for 2.1 by scheduling redraw

pushfoo:
### Changes

tl;dr schedule drawing intervals per [description of breaking changes on Discord](https://discord.com/channels/438622377094414346/477394331112701952/1134825532354216058):
> ...there is a breaking change for 2.1. You need to explicitly scheudle the Window redraw

### ~~I need help with~~ Done

- [x] Testing Windows-specific examples:
     - [x] `python examples/text/advanced_font.py`
     - [x] `python examples/text/font_comparison.py`

- [x] Testing controller input examples (Works according to Ben)
     - [x] `python examples/input/controller.py`
     - [x] `python examples/input/joystick.py`

### Follow-up questions

1. Should we add frame rate constants to make scheduling cleaner?
2. Q: How should we describe the pros and cons of using `schedule` vs `schedule_interval` for readers?
   A: Examples should use `schedule_interval`

### DIFF
--- a/examples/dpi_scaled_window.py
+++ b/examples/dpi_scaled_window.py
@@ -40,4 +40,6 @@ def on_scale(scale, dpi):
     print("Window Pixel Ratio:", window.get_pixel_ratio())
     print("Window Frame Buffer Size:", window.get_framebuffer_size())
 
+
+pyglet.clock.schedule_interval(window.draw, 1/60)
 pyglet.app.run()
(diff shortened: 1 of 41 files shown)


benmoran56:
I made a few comments. Other than those, lets change all instances of `pyglet.clock.schedule` to `pyglet.clock.schedule_interval`.

benmoran56:
I tested the controller and joystick examples on my machine, and they work fine after adding schedule_interval.

benmoran56:
Windows specific examples are tested OK.

benmoran56 (approved):

PR #929, opened 2023-08-26
Optimize geoshader Sprite buffer writes into VertexDomain

cspotcode:
Extracted from #905 

Adds an optimized code path for writing values into a VertexList.  Sprite and other pyglet internals can use this code path.

This code path can only be used by things which exclusively control their VertexList, because they *must* know every time the VertexList is migrated.  This is true for Sprite, since `_vertex_list` is private, and the only way to migrate a Sprite's VertexList is by calling Sprite methods or properties.

Makes code changes for experimental geoshader Sprite to use this optimized code path.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
Before: [448a3ee4 Fix shader (pre-existing bug in geoshader Sprite, unrelated to my changes)](https://github.com/pyglet/pyglet/commit/448a3ee430be44024e5e16c86766d66a818791ab) | 317.981 ± 2.675 | 315.556 | 320.851 | 1.70 ± 0.02 |
After: [31c607fc Optimize buffer writes from geoshader Sprite into VertexDomain](https://github.com/pyglet/pyglet/commit/31c607fc69b43720d8315016fd46f9ec33f6e901) | 186.733 ± 0.799 | 185.929 | 187.527 | 1.00 |

### DIFF
--- a/pyglet/experimental/geoshader_sprite.py
+++ b/pyglet/experimental/geoshader_sprite.py
@@ -7,6 +7,7 @@ from pyglet import clock
 from pyglet import event
 from pyglet import graphics
 from pyglet import image
+from pyglet.graphics.vertexdomain import _vertex_list_set
 
 _is_pyglet_doc_run = hasattr(sys, "is_pyglet_doc_run") and sys.is_pyglet_doc_run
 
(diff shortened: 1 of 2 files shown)


benmoran56:
@cspotcode Even though we didn't merge this directly, a lot of these ideas made their way into pyglet. Thanks for lighting a fire with this PR. :bow:

PR #930, opened 2023-08-26
added on_resize method to widgets so their size can be changed after initialization

CORRUPTOR2037:


matanox:
Currently `Window.set_size` has this line of code in it:

```python
self.dispatch_event('on_resize', width, height)
```

Which means that when called by application code, it may immediately turn the flow of control to an application registered callback handler for the `on_resize` event. This is not trivial for an application expecting its call to `set_size` to immediately return, whereas instead the next line of application code to execute will be that of its `on_resize` registered handler (which is a very awkward flow of control, even if applications can be written to safely manage this flow of control). 

I hope this PR avoids introducing or repeating this kind of unexpected confluence of application code. 

It's mildly related to https://github.com/pyglet/pyglet/pull/1116 in that the latter explains how `Window.dispatch_event` works in these regards.

caffeinepills:
I will close this, as it's really old and no further communication by the PR owner. 

This does more than just add an `on_resize`, it changes the existing behavior of the existing widgets (for example the TextWidget appears as large as the window due to the resizing.) That beign said, I like the direction of this so I will add an `on_resize` event to the frame and widgets as this behavior is still currently missing.

PR #931, opened 2023-08-26
Add textentry hotkeys

CORRUPTOR2037:


pushfoo:
This stacks onto #930.

pushfoo (changes requested):
tl;dr of my review:

1. This is more complicated than it seems due to cross-platform issues with macs
2. I made a PR (#935) which documents part of the steps needed to address it 
3. I'll implement a change as described by that PR as well as follow-ups, then make a PR into your branch so we can have a mac user test the end product


I think the best approach is to do the following:

- [x] Find someone with a mac who can test mac-related changes while we wait for #935 to be merged
- [x] Merge `master` into my touchup branch
- [ ] I'll make a PR into your branch to keep this moving since it's crucial functionality I didn't know pyglet was missing
- [ ] Get the mac user to test Cmd-A
- [ ] Fix any bugs or problems
- [ ] Notify the leads the PR is ready

benmoran56:
Sorry for leaving this hanging for so long. To be short, I cherry picked some of the changes from this PR, and committed to master branch. They will be in the next release.
On careful consideration, the on_resize methods were not included. This is a bit out of scope for the default classes, and not easy for us to support in a uniform way across widgets. It's easy enough to add this behavior in a base class. With that in mind, it would make more sense for us to add something like that as an example, instead of supporting it in the main code. 

I'll close this for now, but please feel free to follow up if you want to discuss further.

PR #940, opened 2023-09-01
Only set up VAOs once

Square789:
VAOs hold information explaining attribute data format/layout and in which buffers the data resides, as well as which shader input the attribute is supposed to be routed to.
Right now, this gets redefined in all four draw functions each call.
This PR reduces the amount of work done in draw calls by only running this setup code on a VertexDomain's initialization together when its VAO is initially created.

Some more polish could be done by seperating the mapped buffer object's `bind` method into something like `upload_dirty` or `commit_changes`; right now this causes buffers that don't need it to be bound.

While I only tested it by running the example catalogue (which worked flawlessly) and slapping some `perf_counter` and `print` statements around `on_draw` in `examples/graphics/shapes.py`, the results seem promising:
```
Before      After
0.03033679  0.00488889
0.00029201  0.00025898
0.00027020  0.00018485
0.00026285  0.00014822
0.00024151  0.00015732
0.00027604  0.00015676
0.00024167  0.00014386
0.00027857  0.00015118
0.00024365  0.00022371
0.00026760  0.00015312
0.00035942  0.00017602
0.00040726  0.00014547
0.00028442  0.00017375
0.00028811  0.00018602
0.00026247  0.00017569
0.00029515  0.00013578
(Time the first few clear and batch.draw calls took, in seconds)
```
I did not explicitly test the probably rather seldom-used `draw_subset` methods, a 2nd look is definitely necessary.

### DIFF
--- a/pyglet/graphics/vertexdomain.py
+++ b/pyglet/graphics/vertexdomain.py
@@ -79,7 +79,6 @@ class VertexDomain:
         self.program = program
         self.attribute_meta = attribute_meta
         self.allocator = allocation.Allocator(self._initial_count)
-        self.vao = vertexarray.VertexArray()
 
         self.attributes = []
         self.buffer_attributes = []  # list of (buffer, attributes)
@@ -98,6 +97,15 @@ class VertexDomain:
             attribute.buffer.element_size = attribute.stride
             attribute.buffer.attributes = (attribute,)
             self.buffer_attributes.append((attribute.buffer, (attribute,)))
+
+        self.vao = vertexarray.VertexArray()
+        self.vao.bind()
+        for buffer, attributes in self.buffer_attributes:
+            buffer.bind()
+            for attribute in attributes:
+                attribute.enable()
+                attribute.set_pointer(buffer.ptr)
+        self.vao.unbind()
 
         # Create named attributes for each attribute
         self.attribute_names = {}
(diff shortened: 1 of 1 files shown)


Square789:
I chose not to bind the index buffer for the IndexedVertexDomain as the buffer is a pain `BufferObject`, but if it should ever become mapped, that omission is not possible anymore.

benmoran56:
Thanks, I've been meaning to do this. The old code was largely left over from 2.0 days. 
I don't see any possible issues with draw subset, but we should test that. It usually gets used when calling the direct draw method on an object.

Square789:
This should hit all the four draw methods through a batch including lines (non-indexed vertex list) and sprites (indexed), as well as shapes; non-indexed ones and as the only indexed one BorderedRectangle: [stuffdrawer.py](https://github.com/pyglet/pyglet/files/12508079/stuffdrawer.py.txt)
I can't notice any different behavior running this before and after the commit.

All examples in `game`, `graphics`, `gui`, `input` (those i could run with only a keyboard and mouse), `media`, `model`, `opengl`, `sprite`, `text` (those available on linux) and `window`*, as well as `file_dialog.py`, `hello_world.py` and `timer.py` work just fine with the commit applied.

*The overlay and transparent examples produce a segfault on exit for me, but also do so before the commit.

benmoran56 (approved):

benmoran56:
Thanks for taking the time to test that. I'll merge this in, and test locally as well.

Issue #888, opened 2023-07-08
Pyglet.Shapes - Add Triangle And Line Rotation Support


nbroderick-code:
**Is your feature request related to a problem? Please describe.**
When I set ".rotation" on a triangle or line, it does not rotate.

**Describe the solution you'd like**
I would like it to rotate these shapes around their center point instead of doing nothing because it isn't implemented.

I appreciate all your hard work! :)

shenjackyuanjie:
Technically to implement `rotation` for triangle and line is simple
but here's the thing
which point should be the center?
the shape's first point?  the shape's geometry center? or some weird center you pointed out?
so which center?

nbroderick-code:
The geometry's center is what most programs do, I think.

pushfoo:
This came up again today [on Discord](https://discord.com/channels/438622377094414346/477394331112701952/1147648807622934590): `Circle` is also affected since it's an n-gon approximation.

I thought it had already been addressed, and apparently so did @caffeinepills :

> I think that makes sense to have some rotation even for circles. 
> Technically the circle could be multiple colors and if you want to spin it even.
> I thought all shapes had a rotation property

The simplest fix which comes to mind is moving this to `ShapeBase`, which I've partly addressed in #943.

To consider this issue finished, I think we need to support `rotation` as a keyword argument on all shapes. `Star` already supports it.

However, there is a problem:  `Line.__init__` sets the `rotation` from he slope of the points: 
https://github.com/pyglet/pyglet/blob/1af5a55f7aedac4a5245e7a306bb1c7cc1bdeda8/pyglet/shapes.py#L1102


How should we resolve Line's behavior? Options include:

1. `rotation: Optional[float] = None` as the rotation argument. If not `None`, override any inferred rotation from geometry data.
4. Remove the rotation keyword argument.

PR #943, opened 2023-09-03
Allow all ShapeBase subclasses to rotate

pushfoo:
Partially addresses #888 

EDIT:  This behavior is left unaddressed, but I didn't intend to fix it in this PR:

> As bestos: so the position works as rotation center and anchor as rotation offset

[Video link](https://discord.com/channels/438622377094414346/463526340168122368/1146295218824548402)

### Changes

1. Add `_rotation` class attribute and `rotation` property to `ShapeBase`
2. Remove `rotation` property from shapes
3. Add top-level new rotation value unit test fixture
4. Add tests for rotation value on shape `__init__`, skipping `Line`
5. Add tests for rotation property setter on all shapes
6. Add shape rotation  & translation visualization utility

### How to test

1. `pytest -k tests/unit`
2. Run `examples/graphics/shapes.py` and look for issues
2. Run `tools/shape_rotation_and_translation_spot_check.py` and interact with it, expecting the following issues:
    1. Shapes default to rotating around their lower left corner
    2. The following exceptions rotate around their centers: star, circle, ellipse, sector, and arc
    3. Beziers and Polygons have broken position properties and "teleport" toward the origin due to ignoring `_x` and _`y`
    4. Line sets its initial rotation value from the slope between the points

### Remaining tasks for this PR
- [x] Update `ShapeBase.rotation` per [Ben's comment](https://github.com/pyglet/pyglet/pull/943#discussion_r1319259653)
- [ ] Add interactive rotation tests iterating over shape classes
     - [ ] Add base interactive unit test for rotation
     - [ ] Figure out if fixtures should be re-used
     - [ ] Present context information for each shape:
          - [ ] "add some additional information to the test, such as explaining the default anchor point under the name. Just an idea, but something like: (anchor=center) or (anchor=bottom left) or (anchor=first vertex)."
          - [ ] Display info about rotation before test action is triggered
- [ ] Clean up visualization utility
     - [ ] Docstring at top of file
     - [ ] Clean up other docstrings
     - [ ] Spacing double check / linter run
     - [ ] Update [gist](https://gist.github.com/pushfoo/6102cc93713b1f6b9c95467aa76d3a61)

### Follow-up work (not yet assigned)

1. Resolve [the dilemma](https://github.com/pyglet/pyglet/issues/888#issuecomment-1704288681) over how to handle a future `rotation` keyword for `Line`
2. Resolve general shape argument / anchor issues
3. https://github.com/pyglet/pyglet/issues/946

### DIFF
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -198,6 +198,7 @@ class ShapeBase(ABC):
     """
 
     _rgba = (255, 255, 255, 255)
+    _rotation = 0
     _visible = True
     _x = 0
     _y = 0
@@ -258,6 +259,19 @@ class ShapeBase(ABC):
         """
         raise NotImplementedError("_update_vertices must be defined"
                                   "for every ShapeBase subclass")
+    @property
+    def rotation(self) -> float:
+        """Clockwise rotation of the shape in degrees.
+
+        It will be rotated about its (anchor_x, anchor_y)
+        position.
+        """
+        return self._rotation
+
+    @rotation.setter
+    def rotation(self, rotation: float) -> None:
+        self._rotation = rotation
+        self._vertex_list.rotation[:] = (rotation,) * self._num_verts
 
     def draw(self):
         """Draw the shape at its current position.
@@ -550,22 +564,6 @@ class Arc(ShapeBase):
                 vertices.extend(chord_points)
 
         self._vertex_list.position[:] = vertices
-
-    @property
-    def rotation(self):
-        """Clockwise rotation of the arc, in degrees.
-
-        The arc will be rotated about its (anchor_x, anchor_y)
-        position.
-
-        :type: float
-        """
-        return self._rotation
-
-    @rotation.setter
-    def rotation(self, rotation):
-        self._rotation = rotation
-        self._vertex_list.rotation[:] = (rotation,) * self._num_verts
 
     @property
     def angle(self):
@@ -902,22 +900,6 @@ class Ellipse(ShapeBase):
         self._b = value
         self._update_vertices()
 
-    @property
-    def rotation(self):
-        """Clockwise rotation of the arc, in degrees.
-
-        The arc will be rotated about its (anchor_x, anchor_y)
-        position.
-
-        :type: float
-        """
-        return self._rotation
-
-    @rotation.setter
-    def rotation(self, rotation):
-        self._rotation = rotation
-        self._vertex_list.rotation[:] = (rotation,) * self._num_verts
-
 
 class Sector(ShapeBase):
     def __init__(self, x, y, radius, segments=None, angle=math.tau, start_angle=0,
@@ -1047,22 +1029,6 @@ class Sector(ShapeBase):
     def radius(self, value):
         self._radius = value
         self._update_vertices()
-
-    @property
-    def rotation(self):
-        """Clockwise rotation of the sector, in degrees.
-
-        The sector will be rotated about its (anchor_x, anchor_y)
-        position.
-
-        :type: float
-        """
-        return self._rotation
-
-    @rotation.setter
-    def rotation(self, rotation):
-        self._rotation = rotation
-        self._vertex_list.rotation[:] = (rotation,) * self._num_verts
 
 
 class Line(ShapeBase):
@@ -1286,22 +1252,6 @@ class Rectangle(ShapeBase):
         self._height = value
         self._update_vertices()
 
-    @property
-    def rotation(self):
-        """Clockwise rotation of the rectangle, in degrees.
-
-        The Rectangle will be rotated about its (anchor_x, anchor_y)
-        position.
-
-        :type: float
-        """
-        return self._rotation
-
-    @rotation.setter
-    def rotation(self, rotation):
-        self._rotation = rotation
-        self._vertex_list.rotation[:] = (rotation,) * self._num_verts
-
 
 class BorderedRectangle(ShapeBase):
     def __init__(self, x, y, width, height, border=1, color=(255, 255, 255),
@@ -1448,22 +1398,6 @@ class BorderedRectangle(ShapeBase):
     def height(self, value):
         self._height = value
         self._update_vertices()
-
-    @property
-    def rotation(self):
-        """Clockwise rotation of the rectangle, in degrees.
-
-        The Rectangle will be rotated about its (anchor_x, anchor_y)
-        position.
-
-        :type: float
-        """
-        return self._rotation
-
-    @rotation.setter
-    def rotation(self, rotation):
-        self._rotation = rotation
-        self._vertex_list.rotation[:] = (rotation,) * self._num_verts
 
     @property
     def border_color(self):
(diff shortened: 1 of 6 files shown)


pushfoo:
Polygon and triangle both appear to rotate around their first vertex rather than their centers, but the PR's scope was intended to make shapes rotate, not make them consistent.

zhengxyz123:
Please don't forget to update the documentation. Previously, to simplify the documentation, only the special properties of each shape were listed separately.

Now you should add the `rotation` property to `ShapeBase` and remove the `rotation` documentation for the other shapes.

pushfoo:
I've updated the doc. If the maintainers are ok with it, we could:
1. merge this
2. leave the rotation fix for a separate PR

[zhengxyz123](https://github.com/zhengxyz123) and As bestos have examined the behavior more thoroughly than I have.

benmoran56:
This looks good, but we should test this before merging. 
Rotation might break rendering, **if** the shape position is also changed. Some shapes (triangle I think) will behave strangely. 

The cause is that we do a lot of tricks to improve performance when moving shapes, because this is the most common task.

This would be an ideal case for making an `interactive` test, as the only real way to confirm the behavior is to test it visually.

pushfoo:
Since we haven't actually agreed on how shapes should behave or what "broken" rotation is, I added [`tools/shape_rotation_translation_spotcheck.py`](https://github.com/pyglet/pyglet/pull/943/files#diff-635471c27d174ded4cc2875e12e0cc45c02b711a74ea4eebadcc2b34d40e9b61) instead of interactive tests (also [available as a gist](https://gist.github.com/pushfoo/6102cc93713b1f6b9c95467aa76d3a61)).

What we do have for certain on `master`:

1. Rotation "works" aside from pre-existing inconsistencies:
    1. `Line` sets an initial rotation value based on the slope between its points
    2. Shapes use either the bottom left or their center as their rotation anchors
2. Beziers and Polygons have working bottom left rotation, but their translation was already broken before I changed anything

The latter two classes "teleport" closer to the origin when their position is set because their `_update_vertices` ignores `_x` and `_y`.

| Before movement | After Movement |
| :--: | :--: |
| image | image |

As mentioned earlier, rotation actually seems to work aside from the inconsistencies outlined earlier:

| Before rotation | After rotation |
| :--: | :--: |
| image | image |

pyglet/shapes.py
```diff
@@ -258,6 +259,19 @@ def _update_vertices(self):
         """
         raise NotImplementedError("_update_vertices must be defined"
                                   "for every ShapeBase subclass")
+    @property
+    def rotation(self) -> float:
+        """Clockwise rotation of the shape in degrees.
+
+        It will be rotated about its (anchor_x, anchor_y)
+        position.
```
benmoran56, line 267: It would probably be a good idea to add something like:   "which defaults to the first vertex point of the shape"

benmoran56:
Great work on that visualization code.

> 1. Rotation "works" aside from pre-existing inconsistencies:
>    
>    1. `Line` sets an initial rotation value based on the slope between its points
>    2. Shapes use either the bottom left or their center as their rotation anchors
> 2. Beziers and Polygons have working bottom left rotation, but their translation was already broken before I changed anything
> 
> The latter two classes "teleport" closer to the origin when their position is set because their `_update_vertices` ignores `_x` and `_y`.

For 2), since these are pre-existing bugs, we can open a new issue to fix those. They don't need to block this PR. 

As far as an interactive test, is up to us to determine that the current visual state is correct.
The same test should then be used before/after any further changes to the Shapes, that might introduce bugs. 

We might also add some additional information to the test, such as explaining the default anchor point under the name. Just an idea, but something like: (anchor=center) or (anchor=bottom left) or (anchor=first vertex).

pyglet/shapes.py
pushfoo: I didn't add any notes on this because I figured we would be changing the rotation behavior soon.

How about something like the RST equivalent of this?

Shapes currently have inconsistent rotation behavior:

* Shapes default to rotating around their lower left corner
* Star, circle, ellipse, sector, and arc rotate around their centers

pushfoo:
> Great work on that visualization code.

Ty! If you add some protocol types and bound tracking, it's sort of the direction I'd like to move the general concept of "layout" toward (`Callable[[item_with_width_and_height, Optional[context_params], Tuple[int, int]]`)

> since these are pre-existing bugs, we can open a new issue to fix those. They don't need to block this PR.

Agreed, this is what I was hoping for.

> As far as an interactive test, is up to us to determine that the current visual state is correct.

I'm not sure I understand. Is the intent of the test only to ask the user  "Does this $SHAPE look right when it rotates?" as a reminder?

It seems like we'd want to be specific for the test to be meaningful. That requires having a decision on what "correct" is. If you want that test in this PR, then I need either of the following:

1. A consensus or project lead decision on what "correct" is
2. Confirmation that I should write a test for the current ugly, broken behavior

benmoran56:
> I'm not sure I understand. Is the intent of the test only to ask the user "Does this $SHAPE look right when it rotates?" as a reminder?

That's really the whole point of the interactive tests. The test runner look at the screen, and confirm that it's doing what it says it should be doing.
The test message must say what to expect:   "Shapes should rotate around their anchor points (as specified). Shapes should not change appearance after repositioning, etc. "

I can personally confirm the existing shapes. If someone modifies, or adds a new shape, it's at that point where we need to verify that it looks OK.

pushfoo:
> "Shapes should rotate around their anchor points (as specified). Shapes should not change appearance after repositioning, etc. "

This is the missing piece of information, thank you. I'll include the differing anchor points for the shape classes as part of the text template for the interactive test. We can update it later as needed.

> Shapes should not change appearance after repositioning, etc. 

Given this PR was intended to be limited to rotation, would it be ok to split interactive translation tests into a follow-up PR if the PR starts getting big? The visualization code would stay in this one since it's a prerequisite for visualizing interactions (or lack thereof) between the two components.

benmoran56:
Yes, that would be fine as well.

pyglet/shapes.py
jej: > Shapes currently have inconsistent rotation behavior

It was discussed here but closed...

#840 & #841

benmoran56: It's inconsistent, but kind of expected when dealing with OpenGL primitives. In any case we shouldn't break behavior for a 2.0 release. 

We can consider this change for 2.1 (development branch), if someone is willing to put in the work.
At a glance, I think the following would be a reasonable approach:
1. Add code to calculate the centroid of each Shape (might be tricky for some shapes), to determine the anchor offsets.
2. change `position` in the shader to a vec4, and place the anchor offset values there.

This should be in it's own thread for future discussion, if someone is willing to put the work in.

pushfoo: I've done the following to address this thread's concerns:

1. Added a commit with the suggested information + a list of exceptions
2. Linked Ben's comment above in the relevant Discord thread so it isn't forgotten.

benmoran56:
I was able to fix the re-positioning for the Polygon and BezierCurve shapes, so I'm going to merge this in (and then apply my fixes). 

When time allows, we can later take your example code and apply other visual checks for the anchor changes.

pushfoo:
Ah, ok. Sorry for the delay. My time's been a bit fragmented lately so I've been focusing on doc and incremental improvement PRs.

PR #954, opened 2023-09-13
Check Homebrew for common libraries on macOS

pythonengineer:
I don't know if this is the most optimal way of including the path, but some Mac users will have the libraries Pyglet can depend upon (like for audio) installed through Homebrew and they may not be caught through the existing library paths that are checked by default.

### DIFF
--- a/pyglet/lib.py
+++ b/pyglet/lib.py
@@ -79,8 +79,8 @@ class LibraryLoader:
         platform = 'win32'
 
     def load_library(self, *names, **kwargs):
-        """Find and load a library.  
-        
+        """Find and load a library.
+
         More than one name can be specified, they will be tried in order.
         Platform-specific library names (given as kwargs) are tried first.
 
@@ -164,6 +164,9 @@ class MachOLibraryLoader(LibraryLoader):
         else:
             self.dyld_fallback_library_path = [os.path.expanduser('~/lib'), '/usr/local/lib', '/usr/lib']
 
+        # check homebrew for libs too
+        self.brew_library_path = ['/opt/homebrew/lib']
+
     def find_library(self, path):
         """Implements the dylib search as specified in Apple documentation:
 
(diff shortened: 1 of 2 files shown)


pyglet/lib.py
```diff
@@ -164,6 +164,9 @@ def __init__(self):
         else:
             self.dyld_fallback_library_path = [os.path.expanduser('~/lib'), '/usr/local/lib', '/usr/lib']
 
+        # check homebrew for libs too
+        self.brew_library_path = ['/opt/homebrew/lib']
```
pushfoo, line 168: This might be better as a keyword argument. The doc seems to strongly recommend against changing the directory, [but it does seem to support it in theory](https://docs.brew.sh/Installation#untar-anywhere-unsupported) if I understood it correctly.

pyglet/lib.py
```diff
@@ -205,11 +208,13 @@ def find_library(self, path):
             search_path.extend([os.path.join(p, libname) for p in self.dyld_library_path])
             search_path.append(path)
             search_path.extend([os.path.join(p, libname) for p in self.dyld_fallback_library_path])
+            search_path.extend([os.path.join(p, libname) for p in self.brew_library_path])
```
pushfoo, line 211: This seems like it shouldn't be an issue, but I'd double check to make sure this doesn't break conda or venvs.

pushfoo (changes requested):
tl;dr:
1. brew seems like it provides but discourages a way to customize install directory
2. the brew directory should probably be a keyword argument with stern warnings

pythonengineer:
The keyword argument with warning is added.

pyglet/lib.py
benmoran56: This should be fine as long as it's at the end. Other locations will already have been tried.

benmoran56:
Closing due to lack of follow up.

PR #955, opened 2023-09-13
1.5: Fix coreaudio arg mismatch (+ brew check)

pythonengineer:
This fixes a crash in the 1.5 branch that occurs when Pyglet tries to load the coreaudio media codec, and includes the open PR #954 for additional Mac support.

### DIFF
--- a/pyglet/lib.py
+++ b/pyglet/lib.py
@@ -113,8 +113,8 @@ class LibraryLoader:
         platform = 'win32'
 
     def load_library(self, *names, **kwargs):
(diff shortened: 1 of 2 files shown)


benmoran56:
Thanks for the PR! Shall we close the other one?

pythonengineer:
> Thanks for the PR! Shall we close the other one?

The other PR contains the same change for Homebrew, but is targeting the master (2.0) branch as opposed to 1.5.

PR #956, opened 2023-09-17
Overwrite `pyglet.font` for complex application

zhengxyz123:
See #908, I override the `create_font`(and rename it to `add_font`) and add an example for it.


caffeinepills:
I apologize I left this open without commenting, I did comment in Discord but I will post it here for clarifications and further input by you and others.

1) I do really like the idea of creating a font via class inheritance instead of creating a create_font function. That is a nice change.

I however would prefer we change the name from `add_font` to something more specific such as `add_user_font` or `add_created_font` to make the function more clear that this is not how you add a file font or system font.

2) The example should also not enforce PIL, it should be optional for demonstration purposes. The initial use case for this is a single bitmap font size provided by the user. It would probably be best there is a basic example showing this usage without PIL. The other example can be with PIL scaling.

3) The example should have much more docstrings at the top, about it's usage and what it does. As well as the functions `get_sga_mappings` and `get_texture_ascii` . The examples are meant to be an explanation and demonstration of the code, so including that would be nice.

zhengxyz123:
I have renamed the function from`add_font` to `add_user_font` and added comments to examples.

I'm thinking of the example that don't use a PIL, and should probably create a new file.

caffeinepills:
Looks good so far, just needs a non-PIL example, and should be good. Thanks!

pyglet/font/__init__.py
```diff
@@ -253,4 +224,4 @@ def add_directory(directory):
             add_file(os.path.join(directory, file))
 
 
-__all__ = ('add_file', 'add_directory', 'load', 'have_font')
+__all__ = ('add_file', 'add_directory', 'create_font', 'load', 'have_font')
```
caffeinepills, line 227: `create_font` no longer exists.

caffeinepills (changes requested):

caffeinepills:
I am going to close this PR. I have created a new PR based off yours. #1067 (Added in 2.0.13)

This one consolidates both of your uses cases into one UserDefined font. The mapping can be passed as a Dict-like object or a Dict (just needs a get function). This way it handles both use cases. Also the resizing is done within the renderer itself, and allows use of the user font without requiring the resizing. Example will be added once we decide on a free to distribute font.

Issue #952, opened 2023-09-12
Pulse Audio Exception - Hard to Reproduce


DavidEBrumbaugh:
**Describe the bug**
I'm using Python Arcade 3.0.0.dev25 and Pyglet 2.0.0.9.  Randomly (usually when I pause a lot when debugging, but also when I've been running the game for a while) I get the following exception thrown:
`Assertion 'q->front' failed at pulsecore/queue.c:81, function pa_queue_push(). Aborting.`

The bug is clearly in pulse audio.  It was listed as a "closed issue":  https://github.com/mackron/miniaudio/issues/235
And fixed here:
https://github.com/mackron/miniaudio/commit/50c9081b19fe4e25c781c78b6b7acfbc5ffdef58

**System Information:**

```
pyglet.version: 2.0.9
pyglet.compat_platform: linux
pyglet.__file__: /home/david/.virtualenvs/venv_arcade/lib/python3.10/site-packages/pyglet/__init__.py
pyglet.options['audio'] = ('xaudio2', 'directsound', 'openal', 'pulse', 'silent')
pyglet.options['debug_font'] = False
pyglet.options['debug_gl'] = True
pyglet.options['debug_gl_trace'] = False
pyglet.options['debug_gl_trace_args'] = False
pyglet.options['debug_gl_shaders'] = False
pyglet.options['debug_graphics_batch'] = False
pyglet.options['debug_lib'] = False
pyglet.options['debug_media'] = False
pyglet.options['debug_texture'] = False
pyglet.options['debug_trace'] = False
pyglet.options['debug_trace_args'] = False
pyglet.options['debug_trace_depth'] = 1
pyglet.options['debug_trace_flush'] = True
pyglet.options['debug_win32'] = False
pyglet.options['debug_input'] = False
pyglet.options['debug_x11'] = False
pyglet.options['shadow_window'] = True
pyglet.options['vsync'] = None
pyglet.options['xsync'] = True
pyglet.options['xlib_fullscreen_override_redirect'] = False
pyglet.options['search_local_libs'] = True
pyglet.options['win32_gdi_font'] = False
pyglet.options['headless'] = False
pyglet.options['headless_device'] = 0
pyglet.options['win32_disable_shaping'] = False
pyglet.options['dw_legacy_naming'] = False
pyglet.options['win32_disable_xinput'] = False
pyglet.options['com_mta'] = False
pyglet.options['osx_alt_loop'] = False

pyglet.window
------------------------------------------------------------------------------
[...]

PR #957, opened 2023-09-21
Warn about PulseAudio in the doc and rework related sections

pushfoo:
### Changes

1. Document the PulseAudio bug (#952) 
2. Add "Choosing audio drivers" heading
3. `%s/Pulseaudio/PulseAudio/g` and rename the Pulse heading to PulseAudio for consistency
4. Expand the PulseAudio section with better pros / cons + bug
5. Expand the OpenAL section with Linux install command examples and other info
6. Add link targets for relevant headings
7. Rearrange + add table footnotes for cleanliness
8. Link the new doc from the quickstart
9. Link the doc from `pyglet/__init__.py`s doc for the `options['audio']` key

### DIFF
--- a/doc/programming_guide/media.rst
+++ b/doc/programming_guide/media.rst
@@ -124,11 +124,26 @@ support hardware audio mixing or surround sound.
 OpenAL
 ^^^^^^
 
-OpenAL is included with Mac OS X. Windows users can download a generic
-driver from either `openal.org`_ or their sound device's manufacturer.
+By default, this is the most preferred driver. It will be used first
+if it is installed unless overridden. See :ref:`guide-audio-driver-order`
+to learn more.
 
-If a Linux distribution does not come with OpenAL pre-installed, it can
-usually be installed through the package manager.
+It is the best target for cross-platform consistency due to being either
+preinstalled or easy to install on all supported platforms. It implements
+software versions of common game audio features such as positional mixing
+on platforms which may not support them by default.
+
+OpenAL's main downside is that platforms other than Mac OS X are not
+guaranteed to have it preinstalled.
+
+Windows users can download a generic driver from either `openal.org`_
+or their sound device's manufacturer.
+
+On Linux, it may already be installed. It is strongly recommended to
+install it since the :ref:`guide-pulseaudio` backend has serious bugs
+and limitations. The commands to install OpenAL on the most common
+Linux distros are listed below. You may need to use ``sudo`` to run
+them.
 
 .. list-table::
     :header-rows: 1
@@ -145,7 +160,12 @@ usually be installed through the package manager.
     * - Fedora, Nobara
       - ``dnf install openal-soft``
 
-OpenAL does not appear to have a package on recent versions of RHEL.
+These commands should be safe to run when OpenAL is already
+installed. You will be informed the package is already installed
+if this is the case, and no action will take place.
+
+Note that OpenAL does not appear to have a package on recent versions of
+RHEL.
 
 .. _guide-pulseaudio:
 
--- a/doc/programming_guide/media.rst
+++ b/doc/programming_guide/media.rst
@@ -194,14 +194,14 @@ occurs:
 It is not known to be affected by whether the PulseAudio implementation
 is the stand-alone or PipeWire compatibility layer.
 
-Do not expect the bug to be fixed quickly for multiple reasons:
+Consider the following when deciding whether to use this back-end:
 
 #. :ref:`guide-openal` is better in almost every way
-#. It unpleasant to fix because it appears to be a concurrency problem.
-   See the `original github issue <https://github.com/pyglet/pyglet/issues/952>`_
-   and discussion below it for more information.
+#. The `current understanding <https://github.com/pyglet/pyglet/issues/952>`_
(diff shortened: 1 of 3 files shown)


doc/programming_guide/media.rst
```diff
@@ -95,25 +119,102 @@ DirectSound is available only on Windows, and is installed by default.
...
-from `openal.org`_, or from their sound device's manufacturer. Most Linux
-distributions will have OpenAL available in the repositories for download.
-For example, Arch users can ``pacman -S openal`` and Ubuntu users can ``apt install libopenal1``.
+By default, this is the most preferred driver. It will be used first
+if it is installed unless overridden. See :ref:`guide-audio-driver-order`
+to learn more.
 
-Pulse
-^^^^^
+It is the best target for cross-platform consistency due to being either
```
benmoran56, line 131: I wouldn't necessary say it's the best for cross-platform, since XAudio2 (or even DirectSound) are fully featured with regards to positional audio. Instead, we _could_ say it's "currently the best choice for posix systems" (or just leave it out).

doc/programming_guide/media.rst
```diff
@@ -95,25 +119,102 @@ DirectSound is available only on Windows, and is installed by default.
...
+The bug is known to crash programs when one or more of the following
+occurs:
+
+#. A debugger pauses or resumes the program
+#. Unpredictably when 2 or more sounds are playing
+
+It is not known to be affected by whether the PulseAudio implementation
+is the stand-alone or PipeWire compatibility layer.
+
+Do not expect the bug to be fixed quickly for multiple reasons:
```
benmoran56, line 196: I would probably leave this out.

doc/programming_guide/media.rst
```diff
@@ -95,25 +119,102 @@ DirectSound is available only on Windows, and is installed by default.
...
+#. PulseAudio is on its way to being replaced by PipeWire, so replacing
+   this driver may be a better use of developer time than fixing it.
+
+Missing features
+""""""""""""""""
+
+PulseAudio was designed with only desktop computer workloads in mind.
+This means it lacks built-in support for common features such as spatial
+audio mixing. The following features will not work properly with this
+driver:
```
benmoran56, line 212: This might read as being misleading to some people, since PulseAudio itself does have support for advanced multi-channel audio.

It would be more accurate to say that it's pyglet's PulseAudio backend that is limited to stereo.

doc/programming_guide/quickstart.rst
```diff
@@ -161,6 +161,16 @@ An example program using keyboard and mouse events is in
 Playing sounds and music
 ------------------------
 
+.. note::
```
benmoran56, line 164: In my opinion, this is too much to add in the quickstart guide.

pushfoo: Agreed. I've cut this down and moved it to the bottom near the footnote.

doc/programming_guide/media.rst
```diff
@@ -95,25 +119,102 @@ DirectSound is available only on Windows, and is installed by default.
...
+    * - Ubuntu, Pop!_OS, Debian
+      - ``apt install libopenal1``
+
+    * - Arch, Manjaro
+      - ``pacman -S openal``
+
+    * - Fedora, Nobara
+      - ``dnf install openal-soft``
+
+These commands should be safe to run when OpenAL is already
```
benmoran56, line 164: I would say that it's not necessary to explain this part, since it's dependent on the package manager (and they'll all pretty much say the same thing anyway).
I also wouldn't bother with the RHEL comment.

caffeinepills:
Personally I am not agreeing with some of the language, and there are some errors:

Calling the PulseAudio buggy implies it's unusable. I think just putting warnings of it may having issues with those 2 issues you mentioned. We've only had one report about such issues and could be system, version, or something else going on.

The 'default' section on audio is incorrect, it says it is: `('openal', 'pulse', 'xaudio2', 'directsound', 'silent')` but that was just an example to show how to re-arrange the order. The default is actually `('xaudio2', 'directsound', 'openal', 'pulse', 'silent')`

Pyglet does not technically consider OpenAL 'preferred', and the language implies people should install it on their platforms. That should be up to the developer if they wish their users to use OpenAL on all platforms.

The default behavior of Pyglet is just so that sound just works on all OS platforms as much as possible, it will go through all drivers until it finds one.  OpenAL should be recommended over PulseAudio simply because of positional audio. However, Windows has XAudio2 (and DirectSound) backends which is more feature complete than OpenAL. 

My 2c.

pushfoo:
@caffeinepills I think I got everything you and Ben mentioned. PulseAudio is now described as limited rather than broken.

doc/programming_guide/quickstart.rst
```diff
@@ -190,9 +190,15 @@ The `examples/media_player.py` example demonstrates playback of streaming
 audio and video using pyglet.  The `examples/noisy/noisy.py` example
 demonstrates playing many short audio samples simultaneously, as in a game.
 
+.. note:: Linux users may want to install OpenAL.
```
benmoran56, line 193: I would remove this part. It's a little too much information for the Quick Start guide IMO.

benmoran56 (approved):

PR #958, opened 2023-09-27
Clarify licensing & fix style + typos in PyQT5 example

pushfoo:
tl;dr PyQT5 is GPL / commercial dual-licensed, which means the example should mention that.

This is a follow-up to discussion around changes to https://github.com/pyglet/pyglet/commit/18ec6130fd591828eeeddd07a8adbd1ece942ac1.

### DIFF
--- a/contrib/qt_sprite_preview.py
+++ b/contrib/qt_sprite_preview.py
@@ -1,5 +1,14 @@
-"""Requires: PyQT5
-   An example using PyQT5 and Pyglet together.
+"""An example sprite previewer using PyQT5 and Pyglet together.
+
+   Important:
+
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

caffeinepills:
Just for clarification you can just replace `PyQt5` with `PySide2` and it still works and behaves the same way. (If you have a need for LGPL that is)

pushfoo:
I figured it was worth investigating, but as a separate PR. Thank you for doing most of the work!

This file also has CLRF which I was going to fix as a follow-up, although that's sort of insignificant since most IDEs can automatically adapt to make the file consistent.

Issue #926, opened 2023-08-16
Memory Leak on Windows when playing audio files


HauseMasterZ:
When using pyglet.media to play a playlist of music audio files on Windows, memory usage continually increases by around 2-10 MB for every audio file based on its length. This is happening both when streaming is set to True or False.
I have tested on .m4a .mp3 and .wav file and all of them exhibit this behaviour.

OS: Windows 10 22H2
Python: 3.11.4
pyglet.version: 2.0.9

The following code reproduces the error. It reads from a folder and then loads the audio files using media.load and then queues it to the player. The program initially consumes around 40MB but for each new song that is played it grows by 2-10MB everytime without any stopping.

This issue is very similar to [this](https://github.com/pyglet/pyglet/issues/283) but it has been marked as solved yet I am still facing this issue in the latest pyglet version.

```
import pyglet
import os

path = 'path/to/music/folder' #change this to the path to your music folder

pool = []
audio = set('.mp3 .wav .ogg .flac .m4a'.split(' '))
for root, dirs, files in os.walk(path, followlinks = True) :
  for name in files:
    if os.path.splitext(name)[1] in audio :
        pool.append(os.path.join(root, name))

player = pyglet.media.Player()

def queue_next() :
  s = pool.pop(0)
  print("Queueing: %s" % (s))
  player.queue(pyglet.media.load(s, streaming = True))
  pool.append(s) 

def on_eos() :
  player.next_source()
  queue_next()

player.on_eos = on_eos

queue_next()
queue_next()
player.play()

window = pyglet.window.Window(100,100,)


pyglet.app.run()
```

caffeinepills:
The issue referencing is regards to Linux, not Windows, so that is not related unfortunately.

We need a little more information, what kind of files are you mostly testing? Each of the difference formats have different decoders, so it may just be one affected or multiple.

benmoran56:
Just to add, there are two potential areas for memory leaking:  The audio backend, and the decoders. 

I would recommend to first start with Wave files ONLY (the source type should show as `pyglet.media.codecs.wave.WaveSource`). This decoder doesn't have any chance of leaks as it's just a simple byte reader.

If the leaks go away, then it's likely an issue with the decoder being used. Please provide the info as caffeinepills requested. 

If you are still seeing leaks, it _could_ be an issue with the default audio backend (which is XAudio2 on Windows). Try manually switching to the legacy DirectSound backend to see if there is any improvement:
```py
import pyglet
[...]

PR #962, opened 2023-10-04
Fix some use-after-frees and memory leak in the wmf decoder

Square789:
This says "some" cause the video part is intentionally using a use-after-free to expose the frame.
In any case, the primary reason for this PR is to fix the jumping leak from https://github.com/pyglet/pyglet/issues/926 stemming from the `_source_reader` not being released, now that the other one turned out to be a Python bug.

### DIFF
--- a/pyglet/media/codecs/wmf.py
+++ b/pyglet/media/codecs/wmf.py
@@ -398,8 +398,6 @@ class WMFSource(Source):
 
     def __init__(self, filename, file=None):
         assert any([self.decode_audio, self.decode_video]), "Source must decode audio, video, or both, not none."
-        self._current_audio_sample = None
-        self._current_audio_buffer = None
         self._current_video_sample = None
         self._current_video_buffer = None
         self._timestamp = 0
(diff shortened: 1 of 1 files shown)


caffeinepills:
This causes audio to skip and crackle on my machine.

Square789:
But even before, the audio data was copied via `memmove`, there's no reasonable way the playback mechanism could still have any kind of access to the IMFBuffer/Samples. Maybe i'm releasing the `imf_sample` too early, the MS docs don't mention anything explicitly, but maybe it deletes its buffers even if they still have references.

caffeinepills:
I updated to the latest with sample being freed last, and it seems to have fixed the crackling for me. (Windows 11, Xaudio2, ogg/mp3 testing)

caffeinepills:
Also, the video portion of WMF can probably just be removed in the future as it shouldn't be used anymore.

benmoran56 (approved):

benmoran56:
Shall we merge this for now, and then strip out the video parts in the `development` branch?

Square789:
I don't really know what the status with the wmf decoder's video section is. I could not get it to play any video in the first place. The _debug at https://github.com/pyglet/pyglet/blob/45fc7d1b6cc0ff61a5090e52eed6659bb9233dcf/pyglet/media/codecs/wmf.py#L576 is triggered, but then this bit: https://github.com/pyglet/pyglet/blob/45fc7d1b6cc0ff61a5090e52eed6659bb9233dcf/pyglet/media/codecs/wmf.py#L590-L593 always fails and reports `[WinError -1072875852] The data specified for the media type is invalid, inconsistent, or not supported by this object` for a variety of videos in different formats i've thrown at it.
I was focusing this PR on fixing issues in the audio section only; i feel any changes regarding the video part are kinda out of scope for this PR

pushfoo:
I'm still hesitant about tearing out the WMF video decoding. If we can get MP4 to work and establish a baseline for the video stream across Mac & Windows, we'd have a 0-pyglet video equivalent to MP3. That could be a significant decrease in download size.

caffeinepills:
> always fails and reports `[WinError -1072875852] The data specified for the media type is invalid, inconsistent, or not supported by this object` for a variety of videos in different formats i've thrown at it.
> I was focusing this PR on fixing issues in the audio section only; i feel any changes regarding the video part are kinda out of scope for this PR

The formats are somewhat limited I believe. Video stuff can be removed at a later date.  

> I'm still hesitant about tearing out the WMF video decoding. If we can get MP4 to work and establish a baseline for the video stream across Mac & Windows, we'd have a 0-pyglet video equivalent to MP3. That could be a significant decrease in download size.

Unfortunately with WMF it's not hardware accelerated without DirectX and I am not quite up to the task of creating a DirectX-OpenGL link just to make video work for whatever formats it supports when FFmpeg does it all better. 

We can leave the video stuff for another PR later down the road.

PR #967, opened 2023-10-09
Some COM interface changes

Square789:
This PR improves on the COM module a bit
- COMObjects do not create a vtable for each instance anymore with the instance's methods bound to the vtable. Instead, the vtable points to a method extracting `self` from a PyObject pointer stored in the generated structs.
  - This removes the `this` parameter found in many Python methods.
- Methods are no longer property acessors that create their prototype, turn it into a COM method and then call it for each and every call.
- Add the `VOIDMETHOD`, which is more easy to understand that `METHOD(None)`.
- Removal of some unused methods and general rewrite of most metaclasses.

As far as i can tell (testing in a rickety VM), it works just as before, but more testing with an actual Windows system would be really welcome


caffeinepills:
Currently, this breaks the `LegacyFontFileLoader` for DirectWrite (loading custom fonts).

Square789:
Next time i'm gonna take a closer look at the test directories, whoops
I tried these new two commits on a project adding a custom font with my old actual win7 partition (which from what i can tell will cause the legacy font loader to be used), and it worked perfectly, while beforehand the error about `_pointers` not existing popped up, similar to the failing test.

caffeinepills:
> Next time i'm gonna take a closer look at the test directories, whoops I tried these new two commits on a project adding a custom font with my old actual win7 partition (which from what i can tell will cause the legacy font loader to be used), and it worked perfectly, while beforehand the error about `_pointers` not existing popped up, similar to the failing test.

You could technically probably comment out the Windows check, it *should* still work on newer OS's. Probably should be something we add to unit tests. 

Glancing over and minimal testing I don't see many issues. Nice work! Working on COM's can be tricky so definitely appreciate you getting it more up to date.

benmoran56 (dismissed):

benmoran56:
Looks good to me too, but I don't understand a lot of the com stuff myself. 
Shall we merge this in?

Square789:
I don't think there's any more problems with it. I've tested code paths using all COMObjects there are (custom font loading, XAudio2 callbacks and audio device notification) both on 3.11.6 32bit and 64bit and those work just like they did before.

There are more interface definitions that are affected by this commit. I did not test all codepaths that those are used in, primarily the image decoders and input devices, but DirectSound, XAudio2 and the wmf decoder still work as before.
So unless the other parts do something really outlandish with the Interfaces (which from a quick skim for `getattr`, `setattr`, `__interface__` and `_methods_`: they don't), this should not negatively affect anything else.

caffeinepills:
So far it's fine on my tests. I would just add the suggestions I put in the review, then it can be merged.

Square789:
> So far it's fine on my tests. I would just add the suggestions I put in the review, then it can be merged.

By suggestion, are you referring to this?
> Probably should be something we add to unit tests or add Windows 7 + 3.8 as a test suite.

Github doesn't provide windows 7 machines, so not really sure how to go about that. Possibly also not worth it as the OS is EOL and Python 3.8 will be dropped 14 Oct 2024.
I guess it would be possible to add a fixture with `params` that monkeypatches all the windows version variables to `False` and then `WINDOWS_VISTA_OR_GREATER` through `WINDOWS_10_CREATORS_UPDATE_OR_GREATER` to `True` increasingly, but i do not know how one would apply that to select unit tests only where it matters and also while not repeating any code.

caffeinepills:
No, the review I made with comments.

caffeinepills (changes requested):

caffeinepills:
Weird, the review I made with comments didn't show up. I just resubmitted.

Square789:
I think that should do it then. Of course i just haaaad to change something else in https://github.com/pyglet/pyglet/pull/967/commits/cac3db973f9e40c9ddd9b4c16b657eef049acda4, but that should by all logical means have the same effect.
It may not be the best idea to have the `IUnknown` methods always forcefully set, maybe a check for `hasattr(cls, "QueryInterface"|"AddRef"|"Release")` would be a good addition.
So far however it hasn't caused any issues and is not a regression; COMObjects before have either not implemented the method as well, or erroneously returned 0 (success) without setting the pointer to anything.

caffeinepills (changes requested):
Thanks for those changes. I would probably squash the rest of your changes, unless Benjamin thinks its fine as is.

caffeinepills (approved):

benmoran56 (approved):

PR #972, opened 2023-10-13
Carry over the soundspace example to pyglet 2.0

Square789:
Now uses shapes to render its elements as opposed to direct legacy OpenGL calls.
This has a few negative effects with lines not showing up and the stippled lines being gone, but otherwise looks similar enough to the 1.5.x version.
I did not modify any logic setting the spatial attributes, so it should by all means behave the same and, from what i can tell, does behave as before.


benmoran56 (approved):

benmoran56:
Thanks for taking this up!

PR #973, opened 2023-10-16
Simplify debug_print and annotate non-codec utils

pushfoo:
### Changes

* Add forward-compatible import of Buffer in addition to existing union type
* Add typing to annotations to non-codec portions of `pyglet.util`
* Alter returned _debug functions  & function which returns them per discussion with Ben

I've left the annotations off of the Codec portions since @Square789 may be moving or changing those in the proposed sound branch.

### DIFF
--- a/pyglet/util.py
+++ b/pyglet/util.py
@@ -9,7 +9,7 @@ import pyglet
 from pyglet.customtypes import ByteString
 
 
-def asbytes(s):
+def asbytes(s: Union[str, ByteString]) -> bytes:
     if isinstance(s, bytes):
         return s
     elif isinstance(s, str):
--- a/pyglet/util.py
+++ b/pyglet/util.py
@@ -38,28 +38,42 @@ def _debug_print_dummy(arg: str) -> bool:
 
 
 def debug_print(pyglet_option_name: str = 'debug') -> Callable[[str], bool]:
-    """Get a debug print callable based on a pyglet option name.
+    """Get a debug printer controlled by the given ``pyglet.options`` key.
 
-    The debug print function returned should be used in an assert so
-    it can automatically be removed when running Python with the -O
-    flag. Both of the returned functions return True to ensure the
-    types for this work.
+    This allows repurposing ``assert`` to write cleaner, more efficient
+    debug output:
+
+    #. Debug printers fit into a one-line ``assert`` statements instead
+       of longer, slower key-lookup``if`` statements
+    #. Running Python with the -O flag makes pyglet run faster by
+       skipping all ``assert`` statements
 
     Usage example::
 
-        from pyglet.debug import debug_print
+        from pyglet.util import debug_print
+
+
         _debug_media = debug_print('debug_media')
 
+
         def some_func():
+            # Python will skip the line below when run with -O
             assert _debug_media('My debug statement')
+
+            # The rest of the function will run as normal
+            ...
+
+    For more information, please see `the Python command line
+    documentation <https://docs.python.org/3/using/cmdline.html#cmdoption-O>`_.
 
     Args:
         `pyglet_option_name` :
-            The name of a pyglet option to load the debug flag from.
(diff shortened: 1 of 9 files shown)


pyglet/util.py
```diff
@@ -16,25 +18,40 @@ def asbytes(s):
         return bytes(s)
 
 
-def asbytes_filename(s):
-    if isinstance(s, bytes):
+def asbytes_filename(s: Union[str, Buffer]) -> Optional[bytes]:
```
benmoran56, line 21: This function shouldn't return None, so it was broken.
That said, lets just delete this function entirely. It was only used in one specific place, but I just inlined the code to save on bloat.

pyglet/util.py
```diff
@@ -16,25 +18,40 @@ def asbytes(s):
...
+def _debug_print_real(arg: str) -> bool:
+    print(arg)
+    return True
+
+
+def _debug_print_dummy(arg: str) -> bool:
+    return True
+
 
+def debug_print(pyglet_option_name: str = 'debug') -> DebugPrintCallable:
```
benmoran56, line 48: I would just put the actual return type, not an alias. `Callable[[str], bool]`

pushfoo: I'll revert for now. If we need the type later as part of test / debug reworks, I'll add it back.

benmoran56 (changes requested):
In my opinion, I think the additions to `customtypes` is adding bloat, but causing a potential negative. 
Typing aliases for things like `Buffer` make sense because it is a real object that is included in recent Python releases. If a user sees a custom type, they then first have to try to cross reference what it is before understanding their typing error.

pyglet/util.py
```diff
@@ -16,25 +18,40 @@ def asbytes(s):
...
-def debug_print(enabled_or_option='debug'):
-    """Get a debug printer that is enabled based on a boolean input or a pyglet option.
-    The debug print function returned should be used in an assert. This way it can be
-    optimized out when running python with the -O flag.
+    The debug print function returned should be used in an assert so
+    it can automatically be removed when running Python with the -O
+    flag. Both of the returned functions return True to ensure the
+    types for this work.
 
     Usage example::
```
benmoran56, line 57: I just noticed this, but there is no such thing as `pyglet.debug`. :sweat:

pyglet/util.py
```diff
@@ -16,25 +18,40 @@ def asbytes(s):
...
+    return True
+
 
+def debug_print(pyglet_option_name: str = 'debug') -> DebugPrintCallable:
+    """Get a debug print callable based on a pyglet option name.
 
-def debug_print(enabled_or_option='debug'):
-    """Get a debug printer that is enabled based on a boolean input or a pyglet option.
-    The debug print function returned should be used in an assert. This way it can be
-    optimized out when running python with the -O flag.
```
benmoran56, line 37: The original description reads more clearly to me. Perhaps the reason why we are using an `assert` in the first place might need more explanation, as we are essentially hijacking the intended behavior here.
The fact that it always returns True is just a way to ensure that it never gives an assertion error. 

We use an assert here because:
1. It allows us to put print statements on one line, instead of adding `if _debug_enabled: ...` statements everywhere. 
2. When Python is run with -O (as when frozen), asserts are automatically optimized out. 

I can update this myself if you like.

pyglet/util.py
pushfoo: Rebased, removed, and force-pushed. :+1:

pyglet/util.py
pushfoo: Good catch, ty. I've fixed it for now, but we might want to rethink some this design in the future. More thoughts on that later.

pyglet/util.py
pushfoo: I've tried to include both the original summary line and your phrasing here. Feel free to make tweaks. :+1:

benmoran56 (approved):

PR #974, opened 2023-10-17
Fix slowdowns and unexpected errors in integration when running all at once

Square789:
- Removal of a test that set the audio driver to None without deleting it, causing later errors
- Some image tests were not cleaning up after themselves, creating two invisible windows that then stalled later tests

### DIFF
--- a/tests/integration/app/test_eventloop.py
+++ b/tests/integration/app/test_eventloop.py
@@ -23,7 +23,7 @@ def test_multiple_start_stop(performance):
     with performance.timer(30.):
         for _ in range(100):
             test_start_stop(performance)
-    
+
 
 def test_events():
     enter_mock = mock.MagicMock()
@@ -58,4 +58,3 @@ def test_sleep(performance):
         event_loop.run()
     assert not event_loop.is_running
     assert _sleep.returned.wait(1.)
-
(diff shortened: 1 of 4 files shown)


Square789:
I should probably also describe the issue instead of just submitting some code that deletes a test:
Currently, when running the integration test suite via `pytest tests/integration/` (on python 3.11.6 64bit, linux), the later tests in `test_driver.py` and `test_player.py` take an abnormally long time of around 3 minutes and then fail with either timeout errors, `'super' object has no attribute 'dispatch_event'` or time differences of `assert 10.0 < 5.0`.

The commit does not completely fix those tests. They now do not stall anymore, completing in a time of ~40 seconds.
However, slightly wrong times are still sometimes reported by audio players; though they look like `assert 5.0 <= 4.964020071428571` and can probably be considered more acceptable.

benmoran56 (approved):

benmoran56:
Thanks Square. 

Regarding the slightly off PulseAudio times,  those ones almost always fail under CI, but also (usually) fail on my local PC as well.  I think it started happening more frequently after I switched to Pipewire, which of course is using a PulseAudio compatibility layer.

Issue #964, opened 2023-10-08
win32_gdi_font tracebacks in python 3.12


ccanepa:
traceback while running examples\text\font_comparison.py under python 3.12, same problem with pyglet 1.5 maintenance

OS: windows 11 Pro Versión	10.0.22621 compilación 22621

pyglet 2 at
(py312_p) D:\dev\pyglet2\examples\text>git log
commit 237c56ce0855f9a52839268696254bde46172912 (HEAD -> master, origin/master, origin/HEAD)
Author: Benjamin 
Date:   Sun Oct 8 15:38:58 2023 +0900

    ComputeShaderProgram: expose MAX values directly on the class. 

```
Traceback (most recent call last):
  File "D:\dev\pyglet2\examples\text\font_comparison.py", line 128, in <module>
    window.on_text_motion(None)
  File "D:\dev\pyglet2\examples\text\font_comparison.py", line 91, in on_text_motion
    self.head = pyglet.text.Label(face, font_size=16, y=0, anchor_y='bottom')
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\dev\pyglet2\pyglet\text\__init__.py", line 444, in __init__
    super().__init__(doc, x, y, z, width, height, anchor_x, anchor_y, multiline, dpi, batch, group, rotation)
  File "D:\dev\pyglet2\pyglet\text\__init__.py", line 241, in __init__
    super().__init__(document, width, height, multiline, dpi, batch, group)
  File "D:\dev\pyglet2\pyglet\text\layout.py", line 924, in __init__
    self.document = document
    ^^^^^^^^^^^^^
  File "D:\dev\pyglet2\pyglet\text\layout.py", line 976, in document
    self._init_document()
  File "D:\dev\pyglet2\pyglet\text\layout.py", line 1448, in _init_document
    self._update()
  File "D:\dev\pyglet2\pyglet\text\layout.py", line 1362, in _update
    lines = self._get_lines()
            ^^^^^^^^^^^^^^^^^
  File "D:\dev\pyglet2\pyglet\text\layout.py", line 1339, in _get_lines
    glyphs = self._get_glyphs()
             ^^^^^^^^^^^^^^^^^^
  File "D:\dev\pyglet2\pyglet\text\layout.py", line 1496, in _get_glyphs
    glyphs.extend(font.get_glyphs(text[start:end]))
[...]

PR #977, opened 2023-10-21
Fixes for Python 3.12

caffeinepills:
Overwrite BYTE for older Python versions to the correct type
Adjust datatypes to ensure proper usage of BYTE.
Add check for localized font name for GDIPlus Font renderer.
Add debug message if font falls back to default font name.

Should resolve: https://github.com/pyglet/pyglet/issues/964

### DIFF
--- a/pyglet/font/win32.py
+++ b/pyglet/font/win32.py
@@ -532,6 +532,12 @@ class GDIPlusFont(Win32Font):
         self._name = name
 
         family = ctypes.c_void_p()
+
+        # GDI will add @ in front of a localized font for some Asian languages. However, GDI will also not find it
+        # based on that name (???). Here we remove it before checking font collections.
+        if name[0] == "@":
+            name = name[1:]
+
         name = ctypes.c_wchar_p(name)
 
         # Look in private collection first:
@@ -540,6 +546,9 @@ class GDIPlusFont(Win32Font):
 
         # Then in system collection:
         if not family:
+            if _debug_font:
+                print(f"Warning: Font '{name}' was not found. Defaulting to: {self._default_name}")
+
             gdiplus.GdipCreateFontFamilyFromName(name, None, ctypes.byref(family))
 
         # Nothing found, use default font.
--- a/pyglet/font/win32.py
+++ b/pyglet/font/win32.py
@@ -336,7 +336,7 @@ class GDIPlusGlyphRenderer(Win32GlyphRenderer):
             pass
 
     def _create_bitmap(self, width, height):
-        self._data = (ctypes.c_byte * (4 * width * height))()
+        self._data = (BYTE * (4 * width * height))()
         self._bitmap = ctypes.c_void_p()
         self._format = PixelFormat32bppARGB
         gdiplus.GdipCreateBitmapFromScan0(width, height, width * 4,
--- a/pyglet/image/codecs/wic.py
+++ b/pyglet/image/codecs/wic.py
@@ -601,7 +601,7 @@ class WICEncoder(ImageEncoder):
 
         frame.SetPixelFormat(byref(default_format))
 
-        data = (c_byte * size).from_buffer(bytearray(image_data))
+        data = (BYTE * size).from_buffer(bytearray(image_data))
 
         frame.WritePixels(image.height, pitch, size, data)
 
--- a/pyglet/input/win32/xinput.py
+++ b/pyglet/input/win32/xinput.py
@@ -113,8 +113,8 @@ ERROR_SUCCESS = 0
 class XINPUT_GAMEPAD(Structure):
     _fields_ = [
         ('wButtons', WORD),
-        ('bLeftTrigger', UBYTE),
-        ('bRightTrigger', UBYTE),
+        ('bLeftTrigger', BYTE),
+        ('bRightTrigger', BYTE),
         ('sThumbLX', SHORT),
         ('sThumbLY', SHORT),
         ('sThumbRX', SHORT),
--- a/pyglet/libs/win32/types.py
+++ b/pyglet/libs/win32/types.py
@@ -1,4 +1,5 @@
 import ctypes
+import sys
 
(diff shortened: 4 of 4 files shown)


ccanepa:
I think this patch is too big, and besides includes things unrelated to fixing the BYTE issue.

Restricting my comments to the BYTE issue

- the change in pyglet/font/win32.py
```
-        self._data = (ctypes.c_byte * (4 * width * height))()
+        self._data = (BYTE * (4 * width * height))()
```
by itself fixes issue[ #954](https://github.com/pyglet/pyglet/issues/964)  ;  why not to do a separate commit with only this and appropriate comments closing that issue?
With this change old pythons will see the same semantic as in older pyglet, so I will wait to see if something else breaks in 3.12 with this change before doing anything else.

- changes in pyglet/image/codecs/bmp.py . I would left the file as it was, except by a comment above the definition of BYTE telling that for python < 3.12 this BYTE is not the same as the one comming from ctypes, and to see issue in cpython tracker. 
Rationale: The code is correct for any python version, this code is in use in any platform, not only Windows, and less change is always better.

- changes in pyglet/libs/win32/types.py . I would not change this file, it pulls the rug under user code the same as ctypes change pulled the rug under pyglet, and is unclear if pyglet code would behave under this change.
 
- pyglet/input/win32/xinput.py . The old code is correct in all python versions, why change? 
- pyglet/image/codecs/wic.py . I'm too tired to review this. A goal should be to not depend on a monkeypatched BYTE. Anyway, it should be a separate commit?

caffeinepills:
> by itself fixes issue[ #954](https://github.com/pyglet/pyglet/issues/964) ; why not to do a separate commit with only this and appropriate comments closing that issue? 

I have separated the commit into a separate one for you.

> * changes in pyglet/image/codecs/bmp.py . I would left the file as it was, except by a comment above the definition of BYTE telling that for python < 3.12 this BYTE is not the same as the one comming from ctypes, and to see issue in cpython tracker.
>   Rationale: The code is correct for any python version, this code is in use in any platform, not only Windows, and less change is always better.

Although BMP is a Window specific format, this is true. If someone wants to decode BMP on another OS, it should be fine. I have removed this.

> * changes in pyglet/libs/win32/types.py . I would not change this file, it pulls the rug under user code the same as ctypes change pulled the rug under pyglet, and is unclear if pyglet code would behave under this change.

Defining `BYTE` in this file does not change `ctypes.wintypes.BYTE`.

> * pyglet/input/win32/xinput.py . The old code is correct in all python versions, why change?
`UBYTE` is not an actual type, the correct type is `BYTE`.

> * pyglet/image/codecs/wic.py . I'm too tired to review this. A goal should be to not depend on a monkeypatched BYTE. Anyway, it should be a separate commit?
`WritePixels` is the same issue disparity between `c_byte` and the required `BYTES` from `WritePixels`. I made my adjustments in a separate commit.

This PR is to accommodate 3.12 changes and changes to `BYTE`, not just your issue with the GDI font bitmap datatype.

ccanepa:
>>changes in pyglet/libs/win32/types.py . I would not change this file, it pulls the rug under user code the same as ctypes change pulled the rug under pyglet, and is unclear if pyglet code would behave under this change.

>Defining BYTE in this file does not change ctypes.wintypes.BYTE.

User code that imported BYTE by way of `from pyglet.libs.win32.types import *` will see a potentially breaking change in py < 3.12 (think user code writen similar to the one broken in #964 ), essentially backporting the problematic change in ctypes 3.12 to earlier pythons.

Actually, I would prefer monkeypatching ctypes.wintypes.BYTE on pyglet start-up: theres no risk of user code mixing ctypes.wintypes.BYTE and pyglet.libs.win32.types.BYTE

Okay, I have changed my mind: let the pyglet bindings using BYTE and POINTER(BYTE) be correct in all pythons, but do it by monkeypatching ctypes.wintypes.BYTE at pyglet start-up, because it ensures user code could not choose a 'wrong' BYTE 

What you think of that @caffeinepills ?

caffeinepills:
> > > changes in pyglet/libs/win32/types.py . I would not change this file, it pulls the rug under user code the same as ctypes change pulled the rug under pyglet, and is unclear if pyglet code would behave under this change.
> 
> > Defining BYTE in this file does not change ctypes.wintypes.BYTE.
> 
> User code that imported BYTE by way of `from pyglet.libs.win32.types import *` will see a potentially breaking change in py < 3.12 (think user code writen similar to the one broken in #964 ), essentially backporting the problematic change in ctypes 3.12 to earlier pythons.
> 
Personally, I think this is fine. If they were using it incorrectly in their own personal ctypes code (in the theoretical scenario they were using our `pyglet.libs.win32.types` definitions instead of `ctypes.wintypes`), they should probably be made aware `BYTE` is unsigned, not signed. We can make sure we include some disclaimer in the release notes.

> Actually, I would prefer monkeypatching ctypes.wintypes.BYTE on pyglet start-up: theres no risk of user code mixing ctypes.wintypes.BYTE and pyglet.libs.win32.types.BYTE

If we monkeypatch it, then we may also affect any other library being used with pyglet. Not sure that would be appropriate.

@benmoran56 What is your opinion on this matter?

benmoran56:
I agree with caffeinepills here.  End users shouldn't be relying on `pyglet.libs.win32.types`. It's undocumented, and only meant for use by pyglet internals. This would potentially affect other libraries that utilize pyglet, rather than end-user projects. A mention in the release notes should be enough I think. Something like:
> `libs.win32.types.BYTE` is now ALWAYS a `c_ubyte`, as per bug fixed in Python 3.12`

I also considered patching `ctypes.wintypes` directly. However, we do know that many people are mixing pyglet with other libraries, so that would potentially cause more issues for end users.

ccanepa:
Okay.
Please don't forget to backport to 1.5-maintenance, and do a release.
Thanks!

benmoran56 (approved):

PR #978, opened 2023-10-21
Add basic clipboard support.

caffeinepills:
These changes allows you to call on the window to either get the text from the clipboard or put text into it. (`get_clipboard_text` and `set_clipboard_text`)

This does not handle any other information such as images or files, only text.

Tested on:
Linux (Ubuntu 22.04) - VM
Mac OS (Monterey) - VM
Windows 10

Basic example provided.

Disclaimer:
I did notice some copy and paste issues with Wayland (other programs would fail to get the clipboard data and paste either nothing, or it would paste sometimes). Researching a bit, this seems to be a common issue with Wayland and is mostly a OS/framework issue that is being worked on.

I also used a VM, so that may have attributed to it. However, a restart did fix the issues I was having. X11 mode tested and worked perfectly there every time.

Feel free to give some feedback or let me know if this worked or didn't work for you.

### DIFF
--- a/examples/clipboard.py
+++ b/examples/clipboard.py
@@ -1,0 +1,18 @@
+import pyglet
+
+"""Simple example demonstrating how to get the text in the clipboard as well as putting text in the clipboard."""
+
+window = pyglet.window.Window()
+
+print("Existing Clipboard Text:", window.get_clipboard_text())
+
+@window.event
+def on_key_press(symbol, modifiers):
+    print("Clipboard changed to: Hello World!")
+
+    window.set_clipboard_text("Hello World!")
+
+    print("Clipboard Text:", window.get_clipboard_text())
+
+
+pyglet.app.run()
--- a/pyglet/libs/darwin/cocoapy/cocoalibs.py
+++ b/pyglet/libs/darwin/cocoapy/cocoalibs.py
@@ -218,6 +218,7 @@ NSApplicationDidUnhideNotification = c_void_p.in_dll(appkit, 'NSApplicationDidUn
 NSApplicationDidUpdateNotification = c_void_p.in_dll(appkit, 'NSApplicationDidUpdateNotification')
 NSPasteboardURLReadingFileURLsOnlyKey = c_void_p.in_dll(appkit, 'NSPasteboardURLReadingFileURLsOnlyKey')
 NSPasteboardTypeURL = c_void_p.in_dll(appkit, 'NSPasteboardTypeURL')
+NSPasteboardTypeString = c_void_p.in_dll(appkit, 'NSPasteboardTypeString')
 NSDragOperationGeneric = 4
 
 # /System/Library/Frameworks/AppKit.framework/Headers/NSEvent.h
--- a/pyglet/libs/win32/__init__.py
+++ b/pyglet/libs/win32/__init__.py
@@ -211,7 +211,18 @@ _user32.RegisterDeviceNotificationW.restype = HANDLE
 _user32.RegisterDeviceNotificationW.argtypes = [HANDLE, LPVOID, DWORD]
 _user32.UnregisterDeviceNotification.restype = BOOL
 _user32.UnregisterDeviceNotification.argtypes = [HANDLE]
-
+_user32.SetClipboardData.restype = HANDLE
+_user32.SetClipboardData.argtypes = [UINT, HANDLE]
+_user32.EmptyClipboard.restype = BOOL
+_user32.EmptyClipboard.argtypes = []
+_user32.OpenClipboard.restype = BOOL
+_user32.OpenClipboard.argtypes = [HWND]
+_user32.CloseClipboard.restype = BOOL
+_user32.CloseClipboard.argtypes = []
+_user32.GetClipboardData.restype = HANDLE
+_user32.GetClipboardData.argtypes = [UINT]
+_user32.SetClipboardData.restype = HANDLE
+_user32.SetClipboardData.argtypes = [UINT, HANDLE]
 
 # dwmapi
 _dwmapi.DwmIsCompositionEnabled.restype = c_int
--- a/pyglet/window/__init__.py
+++ b/pyglet/window/__init__.py
@@ -791,6 +791,25 @@ class BaseWindow(EventDispatcher, metaclass=_WindowMetaclass):
         """
         raise NotImplementedError()
 
+    def get_clipboard_text(self) -> str:
+        """Access the system clipboard and attempt to retrieve text.
+
+        :rtype: `str`
+        :return: A string from the clipboard. String will be empty if no text found.
+        """
+        raise NotImplementedError()
+
+    def set_clipboard_text(self, text: str):
+        """Access the system clipboard and set a text string as the clipboard data.
+
+        This will clear the existing clipboard.
+
+        :Parameters:
+            `text` : str
+                Text you want to place in the clipboard.
+        """
+        raise NotImplementedError()
+
     def minimize(self):
         """Minimize the window.
         """
--- a/pyglet/window/cocoa/__init__.py
+++ b/pyglet/window/cocoa/__init__.py
@@ -21,6 +21,7 @@ NSColor = cocoapy.ObjCClass('NSColor')
 NSEvent = cocoapy.ObjCClass('NSEvent')
 NSArray = cocoapy.ObjCClass('NSArray')
 NSImage = cocoapy.ObjCClass('NSImage')
+NSPasteboard = cocoapy.ObjCClass('NSPasteboard')
 
 quartz = cocoapy.quartz
 cf = cocoapy.cf
@@ -566,6 +567,33 @@ class CocoaWindow(BaseWindow):
 
         NSApp = NSApplication.sharedApplication()
         NSApp.setPresentationOptions_(options)
+
+    def set_clipboard_text(self, text: str):
+        with AutoReleasePool():
+            pasteboard = NSPasteboard.generalPasteboard()
+
+            pasteboard.clearContents()
+
+            array = NSArray.arrayWithObject_(cocoapy.NSPasteboardTypeString)
+            pasteboard.declareTypes_owner_(array, None)
+
+            text_nsstring = cocoapy.get_NSString(text)
+
+            pasteboard.setString_forType_(text_nsstring, cocoapy.NSPasteboardTypeString)
+
+    def get_clipboard_text(self) -> str:
+        text = ''
+        with AutoReleasePool():
+            pasteboard = NSPasteboard.generalPasteboard()
+
+            if pasteboard.types().containsObject_(cocoapy.NSPasteboardTypeString):
+                text_obj = pasteboard.stringForType_(cocoapy.NSPasteboardTypeString)
+                if text_obj:
+                    text = text_obj.UTF8String().decode('utf-8')
+
+        return text
+
+
 
 
 __all__ = ["CocoaWindow"]
--- a/pyglet/window/win32/__init__.py
+++ b/pyglet/window/win32/__init__.py
@@ -653,6 +653,43 @@ class Win32Window(BaseWindow):
 
         return icon
 
+    def set_clipboard_text(self, text: str):
+        valid = _user32.OpenClipboard(self._view_hwnd)
+        if not valid:
+            return
+
+        _user32.EmptyClipboard()
+
+        size = (len(text) + 1) * sizeof(WCHAR)  # UTF-16
+
+        cb_data = _kernel32.GlobalAlloc(GMEM_MOVEABLE, size)
+        locked_data = _kernel32.GlobalLock(cb_data)
+        memmove(locked_data, text, size)  # Trying to encode in utf-16 causes garbled text. Accepts str fine?
+        _kernel32.GlobalUnlock(cb_data)
+
+        _user32.SetClipboardData(CF_UNICODETEXT, cb_data)
+
+        _user32.CloseClipboard()
+
+    def get_clipboard_text(self) -> str:
+        text = ''
+
+        valid = _user32.OpenClipboard(self._view_hwnd)
+        if not valid:
+            print("Could not open clipboard")
+            return ''
+
+        cb_obj = _user32.GetClipboardData(CF_UNICODETEXT)
+        if cb_obj:
+            locked_data = _kernel32.GlobalLock(cb_obj)
+            if locked_data:
+                text = ctypes.wstring_at(locked_data)
+
+                _kernel32.GlobalUnlock(cb_obj)
+
+        _user32.CloseClipboard()
+        return text
+
     # Private util
 
     def _client_to_window_size(self, width, height):
--- a/pyglet/window/xlib/__init__.py
+++ b/pyglet/window/xlib/__init__.py
@@ -2,6 +2,7 @@ import unicodedata
 import urllib.parse
 from ctypes import *
 from functools import lru_cache
+from typing import Optional
 
 import pyglet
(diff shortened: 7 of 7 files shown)


Square789:
Tested it on X11, it works well during runtime, but unfortunately for me (which might be really setup-specific) the shutdown isn't too clean:

https://github.com/pyglet/pyglet/assets/46634729/472fca2a-c5dc-41f6-9aae-3c9e0e932b35

If you check out the very end of this video, you can see the entire taskbar freezing up. I honestly have no idea what part of KDE this is specifically, but i can notice similar behavior when copying very large images and am 99.9% sure it's related to the clipboard getting locked up, before something deep in there notices and hands over/takes back control.

This does not happen when running any other pyglet example. This also does not happen when not writing to the clipboard, running the clipboard example to trigger a read and then closing it by clicking the X.
This additionally does not happen when changing the example to exclusively read from the clipboard. (It never clears the cached `_clipboard_str` and never makes additional requests, as no `SelectionClear` event is ever received, but commenting that out and running a request each time yields the expected result and also does not hang on shutdown.)
Thus, the problem very likely has to do with the writing portion.

As you can see by the fragments in my clipboard i spent an hour trying to make sense of this, but with basically no understanding of X (i do not understand window properties at all) came up fruitless for now.
The only thing i noticed so far: `CurrentTime` is passed in as the `time` argument to `XSetSelectionOwner`: https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#acquiring_selection_ownership
> Clients attempting to acquire a selection must set the time value of the SetSelectionOwner request to the timestamp of the event triggering the acquisition attempt, not to CurrentTime . A zero-length append to a property is a way to obtain a timestamp for this purpose; the timestamp is in the corresponding PropertyNotify event.

caffeinepills:
I suspect this may be whatever clipboard manager you are using. I would try just copy and pasting between normal applications.

Square789:
> Also can you try to uncomment this line: `#xlib.XSync(self._x_display, False)` And see if it makes any difference?
Unfortunately it does not. The clipboard manager also offers no functionality to disable any kind of ownership process.
I'll try to do some kind of side-by-side comparison to another implementation in a few days again, i feel optimistic there might only be a tiny thing missing

caffeinepills:
Unfortunate the sync did not work. What version of KDE is this? Is it running wayland? I did hear some Wayland issues with copy and paste during my implementation.

I did some reading and it says you can disable Klipper in the system tray. If worse comes to worse, maybe try killing any process with clipboard in it? 😅 Just to see if anything improves.

Not really sure where the issue could be. Looking around at other similar implementation, the code is running very similar to GLFW and SDL 2.0.

Square789:
Should be more or less recent, 5.27.8. This is not wayland, but X11
I compared the code to [this example of setting a selection](https://github.com/python-xlib/python-xlib/blob/master/examples/put_selection.py), and while there's some abstractions i think i kinda get it now.
I could also spot some differences: The XSend block was not indented, and the if for comparing is querying the selection owner again instead of comparing the selection atom and window to itself.
Additionally, if [this article here](https://www.uninformativ.de/blog/postings/2017-04-02/0/POSTING-en.html) is correct, TARGETS must be included in the return to the TARGETS request.
I think in an hour it's either gonna work or i'll have given up

caffeinepills:
Sounds good, I did include TARGETS previously, and didn't notice a difference in functionality on my end so I removed it. It can easily be added back. If you can't get it figured out, let me know what distribution and such you are using, if you can send me a download link I can try to investigate further.

Square789:
> The only thing i noticed so far: CurrentTime is passed in as the time argument

Huh. This was the last thing i was going to try and it actually turned out to be the solution

caffeinepills:
> > The only thing i noticed so far: CurrentTime is passed in as the time argument
> 
> Huh. This was the last thing i was going to try and it actually turned out to be the solution. I did notice some other stuff along the way with the returned TARGETS array being reported 32 times as much as it actually was, but that just kinda worked, and so did not returning `TARGETS` in there at all apparently.
> 

Very strange because in all code I've seen, `XSetSelectionOwner` is set to `CurrentTime` (for both SDL and GLFW, and your example link earlier). The requestor time is however used in the `SelectionRequest` event.
 
Are you sure it doesn't hang on a call from `get_clipboard_text` as it's called after a key press? That function does submit an event and waits for one to return, which usually should be quick, but maybe not? Perhaps remove that line from the example on the key press.

I had some time and installed Fedora KDE Plasma on a VM just to mess with Klipper a bit, and try the branch, and I haven't run into any issues with it just yet.

Square789:
That is the weird thing: The python process always completely exits while the taskbar and related elements are frozen. I checked htop and even in the initial video you can see the terminal has returned. Unless something is forking, there is definitely nothing hanging behind.
The `CurrentTime` avoidance is recommended right here: https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#acquiring_selection_ownership and so far it was the only change that really helped eradicate the lag.
But alas, the problem still shows up around 1/15 exits, but that beats 1/1 at least.

Ultimately it's not that big of a deal, and even better if you can't replicate it.
What's really weird and unfortunate is that `set_clipboard_text` will not succeed if it's run as part of the Esc stroke that closes the window; on Windows the data would probably make it to the clipboard no problem in that call. Instead on X11 you just have to declare yourself the owner and wait for something else to approach you? (And in the closing situation it's just not enough time.) What's strange as well is that i can only ever provoke `SelectionClear` by trying to trigger two `set_clipboard_text` calls in very very short succession, or by launching the example twice. It's like no other program ever tries taking the selection for itself, even though i thought that was the point. Possibly there's different mechanisms in place that sidestep X11. Well, i am definitely too tired to continue this now.

I applied some fixes (hopefully that PR on a PR branch thing is going to work out) for what that's worth. For now i'm pretty much all out of ideas, sadly

caffeinepills:
Could you confirm that `echo $XDG_SESSION_TYPE` in terminal is indeed x11? I do have some repeatable issues but only with Wayland.  I will look at the PR in the morning for some more clues.

Square789:
> Could you confirm that `echo $XDG_SESSION_TYPE` in terminal is indeed x11? I do have some repeatable issues but only with Wayland. I will look at the PR in the morning for some more clues.

It is x11, yes. I also tried running this on another somewhat unmaintained computer running gnome 42.9 (x11 as well), where there was no hang with and without a [clipboard manager](https://extensions.gnome.org/extension/4422/gnome-clipboard/) installed. Unfortunately in there, very rarely the selection would be set to nothing/unretrievable by all other programs after quitting. That may be related to the shutdown-by-esc; it makes complete sense for it to disappear if pyglet takes ownership of the selection and then quits before anything can ask it for it.
I will try adding another `XCheckTypedWindowEvent` miniloop in `set_clipboard_text`, but this can be detrimental, as nothing is guaranteed to immediately ask for the selection once pyglet owns it. It probably just happens on desktop systems as those have a clipboard manager installed.

caffeinepills:
> > Could you confirm that `echo $XDG_SESSION_TYPE` in terminal is indeed x11? I do have some repeatable issues but only with Wayland. I will look at the PR in the morning for some more clues.
> 
> It is x11, yes. I also tried running this on another somewhat unmaintained computer running gnome 42.9 (x11 as well), where there was no hang with and without a [clipboard manager](https://extensions.gnome.org/extension/4422/gnome-clipboard/) installed. Unfortunately in there, very rarely the selection would be set to nothing/unretrievable by all other programs after quitting. That may be related to the shutdown-by-esc; it makes complete sense for it to disappear if pyglet takes ownership of the selection and then quits before anything can ask it for it. I will try adding another `XCheckTypedWindowEvent` miniloop in `set_clipboard_text`, but this can be detrimental, as nothing is guaranteed to immediately ask for the selection once pyglet owns it. It probably just happens on desktop systems as those have a clipboard manager installed.

That makes sense, I believe I have read that you need to publish the clipboard on app exit for some. (as specified here: https://www.freedesktop.org/wiki/ClipboardManager/) I haven't added that yet. 

I did some more testing with Wayland (which I am guessing also uses XWayland) and found some very weird results.

If I set the clipboard text, then pasting into a full screen/maximized app either doesn't work or if I hold down ctrl+v, the system gets extremely slow like it's processing the requests, the events come super delayed in the console, and eventually might let one or two pastes through. You can see the cursor of the application become solid like it can't keep up.

If I take the app I'm pasting into and put it into windowed then it pastes instantly. Back to full screen, sometimes it may work, sometimes it just goes back to being broke.

If I go back to windowed, fine.

If I take that same window, and visually cover the Pyglet window entirely, it breaks again.

https://github.com/pyglet/pyglet/assets/17804711/bf39217c-4635-44f9-bf93-742750d8324a

This doesn't occur in X11, only Wayland or XWayland (atleast on my end). Wondering if there is some connection with your issue as well? Such a headscratcher...

Square789:
To cross that off: My idea with "I will try adding another `XCheckTypedWindowEvent` miniloop in `set_clipboard_text`" did not end up solving anything, that's a dead end.

I can unfortunately not report any change simply by `pyglet.app.run(None)`, other than the example now starting up slower.
I can't seem to replicate anything relating to window minimization or obscuring as well, that seems to be a wayland quirk. Absolutely no idea how this might tie into the redraw schedule, really. I am not too familiar with this section of pyglet.

This is the first time i'm seeing the `CLIPBOARD_MANAGER` specification, it might be a really good candidate for implementing (possibly with a bool on the window class to toggle it and otherwise fallback?).
It does apparently not exist on gnome 42.9, but does on KDE via QT; that might play into me getting the hang on close.
The biggest question is where code for this would be placed: "If a client needs to exit while owning the CLIPBOARD selection, it should request the clipboard manager to take over the ownership of the clipboard"
Would this have to be stuffed into the window's `close` function while checking the length of `app.windows`? Doing it like that does sounds like a great source of problems, but there's no real other way as far as i'm aware.

Square789:
These are the problems i am experiencing right now:
|   | KDE Plasma |  Gnome  |
| ------------- | ------------- | ------------- |
| Hang when closing via Esc | Yes |   |
| Clipboard lost when closed via Esc | Yes, but recovers after<br />hang (To precursor of<br />the final "Hello World!") | Only when<br />`set_clipboard_text`<br />was not called before |
| `get_clipboard_text` incorrect when repeatedly<br />getting the selection after it was changed by<br />other applications and `set_clipboard_text` was<br />not called before | Yes | Yes |

Otherwise - from what i can tell - it  works absolutely as you'd expect it to.
These issues are somewhat specific to the example as it unintentionally uncovers a timing issue: Exiting very quickly after interacting with the selection causes issues, which is what happens by hitting Esc. Closing the window via mouse does neither cause a hang nor does it erase clipboard data. 
The fundamental problem remains in that there is no guaranteed clipboard in X you can write data to and after that just be done with it.
Possible solutions would be:
#### Do nothing, tell users to make sure their programs don't write to the clipboard on their way out.
- (+) Easiest option from here.
- (+) Doesn't seem like the worst idea, especially when considering that [even QT has quite a bullet list of X11 weirdness](https://doc.qt.io/qt-6/qclipboard.html#notes-for-x11-users).  
- (-) Introduces an undefinable period of time where clipboard usage is a bad idea.
#### Always enter an eventloop in `set_clipboard_text` that only exits after an appropiate `SelectionRequest` event for the is received and processed successfully.
- (+) Guaranteed point where the selection is stored: Once a request is successfully fulfilled.
- (-) Very exotic implementation needed: it's possible a flurry of `SelectionRequest`s for exclusively unsupported targets comes in, whose end can not be determined unless we'd start using time heuristics. ALTERNATIVELY: Assume that a clipboard manager will always ask for `UTF8_STRING`, which we will be able to deliver.
- (-) Runs on the assumption that the immediately incoming request is in fact a clipboard manager.
- (-) Will block pyglet in case nothing reacts to it taking over the selection.
#### Use the freedesktop standard involving the `CLIPBOARD_MANAGER` selection.
- (+) Guaranteed point where the selection is stored (provided that the clipboard manager works correctly): Once the  `SelectionNotification` for the conversion of `SAVE_TARGETS` is sent.
- (-) If `CLIPBOARD_MANAGER` is not present (Gnome), this would need to fall back to the first option anyways.

A warning note in the docs about not writing to the clipboard on exit (if you plan targeting Linux/X11) and amending the clipboard example honestly sounds like the least headache-inducing option right about now.

caffeinepills:
Thanks for the update, unfortunate that did not work. I appreciate the list. I will take a look at it further and will try to work on some workarounds for those specific issues when I have some time. 

Can you specify the distribution and build version you are using so I know what to test with? I tried the latest Fedora KDE Plasma, and didn't have any of the stated issues. However, there are multiple versions of KDE Plasma. Thanks

Square789:
Sure. It's Manjaro Linux with KDE Plasma 5.27.8, KDE Frameworks 5.110.0 and Qt 5.15.11, according to the "Info Center".
Using `Esc` to close the window is really the primary reason things are going wrong, otherwise the clipboard works just fine.

examples/clipboard.py
```diff
@@ -0,0 +1,18 @@
+import pyglet
+
+"""Simple example demonstrating how to get the text in the clipboard as well as putting text in the clipboard."""
```
pushfoo, line 3: I understand we are still at the proof of concept stage, but I think we could improve UX a lot with very little effort by:

1. Providing instructions at the top of the file
2. Separating copy and paste by assigning them to the standard platform hotkeys

This would make both testing and debugging easier by:
1. Separating copy and paste actions for increased control
2. Mapping clean to existing muscle memory

```suggestion
"""
How to get and set the text of the system clipboard.

This program does the following on launch:

1. Read the clipboard
2. Print the contents to the terminal

Afterward, you can use the normal copy and paste hotkeys for
your platform when the pyglet window has focus:

* Press CTRL-V to print text from the clipboard to the terminal
* Press CTRL-C to set the clipboard to 'Hello world!'
* If you are on Mac, use CMD instead of CTRL.

Output recording actions will be printed to the terminal.
"""
import pyglet
```

pushfoo (changes requested):
tl;dr: I think we should use CTRL-C and CTRL-V to improve UX

1. It doesn't add much complexity
2. It increases control when testing
3. It reduces cognitive load and surprise

We need the last two given how awful debugging the Linux copy and paste situation is.

I'd like to add on-screen instructions at some point as well. There's an ugly mockup in the suggested changes, but we could also wait until later.

pushfoo:
### tl;dr:
1. I haven't had much luck replicating copy and paste issues mentioned in this thread
2. I'm in favor of the free desktop standard + leaving some notes in the doc to cover the rest for anyone brave enough to try
3. Should I try X11 if I'll have to use a VM with LLVMpipe instead of real hardware?

### Testing

The only weirdness I've been able to find so far may have been input focus rather than actual copy and paste problems. This may have been due to Gnome extensions I'll describe further down.

I don't have a second GPU to test X11 in passthrough mode, but I could try to set up a VM with LLVMpipe if we need additional testing. The other option could be Xepyhr, but that might introduce complications without really solving anything.

### System details

I'm on Gnome 3.38.6 on Wayland on dual monitors. As far as I understand, this means pyglet has to be running on XWayland since the native Wayland support isn't finished.

```console
$ echo $XDG_SESSION_TYPE
wayland
```
The system is a stock Debian 11 configuration aside from a few desktop extensions (Pop's tiling via Pop Shell). The extensions are probably the source of the input focus issues I mentioned above.

To my knowledge, I'm not running any sort of clipboard manager.

shenjackyuanjie:
image

intel mac tested succeed

benmoran56 (approved):

PR #984, opened 2023-10-27
Always add GL objects to a context's `doomed*` lists.

Square789:
Any pyglet program that spins up threads that eventually end up running python code, be that directly, or by playing audio, is vulnerable to extremely rare segfaults provoked by the garbage collector calling the `__del__` method of `VertexArray`s or `ShaderProgram`s from foreign threads while the GL context of those objects is active.
This commit should fix this by unconditionally delaying the deletion of such objects into their context's `doomed` list.
Arcade appears to be using a strategy similar to this to navigate around the issue already: https://github.com/pythonarcade/arcade/issues/876, https://github.com/pythonarcade/arcade/blob/a5a0cd70e5482f0b23583226da8d0762119c373d/arcade/gl/context.py#L332

This could have been implemented differently by:
- Comparing `threading.current_thread()` to `threading.main_thread()` or somehow recording the GL context's active thread as well and comparing against that, but i decided against it. That check would be unnecessary work for the only gain of calling the deletion function immediately; delaying the object deletion by one frame should not cause any harm.
- Taking control of the garbage collector. Way too easy for this to go wrong.

Judging by the `and False` in `delete_buffer`, this problem may have been encountered and dealt with already some time ago. It's difficult to tell, as that dates back to the initial commit here: https://github.com/pyglet/pyglet/commit/2971856930543767fce6c73f29af9945c8bfa295

This could introduce more frequent leaking of OpenGL objects, as contexts do not delete resources when they are destroyed.
VAOs can easily be deleted in `destroy`, but for object spaces there'd need to be some kind of usage counter that causes the context to only delete the objects when it's the last one sharing it. Even that may be undesirable in case objects can somehow survive without a context, but i can't seem to find any info on that.
The leaking may not be too much of a problem, as most apps don't create an unbounded amount of windows/contexts in their lifetime, but it should be considered.

Edit: I additionally noticed that the classes `Framebuffer`, `Renderbuffer`, `Shader` and possibly the headless display still directly call `gl` functions in their `__del__` methods, so this commit does not secure against everything.

### DIFF
--- a/pyglet/gl/base.py
+++ b/pyglet/gl/base.py
@@ -347,25 +347,24 @@ class Context:
         return program
 
     def delete_texture(self, texture_id):
-        """Safely delete a Texture belonging to this context.
+        """Safely delete a Texture belonging to this context's object space.
 
-        Usually, the Texture is released immediately using
-        ``glDeleteTextures``, however if another context that does not share
-        this context's object space is currently active, the deletion will
-        be deferred until an appropriate context is activated.
+        This method only marks the given texture for deletion until any
+        context with the same object space becomes active again, where
+        it will then be deleted via ``glDeleteTextures``.
+
+        This makes it safe to call from anywhere, including other threads.
 
         :Parameters:
             `texture_id` : int
                 The OpenGL name of the Texture to delete.
 
         """
-        if self.object_space is gl.current_context.object_space:
-            gl.glDeleteTextures(1, gl.GLuint(texture_id))
-        else:
-            self.object_space.doomed_textures.append(texture_id)
+        self.object_space.doomed_textures.append(texture_id)
 
     def delete_buffer(self, buffer_id):
-        """Safely delete a Buffer object belonging to this context.
+        """Safely delete a Buffer object belonging to this context's object
+        space.
 
         This method behaves similarly to `delete_texture`, though for
         ``glDeleteBuffers`` instead of ``glDeleteTextures``.
@@ -376,30 +375,11 @@ class Context:
 
         .. versionadded:: 1.1
         """
-        if self.object_space is gl.current_context.object_space and False:
-            gl.glDeleteBuffers(1, gl.GLuint(buffer_id))
-        else:
-            self.object_space.doomed_buffers.append(buffer_id)
-
-    def delete_vao(self, vao_id):
-        """Safely delete a Vertex Array Object belonging to this context.
-
-        This method behaves similarly to `delete_texture`, though for
-        ``glDeleteVertexArrays`` instead of ``glDeleteTextures``.
-
-        :Parameters:
-            `vao_id` : int
-                The OpenGL name of the Vertex Array to delete.
-
-        .. versionadded:: 2.0
-        """
-        if gl.current_context is self:
-            gl.glDeleteVertexArrays(1, gl.GLuint(vao_id))
-        else:
-            self.doomed_vaos.append(vao_id)
+        self.object_space.doomed_buffers.append(buffer_id)
 
     def delete_shader_program(self, program_id):
-        """Safely delete a Shader Program belonging to this context.
+        """Safely delete a Shader Program belonging to this context's
+        object space.
 
         This method behaves similarly to `delete_texture`, though for
         ``glDeleteProgram`` instead of ``glDeleteTextures``.
@@ -410,10 +390,24 @@ class Context:
 
         .. versionadded:: 2.0
         """
-        if gl.current_context is self:
-            gl.glDeleteProgram(program_id)
-        else:
-            self.object_space.doomed_shader_programs.append(program_id)
+        self.object_space.doomed_shader_programs.append(program_id)
+
+    def delete_vao(self, vao_id):
+        """Safely delete a Vertex Array Object belonging to this context.
+
+        This method behaves similarly to `delete_texture`, though for
+        ``glDeleteVertexArrays`` instead of ``glDeleteTextures``.
+
+        As they cannot be shared, the given VAO will only be deleted
+        once this exact context is activated again.
+
+        :Parameters:
+            `vao_id` : int
+                The OpenGL name of the Vertex Array to delete.
+
+        .. versionadded:: 2.0
+        """
+        self.doomed_vaos.append(vao_id)
 
     def get_info(self):
         """Get the OpenGL information for this context.


caffeinepills:
While this change does seem necessary for some threading improvements, I feel like there may be some gotchas. It looks like these functions were guaranteeing that calling those functions meant either immediate deletion as long as you were in the main context or dooming it until it could be. If that's no longer the case, it could be a breaking change. For instance if you've never used a thread and you call `del` on a texture, then normally that's going to be immediately removed. However, that will no longer happen. If that's the case, I feel we may need to include a way to have an immediate deletion for some of these as well for people who aren't going to be using threading.

I am not 100% sure on the best practices though on how we should separate the behavior, or if it's even necessary.

Square789:
> I feel we may need to include a way to have an immediate deletion for some of these as well for people who aren't going to be using threading.

These changes behave similar whether multiple threads are present or not. Though for a program that only ever uses the main thread the `doomed` lists are a practically useless delay, yes.

> While this change does seem necessary for some threading improvements

They do not improve threading, they prevent bad OpenGL calls causing a segmentation fault. Currently pyglet can and does segfault with programs that use a number of threads larger than one, especially easily if throwaway functions such as `Label.draw` are used.

> For instance if you've never used a thread and you call `del` on a texture, then normally that's going to be immediately removed. However, that will no longer happen.

You are absolutely right with the textures and their immediate deletion being desirable. Textures can be extremely large and secretly deferring their deletion is definitely a bad idea, i did not quite think of that.
I never mentioned it, but i did miss a `delete` method on textures in the past. I think such a method on textures (and other objects) that simply assumes the user knows what they're doing and calls `glDeleteTextures` (or equivalents) would be a great addition.
While CPython does call `__del__` immediately once the user removes all references to a texture, making a client pick apart all locations where it may be referencing a texture and then hope none was overlooked and `__del__` will be intransparently called is unstable. A way of explicit deletion is far better than this implicit one.
It makes sense to keep `__del__` as an emergency deletion mechanism only in case more basic users simply don't care about managing their resources. Having a "punishment" of a one-frame memory hog for that is acceptable imo, but yeah, you're right, constitutes a behavior change. Reverting the unconditional dooming and adding `threading.current_thread() is threading.main_thread()` would effectively make deletion immedate and safe again.

`__del__` will remain necessary for the immediate user-facing things like shapes, sprites, windows, labels, textures and players. I think it may be a very good idea to remove as many finalizers as can be considered non-public and ensure the others are safe.

Square789:
The project i am using for testing this primarily is https://github.com/nwg-piotr/squarely. (The name may have been a tiny influential factor :^) )
Originally i used it to test the audio branch, but after noticing the segfaults and investigating, it became my goto for this.
It uses a thread to make highscore board requests, plays audio and draws a few labels each frame, making it trigger the problem with some amount of likelihood:
```
Fatal Python error: Segmentation fault

Current thread 0x00007fd52e3eb6c0 (most recent call first):
  File "/home/square/Code/python/pyglet/pyglet/gl/base.py", line 397 in delete_vao
  File "/home/square/Code/python/pyglet/pyglet/graphics/vertexarray.py", line 43 in __del__
  File "/home/square/Code/python/pyglet/pyglet/media/drivers/openal/interface.py", line 251 in unqueue_buffers
  File "/home/square/Code/python/pyglet/pyglet/media/drivers/openal/adaptation.py", line 205 in _handle_processed_buffers
  File "/home/square/Code/python/pyglet/pyglet/media/drivers/openal/adaptation.py", line 193 in _update_play_cursor
  File "/home/square/Code/python/pyglet/pyglet/media/drivers/openal/adaptation.py", line 231 in _get_write_size
  File "/home/square/Code/python/pyglet/pyglet/media/drivers/openal/adaptation.py", line 241 in refill_buffer
  File "/home/square/Code/python/pyglet/pyglet/media/mediathreads.py", line 56 in run
  File "/home/square/.pyenv/versions/3.8.17/lib/python3.8/threading.py", line 932 in _bootstrap_inner
  File "/home/square/.pyenv/versions/3.8.17/lib/python3.8/threading.py", line 890 in _bootstrap

Thread 0x00007fd53f3b9040 (most recent call first):
  File "/home/square/Code/python/pyglet/pyglet/gl/lib.py", line 80 in errcheck
  File "/home/square/Code/python/pyglet/pyglet/graphics/vertexbuffer.py", line 130 in __init__
  File "/home/square/Code/python/pyglet/pyglet/graphics/vertexbuffer.py", line 219 in __init__
  File "/home/square/Code/python/pyglet/pyglet/graphics/vertexdomain.py", line 125 in __init__
  File "/home/square/Code/python/pyglet/pyglet/graphics/vertexdomain.py", line 315 in __init__
  File "/home/square/Code/python/pyglet/pyglet/graphics/__init__.py", line 307 in get_domain
  File "/home/square/Code/python/pyglet/pyglet/graphics/shader.py", line 920 in vertex_list_indexed
  File "/home/square/Code/python/pyglet/pyglet/text/layout.py", line 563 in place
  File "/home/square/Code/python/pyglet/pyglet/text/layout.py", line 1850 in _create_vertex_lists
  File "/home/square/Code/python/pyglet/pyglet/text/layout.py", line 1369 in _update
  File "/home/square/Code/python/pyglet/pyglet/text/layout.py", line 1445 in _init_document
  File "/home/square/Code/python/pyglet/pyglet/text/layout.py", line 1481 in on_style_text
  File "/home/square/Code/python/pyglet/pyglet/event.py", line 401 in dispatch_event
  File "/home/square/Code/python/pyglet/pyglet/text/document.py", line 390 in insert_text
  File "/home/square/Code/python/pyglet/pyglet/text/formats/plaintext.py", line 11 in decode
  File "/home/square/Code/python/pyglet/pyglet/text/__init__.py", line 191 in decode_text
  File "/home/square/Code/python/pyglet/pyglet/text/__init__.py", line 446 in __init__
  File "/home/square/Games/raw_python/squarely/squarely/game_objects.py", line 1523 in score_label
  File "/home/square/Games/raw_python/squarely/squarely/game_objects.py", line 1493 in update_score_labels
  File "/home/square/Games/raw_python/squarely/squarely/tools.py", line 543 in update_scores
  File "main.py", line 423 in update
  File "/home/square/Code/python/pyglet/pyglet/clock.py", line 217 in call_scheduled_functions
  File "/home/square/Code/python/pyglet/pyglet/app/base.py", line 224 in idle
  File "/home/square/Code/python/pyglet/pyglet/app/base.py", line 163 in run
  File "/home/square/Code/python/pyglet/pyglet/app/__init__.py", line 76 in run
  File "main.py", line 448 in main
  File "main.py", line 496 in <module>
zsh: segmentation fault (core dumped)  PYTHONFAULTHANDLER=1 PYENV_VERSION="arcade_runner38" python3 main.py
```

Square789:
I think i'll come up with changes similar to the ones i outlined soon, but since they will be different from this PR, i'll do that in a new one.

PR #987, opened 2023-11-01
Add Width height setter

shenjackyuanjie:
Adding width and height `@setter` to `TextEntry`


shenjackyuanjie:
I think is good to go @benmoran56

benmoran56 (approved):

PR #988, opened 2023-11-02
More careful deletion of OpenGL objects

Square789:
The `Context.delete_*` family of methods now performs checks whether the calling thread is the main thread to avoid headscratch-inducing OpenGL errors or segmentation faults caused by `__del__`. There now is one such method for each implemented OpenGL object.

The deletion paths of all wrapped OpenGL objects (Texture, BufferObject, ShaderProgram and ComputeShaderProgram, Shader, Renderbuffer, Framebuffer and VertexArrayObject) now behave similarly:
- All those objects now reference their owning context by storing the current context at creation.
- `delete` methods have been newly added to `Texture` and the three shader-related classes.
- `delete` will immediately call the appropiate `glDelete*` function and set the object's `id`/`_id` to `None`, without any safety or exception tests.
- `__del__` will not act if the object's id is `None`.
- `__del__` will use the appropiate `Context.delete_*` method to delete the object otherwise.
- `__del__` is wrapped in a suppressing `except` catching two types of errors that tend to ocurr when running the `Context.delete_*` methods on interpreter shutdown:
  - `ImportError`s (caused by a gl method's `errcheck`)
  - `AttributeError`s (caused by module-level variables such as the `gl` or `threading` module being set to `None`).

### DIFF
--- a/pyglet/gl/base.py
+++ b/pyglet/gl/base.py
@@ -454,7 +454,7 @@ class Context:
             `shader_id` : int
                 The OpenGL name of the Shader to delete.
 
-        .. versionadded:: 2.0.XXX
+        .. versionadded:: 2.0.10
         """
         if self._safe_to_operate_on_object_space():
             gl.glDeleteShader(gl.GLuint(shader_id))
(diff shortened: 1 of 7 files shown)


benmoran56 (approved):

matanox:
Should this approach be confined to only deleting objects?

Square789:
@matanox I guess this is about #861 ? From what i can tell, the two problems don't really correlate: This issue was about a problem when the garbage collector invoked `__del__` from different threads that were not allowed to call GL functions.
The other issue seems to relate to multiple-window setups (threads aren't mentioned there so i hope they don't play a role) where a context is unexpectedly switched. This approach can't really be applied to that other problem.

<wrote some stuff going over your issue from #861, will post it over there>

PR #990, opened 2023-11-05
Make WidgetBase.enabled an instance property

pushfoo:
### Why

1. Allow subclasses to respond to changes in `enabled` status
2. Consistency: everything else has:
    1. a protected instance variable
    2. a property to expose the instance variable
3. No apparent conflicts with other UI efforts such as #930

### DIFF
--- a/pyglet/gui/widgets.py
+++ b/pyglet/gui/widgets.py
@@ -23,7 +23,8 @@ class WidgetBase(EventDispatcher):
     def _set_enabled(self, enabled: bool) -> None:
         """Internal hook for setting enabled.
 
-        Override this in subclasses.
+        Override this in subclasses to perform effects when a widget is
+        enabled or disabled.
         """
         pass
 
@@ -31,8 +32,14 @@ class WidgetBase(EventDispatcher):
     def enabled(self) -> bool:
         """Get/set whether this widget is enabled.
 
-        Override this property on subclasses to respond to changes in
-        status cleanly.
+        To react to changes in this value, override
+        :py:meth:`._set_enabled` on widgets. For example, you may want
+        to cue the user by:
+
+        * Play an animation
+        * Set a highlight color or effect around
+        * Play a sound
+
         """
         return self._enabled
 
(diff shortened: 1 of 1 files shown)


benmoran56:
I see where you're going with this, but `enabled` doesn't have an underlying function to tie into (like the x/y/position setters do). We could add one, like
```py
def _set_enabled(self, value):
    pass
```
But it seems like extra code for something that doesn't have an immediate need. I can't think of any use for a hook here.

pushfoo:
> that doesn't have an immediate need.

**tl;dr**: My understanding is that making this a setter:

1. makes it easier to add all kinds of [juice (satisfying UI polish and effects)](https://garden.bradwoods.io/notes/design/juice)
2. prevents some nasty typechecker issues 

### Better State Change Visibility

To my understanding, adding this property and the method you suggested makes it easier to highlight a UI element becoming enabled. This is broadly applicable to many goals, including:

* Tutorials which explain features
* Future interactive tests
* Psychology and neuroscience research

For example, a developer could be building a tutorial for their game. When the user unlocks a new UI feature due to completing part of the tutorial, [confetti could explode outward](https://codepen.io/marcobiedermann/pen/ExvvyLQ) to highlight it.

### Prevent type checker & data model issues

**tl;dr**: using only properties prevents combinatorial explosions of `Union`s

I agree this heading isn't an immediate concern, but it is useful for the future. We could move changes like this to the development branch if you'd like.

As I've mentioned in the context of batchables, we can free ourselves and our users from having to care about inheritance in the UI through `Protocol` types (added in 3.8). However, they add some requirements, such as choosing between the following:

1. Committing to data model consistency
3. Ugly union types to make up for things which can be attributes or properties

The second can grow increasingly ugly as more levels of inheritance get involved. We currently base our widgets around inheritance, which is why I am concerned about it for the future.

benmoran56:
If you want to add a dummy method that this can hook into, then I think we can merge it in. 
Perhaps something like
```py
def _on_enable(enabled: bool):
    pass
```
That will give users something they can override in subclasses.

benmoran56 (approved):

PR #996, opened 2023-11-13
Sort proxy in development branch

shenjackyuanjie:


shenjackyuanjie:
actually, thanks @KayHayen for pointing out of this

benmoran56 (approved):

kayhayen:
Actually, can I convince you to use the `lazy` package. It's easy for Nuitka to support that generically to produce optimal imports and attribute lookups due to a `.pyi` file. It also can be used to create code, as `huggingface_hub` does, that would be wonderful.

benmoran56:
Hi @kayhayen,

We can certainly make some changes to our module proxy if it helps with Nuitka. Unfortunately using the `lazy` package isn't an option as pyglet has no install dependencies. Can you let me know what is causing the difficulty on Nuitka's side?

kayhayen:
Well, the lazy loaders are not visible in static analysis. Nuitka would like to know that using `pyglet.input` attribute should have as a side effect, its import, and then use. So what ends up happening is that every package that does `def __getattr__` on the module level to avoid people having to `import pyglet.input` is making `ImportError` with any attempt to not include the kitchen sink.

I think huggingface_hub vendored `lazy` and generated `.py code, just check its package file, they then do

```__getattr__, __dir__, __all__ = _attach(__name__, submodules=[], submod_attrs=_SUBMOD_ATTRS)``` where the `_attach` is local code. I think `lazy` is not imported, only used to generate it. In your case, `submodules` would be a list.

Nuitka detect this kind of usage and generate what I internally call hard imports out of it. It is even possible to cut out the lazy loading for it, therefore removing its overhead entirely, although that may not matter to you. You also get a mechanism maintained by others, never a bad thing, since it will behave better potentially. I would like to see some standardization happening there, rather than every package doing their own thing. Like e.g. `pydantic` doing that manually.

With the other mode of operation, `.pyi` files become part of the code. There is something to dislike about it, but IDEs would love it so much, but I am sure lazy can also potentially generate .pyi files, but maybe not.

In your case, we cannot `import pyglet` and know which ones exist, except for checking its `__dict__` for instances and checking their internals. Every programmer group is going to do it more or less slightly better, worse, and to me mainly different.

**Comment by benmoran56** [comment]
Thanks for the information @kayhayen. I see what you mean regarding `lazy`. 

To clarify, are you saying the `if TYPE_CHECKING` block is adequate for Nuitka's needs? pyglet does already include that, but I suppose this is not enough.

I had a look at the `_attach` module, and it looks like a fairly straight forward approach using the "newer" module level `__getattr__` and `__dir__`. This would be fine for pyglet to switch to. As long as lazy loading works, the exact implementation is not important. I'll try this in another branch.

**Comment by benmoran56** [comment]
Doing a lazy import approach using `__getattr__` and importlib is easy enough, but this turned up another issue.
pyglet has lots of circular import issues when trying to run without any sort of lazy loading. It seems like we're relying on the more flexible nature of the ModuleProxy class, which is doing a little bit more than just lazy import.

**Comment by kayhayen** [comment]
It's not unusual for lazy imports to cause cyclic dependencies, they are also used as a solution for these, in some packages for a long time, I recall seeing some of them with comments, and actually once, I removed one, to find it breaks the package, may even have been mercurial a decade ago.

Anyway, Nuitka does not use the `TYPE_CHECKING` branch, and removes it. I was just contemplating what downsides the suggested `lazy` usage had, but turns out it's a non-topic, since it does it. 

For Nuitka, it's most relevant to have a way to import a module of yours and extract a config out of it that describes what you do to attribute lookup in terms of side-effects for importing. The `lazy` will allow for generic code to do it. Self made approaches may not expose the data (like yours), may be buggy, may give worse tracebacks, may ... I hope have lots of reasons to convince people to use `lazy` to solve this.

In my mind, module attribute lookup should be changed in the Python core language to have a submodule import as a side effect. From the looks of it, everybody hates to `import x.b` to be able to use `a.b` and hates being forced to pre-load modules that are often unlikely to be used and just made program startup slow. Check out how `os.path` works occasionally for the oldest art in Python of how to do submodules in Python that need no import. That said, some people do import `os.path` because they do not expect that kind of inconsistency.

Depending on if I can convince you to change it, I would love to mention your project as an example and success story in an upcoming posting of mine. I think more projects need to be aware to not waste time of writing lazy loaders by hand.

**Comment by benmoran56** [comment]
Imports in the core language in general is another topic for discussion :smile:

Unfortunately the `lazy` module doesn't completely solve our circular import issues. I've tried the full version of the code without success. I've also boiled it down to it's most basic, and it seems to be doing exactly what the lazy import example from the `importlib` module page shows. Essentially this:
```py
    __getattr__, __dir__, __all__ = _attach(__name__, submodules=submodules, submod_attrs=None)
```
can be distilled down to this. 
```py
    __all__ = ['app', 'canvas', 'clock', 'customtypes', 'event', 'font', 'gl', 'graphics', 'gui',
               'image', 'input', 'lib', 'math', 'media', 'model', 'resource', 'sprite', 'shapes', 'text', 'window']

    def __getattr__(name):
        if name in __all__:
            return importlib.import_module(f"pyglet.{name}")

    def __dir__():
        return __all__
```
This _should_ be perfectly adequate if pyglet didn't have circular import issues. Ideally, we want to see if we can solve those. (There is one main import that seems to be a cause of trouble. I'm going to look into that.)

By the way, in fairness to pyglet's `ModuleProxy` class, it was written over 15 years ago. It's not exactly a case of re-inventing the wheel, since it even predates the `importlib` module itself. :smile:

PR #1004, opened 2023-11-25
Label performance improvements and refactoring

caffeinepills:
This pull request refactors some things, and increases performance in certain cases.

1) Refactor `TextLayout` to include `x`, `y`, `z`, `anchor_x`, `anchor_y`, and `rotation` as initialization arguments. Considering the logic exists in this base class, it doesn't make sense to only be able to set these by creating a subclass, or setting properties after creation.

2) Visible is now a vertex attribute, instead of having to destroy the layout and rebuild it when unhiding, it can zero out the vertices. Increases performance.
```
Tested unhiding 2000 labels of "Test":
Previous:  765978 function calls (763978 primitive calls) in 1.009 seconds
New: 17206 function calls in 0.013 seconds
```

3) Simplified setting position across all layouts. `IncrementalTextLayout` no longer requires destroying the whole document and recreated it every position change. Increases performance.
```
Tested moving 2000 labels of "Test":
Previous: 28098 function calls in 0.029 seconds
New: 19206 function calls in 0.017 seconds

Tested moving 25 labels of 4,436 characters each:
Previous: 15777 function calls in 0.169 seconds
New: 6629 function calls in 0.064 seconds
```

4) Fix and re-order some calls on initialization and other functions to remove redundant calls. Some labels were creating and destroying themselves up to 3 times on creation. This speeds up label creation up to 3 times.
```
Creation time of 5 labels of 4436 characters:
Previous Time: 5.314368200022727 s
New Time: 1.7494019001023844 s

Creation Time of 2000 labels of "Test":
Previous Time: 2.537625700002536 s
New Time: 1.1062152000376955
```

5) Separated anchors from update so it does not require refactoring all of the glyphs again. Also cache the values required for getting anchors to prevent unnecessary rebuilding of lines. Increases performance.
```
Time to change anchor on 2000 labels of "Test":
Previous Time: 770979 function calls (768978 primitive calls) in 1.020 seconds
New Time: 21226 function calls in 0.019 seconds

Time to change anchor on 5 labels of 4436 characters:
Previous Time: 367282 function calls (367277 primitive calls) in 1.491 seconds
New Time:  1366 function calls in 0.008 seconds
```

6) Added `left`, `right`, `bottom`, `top` properties for easier hit testing. Just generally useful, especially if you plan to have some GUI elements.

Some bug fixes here and there. 
`ScrollableTextLayout` for instance was setting the position of 3 attributes directly, which was triggering 3 separate position changes.
Setting the view_x and view_y no longer triggers update to positions if not needed.

Tested on Python 3.10 and Windows 10 using text_input test and other files.

Pretty sure I didn't miss anything, but always looking for another pair of eyes or suggestions.

### DIFF
--- a/pyglet/text/layout.py
+++ b/pyglet/text/layout.py
@@ -1942,8 +1942,7 @@ class ScrollableTextLayout(TextLayout):
         if not self.document.text:
             return
 
-        area = (self._x + self._get_left_anchor(), self._y + self._get_bottom_anchor(self._get_lines()),
-                self._width, self._height)
+        area = (self.left, self.bottom, self._width, self._height)
 
         for group in self.group_cache.values():
             group.scissor_area = area
@@ -2113,8 +2112,7 @@ class IncrementalTextLayout(TextLayout, EventDispatcher):
                          wrap_lines)
 
     def _update_scissor_area(self):
-        area = (self._x + self._get_left_anchor(), self._y + self._get_bottom_anchor(self._get_lines()),
-                self._width, self._height)
+        area = (self.left, self.bottom, self._width, self._height)
 
         for group in self.group_cache.values():
             group.scissor_area = area
--- a/pyglet/text/layout.py
+++ b/pyglet/text/layout.py
@@ -871,7 +871,8 @@ class TextLayout:
     _multiline = False
     _visible = True
 
-    def __init__(self, document, width=None, height=None,
+    def __init__(self, document, x=0, y=0, z=0, width=None, height=None,
+                 anchor_x='left', anchor_y='bottom', rotation=0,
                  multiline=False, dpi=None, batch=None, group=None,
                  wrap_lines=True, init_document=True):
         """Create a text layout.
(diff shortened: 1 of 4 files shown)


shenjackyuanjie:
looks like there is no issue
good joob!

pyglet/text/layout.py
```diff
@@ -1102,30 +1126,31 @@ def position(self):
...
-                vertices[::3] = [x + dx for x in vertices[::3]]
-                vertices[1::3] = [y + dy for y in vertices[1::3]]
-                vertices[2::3] = [z + dz for z in vertices[2::3]]
-                vertex_list.position[:] = vertices
-            self._x = x
-            self._y = y
-            self._z = z
+            self._update_translation()
+
+    def _update_translation(self):
```
benmoran56, line 1139: Nice!

benmoran56 (approved):

shenjackyuanjie:
\o/

PR #1005, opened 2023-11-26
Add support for setting and getting arrays in a Uniform via wrapper.

caffeinepills:
Allow setting arrays in uniforms in GLSL: `uniform vec3 positions[9]`. 

When accessing the shader via `program["positions"]` it will return this:
```py
UniformArray(uniform=positions, data=[(128.0, 128.0, 128.0), (57.0, 123.0, 149.0), (36.0, 64.0, 76.0), (255.0, 255.0, 255.0), (255.0, 0.0, 255.0), (52.0, 85.0, 100.0), (151.0, 151.0, 151.0), (0.0, 0.0, 0.0), (178.0, 184.0, 216.0)])
```

Can be used as a list for the most part, you can iterate over it and do most of the things a list can do. It does not support sub-list assignment. (for example you can't set a `program["positions"][0][1] = 20.0` you would have to pass the full data type type via `program["positions"][0] = (12.0, 10.0, 20.0)`)

Set data in full via `program["positions"] = new_full_array`

Supports index getting: `program["positions"][0]` and setting: `program["positions"][0] = (10.0, 20.0, 30.0)` or:
```py
array = program["positions"]
array[0] =  (10.0, 20.0, 30.0)
```
Also supports slicing.

Add checks for multidimensional arrays, as they can't be properly inspected. Not sure if I should raise an exception for them, warn, or just ignore, but went exception route for now. Can change if needed.

Looking for input on a change, when shader debugging is on, to either ignore or warn about getting/setting missing uniforms. When debugging or adjusting shaders, it gets quite annoying to have the program put out an exception when I have temporarily commented out some uniforms or ones that were optimized out. Would be nice if we can have an option to bypass this or at the least change to a warning.

### DIFF
--- a/pyglet/graphics/shader.py
+++ b/pyglet/graphics/shader.py
@@ -1,11 +1,10 @@
+import warnings
 from ctypes import *
 from weakref import proxy
 
 import pyglet
-
 from pyglet.gl import *
 from pyglet.graphics.vertexbuffer import BufferObject
-
 
 _debug_gl_shaders = pyglet.options['debug_gl_shaders']
 
@@ -925,9 +924,14 @@ class ShaderProgram:
         try:
             uniform = self._uniforms[key]
         except KeyError as err:
-            raise ShaderException(f"A Uniform with the name `{key}` was not found.\n"
-                                  f"The spelling may be incorrect, or if not in use it "
-                                  f"may have been optimized out by the OpenGL driver.") from err
+            msg = (f"A Uniform with the name `{key}` was not found.\n"
+            f"The spelling may be incorrect, or if not in use it "
+            f"may have been optimized out by the OpenGL driver.")
+            if _debug_gl_shaders:
+                warnings.warn(msg)
+                return
+            else:
+                raise ShaderException(msg) from err
         try:
             uniform.set(value)
         except GLException as err:
@@ -937,9 +941,14 @@ class ShaderProgram:
         try:
             uniform = self._uniforms[item]
         except KeyError as err:
-            raise ShaderException(f"A Uniform with the name `{item}` was not found.\n"
-                                  f"The spelling may be incorrect, or if not in use it "
-                                  f"may have been optimized out by the OpenGL driver.") from err
+            msg = (f"A Uniform with the name `{item}` was not found.\n"
+            f"The spelling may be incorrect, or if not in use it "
+            f"may have been optimized out by the OpenGL driver.")
+            if _debug_gl_shaders:
+                warnings.warn(msg)
+                return
+            else:
+                raise ShaderException() from err
         try:
             return uniform.get()
         except GLException as err:
@@ -1140,9 +1149,14 @@ class ComputeShaderProgram:
         try:
             uniform = self._uniforms[key]
         except KeyError as err:
-            raise ShaderException(f"A Uniform with the name `{key}` was not found.\n"
-                                  f"The spelling may be incorrect, or if not in use it "
-                                  f"may have been optimized out by the OpenGL driver.") from err
+            msg = (f"A Uniform with the name `{key}` was not found.\n"
+            f"The spelling may be incorrect, or if not in use it "
+            f"may have been optimized out by the OpenGL driver.")
+            if _debug_gl_shaders:
+                warnings.warn(msg)
+                return
+            else:
+                raise ShaderException() from err
         try:
             uniform.set(value)
         except GLException as err:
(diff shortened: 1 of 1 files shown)


benmoran56 (dismissed):

pyglet/graphics/shader.py
```diff
@@ -236,31 +236,128 @@ def __repr__(self):
         return f"Attribute(name='{self.name}', location={self.location}, count={self.count})"
 
 
+class _UniformArray:
+    """Wrapper of the GLSL array data inside a Uniform.
+    Allows access to get and set items for a more Pythonic implementation.
+    Types with a length longer than 1 will be returned as tuples as an inner list would not support individual value
+    reassignment. Array data must either be set in full, or by indexing."""
+
+    _slots_ = ('uniform', 'gl_type', 'gl_getter', 'gl_setter', 'is_matrix', 'dsa', '_c_array', '_ptr')
```
benmoran56, line 245: Looks like the `_slots_` should be `__slots__`, and some of the name are missing the leading underscore.

pyglet/graphics/shader.py
```diff
@@ -236,31 +236,128 @@ def __repr__(self):
...
+
+    def __len__(self):
+        return self._uniform.size
+
+    def __delitem__(self, key):
+        raise ShaderException("Deleting items is not support for UniformArrays.")
+
+    def __getitem__(self, key):
+        """Return as a tuple.
+        Returning as a list may imply setting inner list elements will update values.
```
benmoran56, line 269: Just a nitpick, but this might be better off as a programmer comment.

pyglet/graphics/shader.py
```diff
@@ -236,31 +236,128 @@ def __repr__(self):
...
+    _slots_ = ('uniform', 'gl_type', 'gl_getter', 'gl_setter', 'is_matrix', 'dsa', '_c_array', '_ptr')
+    def __init__(self, uniform, gl_getter, gl_setter, gl_type, is_matrix, dsa):
+        self._uniform = uniform
+        self._gl_type = gl_type
+        self._gl_getter = gl_getter
+        self._gl_setter = gl_setter
+        self._is_matrix = is_matrix
+        self._dsa = dsa
+
+        if self._uniform.length > 1:
```
benmoran56, line 254: Assuming `_uniform.length` is always 1 for "normal" uniforms, could we skip the `if` check and always just do: ?
```py
self._c_array = (gl_type * self._uniform.length * self._uniform.size)()
```
The result should be the same, and we can avoid the if/else cost.

pyglet/graphics/shader.py
```diff
@@ -236,31 +236,128 @@ def __repr__(self):
...
+            else:
+                return tuple([data for data in sliced_data])
+
+        value = self._c_array[key]
+        return tuple(value) if self._uniform.length > 1 else value
+
+    def __setitem__(self, key, value):
+        if isinstance(key, slice):
+            start, stop, step = key.start, key.stop, key.step
+            self._c_array[start:stop:step] = value
```
benmoran56, line 285: I could be wrong, but I _think_ we can just pass the slice in directly (instead of splitting into start/stop/step). IE:
```py
            self._c_array[key] = value
```

benmoran56:
This looks like a reasonable approach.  I also prefer the exception instead of warning, since it might cause confusion if the warning is missed.

I made some code comments, but the design overall looks good to me.

pyglet/graphics/shader.py
caffeinepills: Good point.

pyglet/graphics/shader.py
caffeinepills: Unfortunately not, by doing `gl_type * self._uniform.length * self._uniform.size` it will create an array of an array even if length is 1. Which is fine for > 1, but if you want an array of floats for instance, then all of your floats are going to be sublists or tuples with only 1 value, which isn't very user friendly if we want to make it behave more like a list.

pyglet/graphics/shader.py
caffeinepills: I just tested, this seems to be the case. Nice!

caffeinepills:
Made the changes you outlined with the exception of making everything gl_type * length * size as 1 length values would end up becoming arrays of arrays, which would mean single values would return as tuples and setting data they would have to set fully as [(1,), (2,), (3,)] or [[1], [2], [3]] which is not very user friendly.

Added warnings if debugging is on instead of exceptions for uniform setting and getting, still keeps the exceptions normally. In the future I can expand this to attributes as well, but not sure of the ramifications of bypassing that, at the moment.

pyglet/graphics/shader.py
benmoran56: Ah, you're right. I forgot about that ctypes multiplication behavior.

benmoran56 (approved):

PR #1008, opened 2023-12-04
Improve VBO binding.

caffeinepills:
When tracing I noticed that the VBO's we use for binding were getting binded every draw call.

```py
glBindVertexArray
glBindBuffer
glBindBuffer
glBindBuffer
glBindBuffer
glBindBuffer
glBindBuffer
glDrawElements
```

Binding is normally done only when uploading data from what I read. Looking further this is done because we are grouping changes all at once in our data before submitting it to the GPU. The problem is that we tie this data upload to `bind` which was binding the VBO regardless of if it was 'dirty'.  Maybe there is a better way to do this, but I have just created a dirty set which can be iterated only when an attribute changes data, rather than every time.

Also, set the __name__ on GL functions so they don't all show up as <WinFunctionType object at X> during trace, so it's readable.

Changing this with 20,000 sprites gained me ~20% more FPS. 450-560

### DIFF
--- a/pyglet/gl/lib.py
+++ b/pyglet/gl/lib.py
@@ -83,6 +83,7 @@ def decorate_function(func, name):
     if _debug_gl:
         if name not in ('glGetError',) and name[:3] not in ('glX', 'agl', 'wgl'):
             func.errcheck = errcheck
+            func.__name__ = name
 
 
 link_AGL = None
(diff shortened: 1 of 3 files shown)


benmoran56:
Also, how about the AttributeBufferObject itself just having a `self._dirty` attribute as a boolean?
We could add an `if self._dirty` check at the top of the `bind` method, and have an early out.

benmoran56 (approved):

PR #1011, opened 2023-12-08
Fix bad behavior when clearing `XAudio2AudioPlayer`

Square789:
These commits will ensure that - when `XAudio2AudioPlayer`s get a new voice from the driver by getting cleared, they'll also set all player properties on it. Most noticeable is a discrepancy in volume, easily shown through
```py
import pyglet

player.volume = 0.1
player.loop = True
player.queue(pyglet.media.synthesis.Sine(0.4))
player.play()

pyglet.app.run()
```
where the volume snaps to the default of 1.0 for every other loop.

3d calculations are applied immediately to prevent very short bursts from a misplaced and not updated voice (Tested only via the soundspace example)

### DIFF
--- a/pyglet/media/drivers/xaudio2/adaptation.py
+++ b/pyglet/media/drivers/xaudio2/adaptation.py
@@ -183,6 +183,8 @@ class XAudio2AudioPlayer(AbstractAudioPlayer):
             self._xa2_source_voice.cone_orientation = _convert_coordinates(self.player.cone_orientation)
             self._xa2_source_voice.cone_outside_volume = self.player.cone_outer_gain
             self._set_cone_angles()
+            self.driver._xa2_driver.apply3d(self._xa2_source_voice, 2)
+            self.driver._xa2_driver._xaudio2.CommitChanges(2)
 
     def on_buffer_end(self, buffer_context_ptr: int) -> None:
         # Called from the XAudio2 thread.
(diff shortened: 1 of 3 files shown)


caffeinepills:
I think this is still incorrect.

When returning a voice, it should just remove itself from the emitters and reset it's volume, frequency, and position. There should be no need to calculate 3D, apply 3D, or commit changes (this is also assuming 3D is being used).

Square789:
> I think this is still incorrect.
> 
> When returning a voice, it should just remove itself from the emitters and reset it's volume, frequency, and emitter settings. It's no longer being used, there should be no need to calculate 3D, apply 3D, or commit changes (this is also assuming 3D is being used). This also needs to apply when calling .delete as well.

Pooled voices are in a completely unclear state that can not be assumed to correspond to the player's needs. They might be positioned anywhere with any orientation and have any volume, so will need to get adjusted when retrieved. It's no use to reset the voice's attributes to anything default-ish: what if the next player receiving it is playing at volume 0.8 with an orientation of 142.4092°?

If a voice is no longer being used, it is removed from `_emitting_voices`, which means it won't be picked up by any runs of `_calculate_3d_sources` anymore, so it won't eat up resources in that regard. It gets stuck in virtual audio space whereever it last was, does not produce any more output and repositioned immediately again when next unpooled.

PR #1017, opened 2023-12-15
Update doc related to PulseAudio bug

pushfoo:
### Changes

Soften PulseAudio bug phrasing to be specific about pyglet < 2.0.10 since @Square789 seems to have fixed the bug in question.


Square789:
The issue is not yet fixed in 2.0.10; I think it's enough to change mentions of the version to 2.0.11, assuming the PR makes its way into there successfully.

benmoran56 (approved):

PR #1018, opened 2023-12-20
Attributebufferwork

benmoran56:
This PR simplifies the AttributeBufferObject by using the actual Attribute c_type for it's local array.


caffeinepills (approved):
Looks good to me. Much easier to read as well.

Issue #1026, opened 2023-12-23
Changing pyglet.shape.Arc to triangles similar to Line to fix drawing issue


kingadami:
In my quest to learning more about pyglet I have been porting over a simple tic tac toe game I originally wrote in pygame.  It doesn't use any images and creates all of its shapes by "drawing" them.  I was using the Arc class to create my 'O' and noticed artifacts when you started to change the glLineWidth value to increase the width of the arc as shown in the following picture as 'notches' in the arc: ![ArcArtifact](https://github.com/pyglet/pyglet/assets/1959762/fba9725f-9717-4715-aeb5-446c49993331)

I was inspired by this post from [stackoverflow](https://stackoverflow.com/questions/3484260/opengl-line-width) on how to compute proper geometry to follow a line.  I have not implemented it as a shader like the poster did because that is still beyond my knowladge, but I have started to implement it when creating the vertices for the arc.  Currently my implementation gives the following results:
![ArcNoArtifact](https://github.com/pyglet/pyglet/assets/1959762/2c4bbdca-1751-4feb-a8e8-61dea3495904)

You can specify a width to the Arc like you can for a line.  I still have a few more corner cases, cleanup, and hopefully some simplifications to make, but I am planning on doing a pull request shortly.  The test program I am using is:

```
#!/usr/bin/env python3

import pyglet
import math

class Board(pyglet.window.Window):
    def __init__(self):
        super().__init__(caption="Test")
        # Required to set to change the width of an arc
        pyglet.gl.glLineWidth(20)

        self.batch = pyglet.graphics.Batch()
        # Old arc without a width parameter
        self.a1 = pyglet.shapes.Arc(self.width / 2, self.height / 2,min(self.width / 4, self.height / 4), segments=60, color=(0,0,255), closed=True, batch=self.batch)

        # New arc with a width parameter
        #self.a1 = pyglet.shapes.Arc(self.width / 2, self.height / 2,min(self.width / 4, self.height / 4), width=10, segments=60, color=(0,0,255), closed=True, batch=self.batch)

    def on_draw(self):
        self.clear()
        self.batch.draw()

if __name__ == "__main__":
    board = Board()
    pyglet.app.run()
```

The current repo I am using is [here](https://github.com/kingadami/pyglet/tree/shape-improvements).

benmoran56:
Thanks for opening the issue.  That would indeed be a welcome addition. 
[...]

PR #1028, opened 2023-12-26
Arc Shape improvements

kingadami:
This adds full support for width to the pyglet.shape.Arc class. Closes issue #1026

### DIFF
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -478,8 +478,8 @@ class ShapeBase(ABC):
 class Arc(ShapeBase):
     #_draw_mode = GL_LINES
 
-    def __init__(self, x, y, radius, segments=None, width=1, angle=math.tau, start_angle=0,
-                 closed=False, color=(255, 255, 255, 255), batch=None, group=None):
+    def __init__(self, x, y, radius, segments=None, angle=math.tau, start_angle=0,
+                 closed=False, width=1, color=(255, 255, 255, 255), batch=None, group=None):
         """Create an Arc.
 
         The Arc's anchor point (x, y) defaults to its center.
@@ -496,8 +496,6 @@ class Arc(ShapeBase):
                 the arc should be made from. If not specified it will be
                 automatically calculated using the formula:
                 `max(14, int(radius / 1.25))`.
-            `width` : float
-                The desired width of the line used for the arc.
             `angle` : float
                 The angle of the arc, in radians. Defaults to tau (pi * 2),
                 which is a full circle.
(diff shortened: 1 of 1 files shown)


pyglet/shapes.py
```diff
@@ -49,6 +49,7 @@ def on_draw():
 from abc import ABC, abstractmethod
 
 import pyglet
+import time
```
benmoran56, line 52: Unused import left over from testing?

pyglet/shapes.py
```diff
@@ -555,15 +561,109 @@ def _get_vertices(self):
...
+                v3 = (p1[0] - miter1ScaledP[0], p1[1] - miter1ScaledP[1])
+                v4 = (p2[0] + miter2ScaledP[0], p2[1] + miter2ScaledP[1])
+                v5 = (p2[0] - miter2ScaledP[0], p2[1] - miter2ScaledP[1])
+                v6 = (p1[0] - miter1ScaledP[0], p1[1] - miter1ScaledP[1])
+
+                # Setup for next run of function
+                prev_miter = v_miter2
+                prev_scale = scale2
+
+                return (v1[0],v1[1],v2[0],v2[1],v3[0],v3[1],v4[0],v4[1],v5[0],v5[1],v6[0],v6[1])
```
benmoran56, line 636: Could you add spaces between the items to keep the linters happy?

benmoran56:
I made a few inline comments, but overall looks good! 

One thing I was thinking about is the usage of `width`.  I ran into a case with the `Box` shape (made up of four lines), which already uses `width` to specify the overall width of the object. I instead opted to use `thickness` for the line widths. What do you think about using that here as well, for consistency. Your call on that.

pyglet/shapes.py
kingadami: Yes it was and is now removed

kingadami:
> How about `pyglet.shape.BezierCurve`? It's similar to the Arc.

I am definitely planning on doing that as well.  Since this was my first pull request I was wanting to keep it fairly contained and small so I can get feedback on style and preference.  If this gets accepted then I plan on changing BezierCurve to use this method as well.  I will probably move the _get_segments function outside of the Arc class since the code will be basically the same.

I was also thinking about either extending the `pyglet.shape.Line` to allow multiple points or adding something like a `pyglet.shape.PolyLine` and then re-writing the `pyglet.shape.Line` class as a specialization of the `pyglet.shape.PolyLine` class or something like that.  It would use the same _get_segment function as well, assuming people are open to the idea.

pyglet/shapes.py
kingadami: Done

benmoran56 (approved):

benmoran56:
That's a solid idea. The `_get_segment` code can easily be a function at the top of the file, and just take in prev_miter/scale as arguments.

PR #1030, opened 2023-12-30
Implementation of segments for pyglet.shapes.BezierCurve

kingadami:
Created pull request to allow comments and suggestions.


benmoran56 (approved):

benmoran56:
Looks good! There are some small linter complaints, but i'll take care of those locally. after merging.

PR #1039, opened 2024-01-19
Mathematical operation for Quaternion

zhengxyz123:
As the title says, I add more contents to Quaternion that can do some mathematical operation.

Besides, I think the two methods `from_mat3` and `from_mat4` are too difficult to implement. How about change them to `from_rotation`?

### DIFF
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -946,12 +946,17 @@ class Quaternion(tuple):
     def z(self) -> float:
         return self[3]
 
-    def conjugate(self) -> Quaternion:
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

benmoran56 (approved):

benmoran56:
Thanks for the pull request.

Sure, `from_rotation` is useful to add. I didn't have time to implement these things yet. 
We can leave `from_mat3/4` alone. If someone finds it useful, we can implement it later.

PR #1054, opened 2024-02-25
Add multiline shape

Burnberry:
General line shape consisting of multiple line segments by giving multiple point coordinates.

- Reuses _get_segement (also used in BezierCurve)
- Doesn't have contains implemented

Example:
image

### DIFF
--- a/doc/modules/shapes.rst
+++ b/doc/modules/shapes.rst
@@ -111,3 +111,9 @@ pyglet.shapes
 
 .. autoclass:: Polygon
   :show-inheritance:
+
+
+.. autoclass:: MultiLine
+  :show-inheritance:
+
+  .. autoattribute:: thickness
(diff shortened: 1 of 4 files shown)


pushfoo (changes requested):
At first glance, this seems to look good. However, could you please add this shape to unit tests?

If I remember correctly, most standard shape behavior should be covered by adding an entry here:
https://github.com/pyglet/pyglet/blob/4c4336c6431e9f2ddc94a4a04e6630194404e665/tests/unit/shapes/test_shapes.py#L8-L25

Burnberry:
Done and all tests passed

benmoran56 (approved):

benmoran56:
Also, since this is a standalone addition, you could make another PR targeting the master branch. We can get it in the next point release sooner. The development branch is for the upcoming pyglet 2.1.

PR #1075, opened 2024-03-17
Added a None check to ShapeBase.delete

James-Bonner:
ShapeBase.delete was not checking if self._vertex_list was not None before calling self._vertex_list.delete(). This could occasionally result in AttributeError: 'NoneType' object has no attribute 'delete'


benmoran56 (approved):

benmoran56 (approved):

benmoran56:
Thanks for the PR!

PR #1076, opened 2024-03-18
Re-raise error for attributes during shader creation to a little better description.

### LINKED ISSUE #1056: Re-raise attribute length exceptions with the name of the attribute being set
### What

Make `VertextList.set_attribute_data` and any other similar attribute-setting methods raise a clearer `IndexError` that says which attribute is being

### Why

It would be faster for developers if they knew which attribute had the wrong length. Traceback like this is inscrutable:
```
  File "/home/user/src/pyglet/.scratch/regedgs.py", line 99, in <module>
    program.vertex_list_indexed(8, GL_TRIANGLES, batch=batch,
  File "/home/user/src/pyglet/pyglet/graphics/shader.py", line 928, in vertex_list_indexed
    vlist.set_attribute_data(name, array)
  File "/home/user/src/pyglet/pyglet/graphics/vertexdomain.py", line 295, in set_attribute_data
    attribute.set_region(attribute.buffer, self.start, self.count, data)
  File "/home/user/src/pyglet/pyglet/graphics/shader.py", line 240, in set_region
    data = (self.c_type * array_count)(*data)
IndexError: invalid index
```

### How

1. Catch `IndexError`
2. Raise a new one with a better message which includes the attribute name

caffeinepills:
Should resolve #1056 sufficiently.

This will re-raise the error from passing an invalid sized list of data with a newly created list via `program.vertex_list_indexed` and `program.vertex_list`:

From:
 `ValueError: Can only assign sequence of same size`
to:
`ValueError: Invalid data size for 'position'. Expected 12, got 13.`

Did benchmarks and could not find any big difference in speed. With 6000 `set_attribute_data` calls, time with and without a try: block was 12 ms and own time was 9 ms with a 1-2ms variance. Tried and assertion was measurably slower. Although it can be optimized out, I wasn't sure if it was worth it, as it requires checking the size every call without optimizing.

When assigning invalid data from `vertex_list.position[:]` this will produce a `ValueError: Can only assign sequence of same size`. I think this should be sufficient clue when someone gives the wrong sized data when setting attribute data manually. With the initialization you don't know which attribute is erroring, but when you are setting the attribute directly, you know. I don't think we need to re-raise for that, we should let the assigning after creation be as fast as possible.


benmoran56 (approved):

benmoran56:
I agree with this approach. "No-cost" exception handling was implemented in Python 3.11, so `try` statements shouldn't cause much (any?) impact.

Issue #942, opened 2023-09-03
Feature Request: Provide IDE integration workarounds for ModuleProxy


pushfoo:
### Feature request

Explore ways of enabling auto-complete despite the use of `ModuleProxy` for lazy loading. This is a low-priority edge case, but it would be a quality of life improvement when using `inspect` in the REPL.

### The Problem

tl;dr `ModuleProxy` instances interfere with autocomplete in PyCharm or the IDE

Example:
1. Open PyCharm's REPL
2. `from pyglet import shapes`
3. Attempt to access members of `shapes`
4. No autocomplete occurs

### Potential Solutions to Explore

- [x] Ship manually generated [`.pyi` files](https://peps.python.org/pep-0484/#stub-files)
- [ ] Ship auto-generated [`.pyi` files](https://peps.python.org/pep-0484/#stub-files)

pushfoo:
This also seems to affect more regular usage such as auto-complete for key constants.

caffeinepills:
I'm for including a manually generated stub file for the lazy loading. I use one and it does indeed do autocompletion with PyCharm.

The only caveat I can see in the future is if anyone were to do any sort of enforced loading type checking. Since this makes my  IDE (PyCharm) mark the `ModuleProxy`'s as an incorrect type.

caffeinepills:
Stub file merged into master.

PR #1077, opened 2024-03-18
Stub file for package to workaround IDE inspections

caffeinepills:
Due to our use of lazy loading, it brings up a few issues with some IDE's, mostly PyCharm, which is unfortunately behind in a lot of modern typing/inspection features. VSCode seems to handle all scenarios properly without a stub file. I'm not sure why, but it most likely takes the `TYPE_CHECKING` block that's in the `__init__.py` for proper typing. While PyCharm seems to give the stub file priority over it's own inspections/typing.

For example, while PyCharm can handle `pyglet.app.` to get completion, it can't handle `from pyglet import app` as seen here:
 
image

Furthermore when you want to use the "Jump to" feature (ctrl click) to go to the module file, it will just jump to variable where `_ModuleProxy` was defined. Even with `import pyglet; pyglet.app` and `from pyglet import app` trying to jump to `app` will just take you to the `pyglet.__init__.py`.

With the stubfile, I no longer get these two issues. VSCode still seems to behave properly as well.

PyCharm:
image

The only caveat is I have to add `# type: ignore` on the module definitions to make it ignore errors otherwise it will produce an inspection error such as: `Expected type 'sprite.py', got '_ModuleProxy' instead`. MyPy seems to not complain about it without the `# type: ignore`, and neither does pylance/pyright. So it seems like it's strictly a PyCharm inspection problem when it comes to types, however this could also affect smaller IDE's. Those are the only two I've tested. 

Due to the popularity of PyCharm, I think this may be worth to add just to improve user usage, even if it's not technically our issue. 

Feel free to add onto this or deny it, but I think it's the only solution to https://github.com/pyglet/pyglet/issues/942 .

Just my perspective, but in the couple years I've been using PyCharm, I can't think of a time they've improved inspections or typing. (I've reported quite a few that sit open, marked as reproduced, to this day) So this might be the better route than just waiting for PyCharm to update while some users suffer.

We can always remove it in the future when most IDE's catch up to eachother.

### DIFF
--- a/pyglet/__init__.py
+++ b/pyglet/__init__.py
@@ -6,7 +6,7 @@ More information is available at http://www.pyglet.org
 import os
 import sys
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Dict
 
 #: The release version
 version = '2.0.14'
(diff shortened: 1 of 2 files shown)


benmoran56 (approved):

benmoran56 (approved):

benmoran56:
This is a reasonable approach. Hopefully they'll get to it some day, but in the mean time, there isn't any issue with adding the stub file.

PR #1085, opened 2024-04-05
Fix "more" button outline in soundspace example

Square789:
They turned into large blobs around this commit: 6d671c9e8503367b0f845f844a9d592735ad68af
I have to admit i didn't exactly think too deeply as to why this fixes it, but it does. The outline arc is now also in line with the `1 / zoom` thickness setting on the lines and slider box.


benmoran56 (approved):

benmoran56:
Thanks for fixing this

Issue #1084, opened 2024-04-03
pyglet needs a full docstring review and cleanup


benmoran56:
pyglet has gone through several documentation build tools in its life, leading to variation in the docstrings across the library.
Currently we use Sphinx to build the docs, which is capable of building a variety of formats. This works for the most part but, with recent additions of type hints across the library, there is a lot of inconsistency with how the API docs are rendering. 

The purpose of this ticket is to get the ball rolling on a full review and refactoring of the docstrings. This can be done in multiple parts. 

- [x] Decide on a consistent style that we will use. This is currently under discussion, and includes looking at additional sphinx plugins that may improve things with regards to type hints. 
- [x] Update the section in the documentation that describes our desired docstring format. This is still referencing `rst` style, but will potentially change. This link is here, but will need to be updated after 1. is complete: https://pyglet.readthedocs.io/en/latest/internal/doc.html
- [x] Go through each module and update the docstrings to the new format. There are many cases where things are undocumented. It would be more time effective to skip over missing things at this time, and just focus on conversion. Many modules are still missing type hints as well, but again, it would be more time effective to leave this for later.

--------------------

- [x] app
- [x] canvas
- [x] clock
- [x] customtypes
- [x] event
- [x] font
- [x] gl
- [x] graphics
- [x] gui
- [x] input
- [x] image
- [x] lib
- [x] math
- [x] media
- [x] model
- [x] resource
- [x] sprite
- [x] shapes
- [x] text
- [x] window

pushfoo:
Starting on shapes.

benmoran56:
`app` and `canvas` base classes have been completed.

benmoran56:
finished `event`

benmoran56:
Finished `sprite`.

benmoran56:
Finished the `model` module.

benmoran56:
Working on `media` now.

benmoran56:
First pass of `media` is done.

pushfoo:
TL;DR: `pyglet.libs` is up for grabs

1. I've unexpectedly had much less free time & access to an ARM64 Mac than I'd planned
2. To make sure I don't block anyone or anything:
    * I PR'd my already-finished changes in #1108
    * I'm handing off further work on `pyglet.libs` to anyone who'd like to try

matanox:
While reviewing my open issues v.s. issues here, I stumbled upon here and it struck a humble chord: 

[...]

PR #1088, opened 2024-04-14
Convert shapes to pyglet's new type annotation + docstring style

pushfoo:
TL;DR: Help with #1084 by addressing shapes.

### Changes

1. Move typing from docstrings to signatures:
   * Use the pipe syntax since it's been approved and somewhat tested on 3.8 with `from __future__ import annotations`
   * Place some annotations on default class attributes, but not all
   * Add comments explaining why
2. Convert docstrings to new formats
3. Add improvements for clarity and usability:
   1. Fix outdated references to blend modes and opacity in `__init__` doc
   2. Document broken `ShapeBase.group` behavior when `self.batch == None`
   3. Attempt to clarify rotation and anchor behavior
   4. Add cross-references
   5. Use `.. warning::` and :.. tip::` directives
   6. Use more flexible typing on `Polygon` and `MultiLine` types since they seem to target advanced users

### DIFF
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -721,7 +721,16 @@ class Arc(ShapeBase):
 
 class BezierCurve(ShapeBase):
 
-    def __init__(self, *points, t=1.0, segments=100, thickness=1, color=(255, 255, 255, 255), batch=None, group=None):
+    def __init__(
+            self,
+            *points: tuple[float, float],
+            t: float = 1.0,
+            segments: int = 100,
+            thickness: int =1,
+            color: tuple[int, int, int, int] | tuple[int, int, int] = (255, 255, 255, 255),
+            batch: Batch | None = None,
+            group: Group | None = None
+    ):
         """Create a Bézier curve.
 
         The curve's anchor point (x, y) defaults to its first control point.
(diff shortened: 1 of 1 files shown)


pyglet/shapes.py
```diff
@@ -9,8 +9,22 @@
 The Python ``in`` operator can be used to check whether a point is inside a shape.
 (This is approximated with some shapes, such as Star).
 
-To create more complex shapes than what is provided here, the lower level
-graphics API is more appropriate. See the :ref:`guide_graphics` for more details.
+If you the shapes in this module don't suit your needs, you
```
benmoran56, line 12: Looks like an extra "you" here.

pyglet/shapes.py
```diff
@@ -345,128 +377,167 @@ def rotation(self, rotation: float) -> None:
...
         self._vertex_list.rotation[:] = (rotation,) * self._num_verts
 
-    def draw(self):
-        """Draw the shape at its current position.
+    def draw(self) -> None:
+        """Debug method to draw a single shape at its current position.
+
+        .. warning:: Avoid this inefficient method for everyday use!
+
+                     Instead, add shapes to a :py:class:`Batch` and call
```
benmoran56, line 385: "Instead" is being used twice, which can be simplified.

benmoran56 (approved):

PR #1090, opened 2024-04-15
Fix ShapeBase.group setter to only migrate batch when batch is not None

pushfoo:
TL;DR: Make `ShapeBase.group`'s setter act like `Sprite.group`'s

Follow-up to #1088 based on dicussion with ballman and Ben.

### Changes

1. Only migrate when `self._batch` isn't None
2. Update docstrings to account for this


pushfoo:
The test failure is a transient sound issue. #1092 has more detail and attempts to fix it.

benmoran56 (approved):

PR #1098, opened 2024-04-28
ioctl: add ioctl wrapper for pyglet.

zhengxyz123:
This PR provides a wrapper of ioctl: `pyglet.ioctl`. It's similar to
```c
#include <linux/ioctl.h>
```
when importing the module.

This module also provides several helper functions like `_IOR_str`.

This module was excluded from docs.

Two related modules `pyglet.input.linux.evdev` and `pyglet.experimental.hidraw` were modified without test (because I don't have an input device).

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -116,7 +116,6 @@ implementations = ["cocoa", "win32", "xlib"]
 # For each module, a list of submodules that should not be imported.
 # If value is None, do not try to import any submodule.
 skip_modules = {"pyglet": {
-                     "pyglet.ioctl": None,
                      "pyglet.lib": None,
                      "pyglet.libs": None,
                      "pyglet.app": implementations,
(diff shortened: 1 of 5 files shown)


pushfoo:
Shouldn't this go in lib?

benmoran56:
Thank you for making the PR. 
I agree though, this would be better to put as `pyglet.libs.ioctl`. 
Can you make that change?

benmoran56 (approved):

benmoran56:
There were some minor import issues, but it's working fine after fixing those.

Since this is a semi-public module, I'm also going to rename the functions to remove the leading `_`. 
Also, i'm thinking about renaming them to lower case. :thinking: 
`pyglet.libs.ioctl._IOR_len` --> `pyglet.libs.ioctl.ior_len` (and so on).

zhengxyz123:
If you have the strongest desire to rename them, do it. Or don't change their name because they are common in C.

PR #1103, opened 2024-05-06
Fix uniforms in pyglet.graphics.shader

salehqt:
Fixing a bug in Pyglet where the `Uniform` object doesn't have the reference to the `ShaderProgram` it belongs to.

This causes problems when we try to write a numpy array to the uniform object.

When we try to assign to a uniform from Numpy array, it throws an error without this patch:
```
        self.program["thresholds"] = np.array([self.threshold1, self.threshold2, 0])
```

Where `thresholds` is defined like this in the shader source code:
```
uniform float thresholds[3];
```

### DIFF
--- a/pyglet/graphics/shader.py
+++ b/pyglet/graphics/shader.py
@@ -322,12 +322,13 @@ class _UniformArray:
 
 
 class _Uniform:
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

benmoran56:
Thanks for catching that, and for the PR. This will get into the next point release.

PR #1104, opened 2024-05-09
ShapeBase.delete scenarios clarification

matanox:
tentatively avoids the self-referencing description and provides more direct clarity (?) to `ShapeBase.delete()`

### DIFF
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -395,12 +395,14 @@ class ShapeBase(ABC):
     def delete(self) -> None:
         """Force immediate removal of the shape from video memory.
 
-        You should usually call this whenever you delete a shape. Unless
-        you are using manual garbage collection, Python might not call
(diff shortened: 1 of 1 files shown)


matanox:
Obviously not any great contribution

benmoran56 (approved):

benmoran56:
Thanks for the PR. Documentation fixes/clarifications are always helpful.

matanox:
Would it make sense to add a `delete` method to `pyglet.graphics.Batch`, deleting all included `Shape` objects and freeing all other video resources of the batch, extending deletion one level higher in the hierarchy of abstraction?

benmoran56:
It would be possible, but doesn't account for the Python objects. 
If you make a bunch of Sprites/Labels/etc. that are added to a Batch, and then delete that Batch, the Python objects remain.

PR #1105, opened 2024-05-13
Backport media thread join exception catch from 2.0 (9db3971)

pythonengineer:
An error is thrown commonly on application shutdown while the audio player is unloading on 1.5; this just simply backports the exception catching from 2.0.


benmoran56 (approved):

benmoran56:
Thanks for this, and the other PR. I'll push these out in a new point release shortly.

PR #1107, opened 2024-05-14
Add positional parameter support for match statement

mjcaley:
Adding the __match_args__ class attribute to vector classes. This adds support to use them in a match statement using positional arguments.

Example:
```
match Vec2(4, 2):
  case Vec2(0, 0):
    print("Match position 0, 0")
  case Vec2(x, y):
    print(f"Match at position {x}, {y}")
```


benmoran56 (approved):

benmoran56:
Thanks for the PR!

PR #1116, opened 2024-05-24
documenting the facet of flow of control in Window.dispatch_event v.s. app.platform_event_loop.post_event

matanox:
Hi,

As can be read from [Window.dispatch_event](https://github.com/matanox/pyglet/blob/29e3c0a7e637383fdd632b39e2f0bcd1c5b73884/pyglet/window/__init__.py#L670-L674) and experienced in practice, dispatching an event via `Window.dispatch_event` may result in immediate execution of the registered callbacks, before `dispatch_event` returns. This flow of execution might be surprising to application developers until they notice it when their application logic fails to assume that their callbacks registered for the event may start and finish before their call to `Window.dispatch_event` returns control.

This is addressed by the tentative docstring changes comprising this PR. 

It is meant to make the choice of flow of control explicit for application developers, when choosing between `Window.dispatch_event` and `pyglet.app.platform_event_loop.post_event`. 

Since `Window.dispatch_event` has not a docstring of its own, I hinged my docstring update in `EventDispatcher.dispatch_event` itself, if this is not too vague. It can be worked up to say "In case the EventDispatcher is a Window ....".  Maybe additional overriding classes mirror this behavior. 

---

Otherwise we could add an argument to `Window.EventDispatcher` where the caller can choose the flow of control even when using that function, as an alternative affordance, but as I'm not sure why both methods are application API and not just one of them, I can't really suggest that.

### DIFF
--- a/doc/programming_guide/examplegame.rst
+++ b/doc/programming_guide/examplegame.rst
@@ -1147,11 +1147,3 @@ I’m going to leave it as an exercise for you to do the following::
 Good luck!  With a little effort, you should be able to figure out most of
 these things on your own. If you have trouble, join us on the pyglet
 `Discord server <https://discord.gg/QXyegWe>`_.
-
-Also, in addition to this example game, there is yet *another* Asteroids clone
-available in the `/examples/astraea/` folder in the pyglet source directory.
-In comparison to this example game excercise we've just completed,
-Astraea is a complete game with a proper menu, score system, and additional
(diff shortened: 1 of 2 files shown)


matanox:
I assume that raising events from application code is a permitted (or even encouraged) blueprint. In my humble case raising events is one component to a testability framework for the pyglet application where tests raise events simulating user input to a dependency injected application class.

PR #1121, opened 2024-06-11
Concave polygon can be drawn by earcutting a polygon.

zhengxyz123:
I added an `earcut` submodule to `pyglet.extlibs` to split a polygon to several triangles.

So, pyglet can now draw concave polygon.

### DIFF
--- a/pyglet/extlibs/earcut.py
+++ b/pyglet/extlibs/earcut.py
@@ -386,7 +386,7 @@ def indexCurve(start, minX, minY, size):
     while do or p != start:
         do = False
 
-        if p.z == None:
+        if p.z is None:
             p.z = zOrder(p.x, p.y, minX, minY, size)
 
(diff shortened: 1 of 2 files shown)


benmoran56:
I checked the license for that module, and it looks OK (BSD compatible).
I think this PR is OK to merge. 

In the future, it would be nice to write our own algorithm. It could be useful for other things, and can be a bit faster than the code in that library.

zhengxyz123:
Thank you for reviewing the code.

For earcut-like algorithm, I don't think someone of us has the capability to write a triangulation library. But we can seek for some faster algorithms.
> Reading papers to reproduce an algorithm or using an existed library, which one would you choose?

We didn't take full use this library, it also support triangulating 3D polygons.

benmoran56 (approved):

PR #1127, opened 2024-06-21
Sprite Improvements and Adjustments

caffeinepills:
I've run into various issues with trying to make custom sprites, particularly with custom groups.

Here are my proposed changes:
- Merge AdvancedSprite to Sprite and deprecate it. Allow program in the initial Sprite class.
- Add `get_sprite_group` to Sprite to allow users to change initialization of a custom SpriteGroup without having to overwrite multiple functions.
- Add `blend_mode` property to allow sprites to change the default blend modes.
- Use `_user_group` to specify the intended group rather than relying on `group.parent` which may cause confusion with deeper nested groups.
- Update docs and typing.

Open to any feedback, particularly around the `get_sprite_group` function naming.

### DIFF
--- a/contrib/qt_sprite_preview.py
+++ b/contrib/qt_sprite_preview.py
@@ -52,15 +52,15 @@ both frag and vertex (IE: test becomes test.vert and test.frag)
 
 Scrolling the mousewheel also zooms in.
 """
+from __future__ import annotations
 
+import argparse
 import os
 import sys
 import traceback
-import argparse
 from pathlib import Path
-
 from textwrap import dedent
-from typing import TYPE_CHECKING, Final, Dict, List, Optional, Tuple, Mapping
+from typing import TYPE_CHECKING, Final, Mapping
 
 # Constants for choosing a Qt backend and summarizing their licensing
 ENV_VARIABLE: Final[str] = 'PYGLET_QT_BACKEND'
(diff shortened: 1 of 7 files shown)


benmoran56 (approved):

benmoran56:
I was also thinking about exposing `program` setting on the default Sprite, so no complaints there. 
Changing the blend modes looks useful. It's better than having to re-create a Sprite. 

get_sprite_group looks fine as it is, but I guess I don't see a need for it to be a public method. What did you have in mind?  

One thing we might do is add an `assert` to the @program.setter that checks if the user ShaderProgram has all the same `ShaderProgram.attributes.keys()` as the default one.  If they don't have the vertex attributes, we could warn that some default Sprite functionality won't work/crash. We can leave this idea for later though.

caffeinepills:
> I was also thinking about exposing `program` setting on the default Sprite, so no complaints there. Changing the blend modes looks useful. It's better than having to re-create a Sprite.
> 
> get_sprite_group looks fine as it is, but I guess I don't see a need for it to be a public method. What did you have in mind?

I was mostly just making it public, so that the docs can print it out for users, but maybe it's fine to make it private?

> One thing we might do is add an `assert` to the @program.setter that checks if the user ShaderProgram has all the same `ShaderProgram.attributes.keys()` as the default one. If they don't have the vertex attributes, we could warn that some default Sprite functionality won't work/crash. We can leave this idea for later though.

That sounds like a good idea.

Issue #1128, opened 2024-06-22
Types referencing class defined later


RedbeanGit:
**Describe the bug**

When importing some Pyglet classes, Python interpreter throw a `NameError: name 'some pyglet class' is not defined`. This is because the `some pyglet class` is a type hint referencing a class defined later.

Example of code throwing this kind of error (from `pyglet/image/__init__.py`):

```python
def create(width: int, height: int, pattern: ImagePattern | None=None):
    # rest of the code of this function

# ... some code here

class ImagePattern:
    # rest of the code of the ImagePattern class
```

In the above code, the `ImagePattern` is referenced in type hint for the `pattern` parameter but is defined later so the interpreter is throwing an error.

**System Information:**

Python version: 3.12.1
Pyglet version: 2.0.15
Operating System: Ubuntu (Linux)
GPU: Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics]
CPU: Intel Core i7-1165G7

**How To Reproduce**

```python
import pyglet

window = pyglet.window.Window()
test_button = pyglet.gui.PushButton(0, 0, None, None)
```

You get the following error:

```
NameError: name 'ImagePattern' is not defined
```

caffeinepills:
Your test code does not reproduce the issue.

However, we are currently in the process of adding type hints. See https://github.com/pyglet/pyglet/issues/1084

This issue is due to the forward references from what you are describing.

One way to resolve this, is to use the `typing.TYPE_CHECKING` which only runs for type checkers, not at run time.

```py
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from pyglet.image import ImagePattern
```

RedbeanGit:
> Your test code errored for me, so it didn't reproduce the issue.

Of course this code fails, that's the point.

> This issue is due to the forward references from what you are describing.
> 
> One way to resolve this, is to use the `typing.TYPE_CHECKING` which only runs for type checkers, not at run time.
> 
> ```py
> from typing import TYPE_CHECKING
> if TYPE_CHECKING:
>     from pyglet.image import ImagePattern
> ```
> 
Thanks for this work around. I also open a pull request fixing this (PR #1129).

benmoran56:
I'm also unable to replicate the issue. Running your example code against `2.0.15` produces a NoneType error as expected, but no NameErrors. Are you sure you are on an un-modifed release of 2.0.15?

[...]

PR #1129, opened 2024-06-22
image: reorder some classes to prevent NameError

RedbeanGit:
Move `ImagePattern`, `AbstractImage` and `ImageData` before their usage in `create` function as parameter type hint.

See #1128 for more details.


RedbeanGit:
Fixed in new versions of Pyglet so the pull request is not needed anymore

benmoran56:
No worries, always happy to get PRs.

PR #1131, opened 2024-06-24
add Rounded Rectangle

Kiera38:
I have created a `pyglet.shapes.RoundedRectangle` class for rectangle with rounded corners.

There is an example:
```py
import pyglet

rect = pyglet.shapes.RoundedRectangle(100, 125, 100, 50, radius=(10, 20, 5, 7))

@window.event
def on_draw():
    window.clear()
    rect.draw()

pyglet.app.run()
```
Снимок экрана 2024-06-24 173428

### DIFF
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -68,7 +68,7 @@ from __future__ import annotations
 import math
 
 from abc import ABC, abstractmethod
-from typing import Sequence, TYPE_CHECKING
+from typing import Sequence, TYPE_CHECKING, Union, Tuple
 
 import pyglet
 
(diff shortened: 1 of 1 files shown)


benmoran56 (approved):

benmoran56:
Thank you for the PR. I may make some minor tweaks to this after merging, but overall it looks good.

matanox:
Thanks for adding this feature

DevRopehero:
Thanks, i really needed this!

matanox:
Thanks again for this. It's kind of easy to crash at division by zero for the number of segments used to draw the corners, if providing a small value to the radius argument to this class constructor. 

The class documentation string says:


```
                One or four values to specify the radii used for the rounded corners.
                If one value is given, all corners will use the same value. If four values
                are given, it will specify the radii used for the rounded corners clockwise:
                bottom-left, top-left, top-right, bottom-right. A value can be either a single
                float, or a tuple of two floats to specify different x,y dimensions.

```
                
Maybe it could just further elucidate in what sense is the radius argument interpreted as radii. 

By the way, why does it draw the rounded corners using triangles?

Obviously these are obvious to anyone who ever implemented rounded corners in OpenGL, but a compact statement about the meaning of the radii for the radius argument might likely help. 

I'm sure the explanations would make the questions foolish ...

Thanks for your comments!

matanox:
Now that there's a rounded rectangle, and there is also a [bordered rectangle](https://pyglet.readthedocs.io/en/latest/modules/shapes.html#pyglet.shapes.BorderedRectangle), it's nice if there were also a bordered, rounded, rectangle :-)

matanox:
What would be a concise formula representation of the radius rendering path? I assume it's standard in graphics but which point inside the rectangle is the radius emanating from? 

The inline docstring says:
```
            radius:
                One or four values to specify the radii used for the rounded corners.
                If one value is given, all corners will use the same value. If four values
                are given, it will specify the radii used for the rounded corners clockwise:
                bottom-left, top-left, top-right, bottom-right. A value can be either a single
                float, or a tuple of two floats to specify different x,y dimensions.
```
And how precisely do the docstring mentioned x,y dimensions translate into a corner arc?

benmoran56:
You can see the math here (it is just calculated from the radius): https://github.com/pyglet/pyglet/pull/1131/files#diff-d58a3d97408215fcb68d071cf3cbbf68dcd3f1af711498f44d01cc1a7e7032c5R2054

Internal min checks could be added to prevent 0 division errors, but could be unnecessarily slow.

matanox:
Thanks. So the radius arguments (x,y) are the vertical and horizontal distance from where the corner of the rectangle would be when it's not rounded. Got it. Should have guessed that.

matanox:
Was there an integration test to this PR?
A question about adding a border option to it at https://github.com/pyglet/pyglet/pull/1280.

PR #1135, opened 2024-06-27
Dev ruta

ruta2mb:


zhengxyz123:
First, thanks for opening the PR. You leave the description empty, so can you tell us what you and your 3 firends want to improve?

In my opinion, I think rename `README.md` to `README_original.md` and replace the original README with some coverage test detail is a very unappropriate decision.
> 4 individual contributors are found on the page, but pyglet has 126 contributors. Where are they?

Then, I found the `pyglet.math` is modified to add some coverage test flags. You should separate tests from source code.

Finally, I found you improve some tests, this is great. But I'm not good at it. So other's suggestion is vital, too.

PR #1142, opened 2024-07-02
Fix outdated doc build instructions and add error text to make.py

pushfoo:
### Changes

* Move the build target info from outdated `doc/README/md` to `doc/internal/doc.rst`
* Remove outdated `doc/README.md`
  * It was last updated 2019
  * it used the old `setup.py`-based build
* Update README.md with `./make.py docs --open`
* Add link to README.md pointing to new doc build information
* Add a helpful message when doc build fails to explain why `--open` skipped previewing


benmoran56 (approved):
Thanks, looks good to me!

PR #1149, opened 2024-07-06
OpenGL ES docs

einarf:
We should really make users aware that we have experimental support for this.


benmoran56:
Thanks for taking the time to write this up. It should help a lot of people!
We can always add to it if we get any feedback about specific devices, etc.

PR #1152, opened 2024-07-07
Inline matrix operations for small speedup

DragonMoffon:
As the title says. In every case, the `sumprod` operation is slower than simply inlining the methods. I had the patience to write them all out.

Also adds inverse to Mat3

### DIFF
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -470,11 +470,11 @@ class Vec4(_typing.NamedTuple):
 
 
 class Mat3(tuple):
-    """A 3x3 Matrix
+    """A 3x3 Matrix.
 
     `Mat3` is an immutable 3x3 Matrix, wich includes most common operators.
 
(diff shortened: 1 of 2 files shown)


benmoran56:
Nice, multiplication is quite a bit faster: 
```sh
ben@ncase ~/Documents/PycharmProjects/pyglet $ python matrixbench3.py 

function                            sumprod      inline       improvement %
---------------------------------------------------------------------------
Instantiation                       0.2477       0.2492       -0.61 %
Multiplication                      3.8496       2.0451       88.24 %
From Rotation                       5.5438       5.5508       -0.13 %
Inversion                           2.7191       2.7216       -0.09 %
```

Inversion on the other hand looks identical (if not a fraction slower), but it is more readable so lets keep those changes too.

benmoran56 (approved):

Issue #1134, opened 2024-06-27
Visual artifacts of bezier curves with custom thickness 


MassPig:
**Describe the bug**
Some bezier curves seem to result in visual bugs if thickness > 1 is used.
Example:
```
import pyglet
from pyglet.shapes import BezierCurve, Circle

window = pyglet.window.Window(width=400, height=300)
batch = pyglet.graphics.Batch()

bezier = BezierCurve((100, 100), (300, 300), (300, 100), (100, 300), thickness=10, color=(0, 0, 128), batch=batch)

@window.event
def on_draw():
    window.clear()
    batch.draw()

pyglet.app.run()
```
Output:
![grafik](https://github.com/pyglet/pyglet/assets/69081459/70f2b7d5-8ad9-49f3-a731-ceb434dd2dd9)
The blue spikes are inconsistent with other bezier curve tools I have used so far. For example, in Inkscape a similar curve looks like this:
![grafik](https://github.com/pyglet/pyglet/assets/69081459/d0626283-8ebf-43b7-8f7e-450a0b23a7df)
Without any artifacts. I guess this behavior is not desired by the developers of this project.

The thickness seems to be a key parameter here. For thickness=1 the curve looks fine, and the artifacts get bigger with higher thickness.

**System Information:**

Here is my `python -m pyglet.info`, I deleted some more personal information (paths).
Platform
------------------------------------------------------------------------------
platform:  Linux-5.15.0-107-generic-x86_64-with-glibc2.35
release:   5.15.0-107-generic
machine:   x86_64

Python
------------------------------------------------------------------------------
implementation: CPython
sys.version: 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
sys.maxint: 9223372036854775807

pyglet
------------------------------------------------------------------------------
pyglet.version: 2.0.15
pyglet.compat_platform: linux

pyglet.options['audio'] = ('xaudio2', 'directsound', 'openal', 'pulse', 'silent')
pyglet.options['debug_font'] = False
pyglet.options['debug_gl'] = True
pyglet.options['debug_gl_trace'] = False
pyglet.options['debug_gl_trace_args'] = False
pyglet.options['debug_gl_shaders'] = False
pyglet.options['debug_graphics_batch'] = False
pyglet.options['debug_lib'] = False
pyglet.options['debug_media'] = False
pyglet.options['debug_texture'] = False
pyglet.options['debug_trace'] = False
[...]

PR #1155, opened 2024-07-08
Fix miter artifacts in issue #1134

kingadami:
This adds a fix to keep the miter scale from getting too large for shapes which use the miter functions.


kingadami:
In the long run we might should look at emulating the behavior of inkscape and don't even do a miter for angles which exceed a given scale.  Just a thought.

benmoran56 (approved):

benmoran56:
Thank you for the PR kingadami.  Mimicking Inkscape sounds reasonable to me, but i'll let someone more familiar with the matter make that decision.

PR #1158, opened 2024-07-09
Allow changing initial state of ToggleButton

izxxr:
Adds a `toggled` parameter to `gui.ToggleButton` that determines the initial toggled state of the widget.


benmoran56:
Sorry for the delay. 
To avoid the extra code in the library, the `value` property exists for this purpose. All Widgets contain this property to be used for manually setting the value (such as initial value). For example: 
```py
my_pushbutton = PushButton(...)
my_pushbutton.value = True
```

izxxr:
Oh alright, didn't know that. Thanks.

PR #1166, opened 2024-07-13
Add multi-texture sprite support.

kingadami:
The multi-texture sprite supports overlaying multiple textures for a single sprite.  It assumes you set the images alpha value for transparancy effects in the overlay.  There are still limitations to this implementation.

What was added:
- Added MultiTextureSprite to the experimental directory.
- Added an example of using it for both animations and static images.

Outstanding items:
- Need to make a decision on how big the sprite will be.  Probably will just make it the max size of all of the layers and then center each layer.
- Need to test with an animation with frames spanning different texture ids.
- Currently uses pyglet.sprite.Sprite as the base class which means there are methods which need to be cleaned up and all features of sprites need to be tested to make sure it didn't break things.


kingadami:
This is still a work in progress and there are still features I want/need to add.  I am looking for feedback on the concept and implementation.

caffeinepills:
Thanks for the PR! I like the idea of expanding the current multi texture sprite example [here](https://github.com/pyglet/pyglet/blob/master/examples/sprite/multi_texture_sprite.py) that can be potentially part of the sprite module.

A couple things that come to mind from the current PR that I like and don't like: I like and don't like the idea of automatic shader creation. I like it because it's pretty clever, and something we could potentially use for very specific use cases. 

However, the biggest problem right now with this implementation is there is no current way for a user to pass in their own shader program or easily modify the existing one.  It would also be hard for a user to subclass this to make their own shader. Even if they try to modify from the `__init__` of the class, I feel like the debug output may not be helpful for line numbers. All the += behavior makes it very hard to understand it, much less modify it. Maybe this would look better with some sort of f-string usage, regex, or function that can generate it. 

I also think limiting the texture names to specific built in names like `tex_{idx}` is not ideal for general usage (I think it would be fine for an example). If someone is messing with 3+ sampler names, that can potentially get confusing without more descriptive names with a larger shader. I would prefer more of a dict[name, texture] behavior like the original, but that's just me.

Personally I think just not worrying about shader generation for the actual sprite class itself is probably best.  With a multi-texture sprite, I feel like those are going to be tailor made for a user's situation, so they will generally be supplying their own. Having it just multiply 2+ textures is pretty limiting for general usage. Again, I think it's fine as an example, but not for it's only behavior. So I think it would make sense for them to just supply their own shader. (And maybe I'm wrong here, but anyone else feel free to chime in)

I like the idea of being able to potentially change the textures after the fact and with animation support, especially across different textures. 

I think technically shaders can support multiple sampler types in the same shader (sampler2D and sampler2DArray), maybe that would be nice to remove the restriction on that, so it's more robust. (I only limited it in my example at the time because I didn't know if it was supported or not)

kingadami:
> However, the biggest problem right now with this implementation is there is no current way for a user to pass in their own shader program or easily modify the existing one. It would also be hard for a user to subclass this to make their own shader. Even if they try to modify from the `__init__` of the class, I feel like the debug output may not be helpful for line numbers. All the += behavior makes it very hard to understand it, much less modify it. Maybe this would look better with some sort of f-string usage, regex, or function that can generate it.

Maybe a potential solution to this is have the concept similar to a python generator for the shader program and allow that to be passed in (I do mean in the loose sense not the literal python generator for lists etc.).  This could be as simple as a string containing the static program similar to how the `get_default_shader` function works today, or a complicated function which can do things I can't even think of yet.  We could then supply some predefined ones to lower the barrier of entry for using multi-texture sprites.  They would of course have to make assumptions but the user doesn't have to use them.  I would just have to figure out how to efficiently generate the vertex list potentially for a program I don't have any control over, but I am sure it can be done.

kingadami:
>I also think limiting the texture names to specific built in names like tex_{idx} is not ideal for general usage (I think it would be fine for an example). If someone is messing with 3+ sampler names, that can potentially get confusing without more descriptive names with a larger shader. I would prefer more of a dict[name, texture] behavior like the original, but that's just me.

I was switching between keeping it as a dict like the original example or making a list like I did.  I decided to make it a list because I needed to specify an order and from what I understand as of 5 minutes ago was that dicts don't guarantee order.  But... I am now reading that as of python 3.7 insertion order is maintained so I can change back to a dict.  I actually like that way as well.

pyglet/experimental/multitexture_sprite.py
```diff
@@ -0,0 +1,331 @@
...
+            if texture.id != other.textures[name].id or texture.target != other.textures[name].target:
+                return False
+
+        # Made it this far so just check the remainder
+        return (self.program is other.program and
+                self.blend_src == other.blend_src and
+                self.blend_dest == other.blend_dest)
+
+    def __hash__(self) -> int:
+        tex_id = tuple([texture.id for texture in self.textures.values()])
```
caffeinepills, line 87: Maybe a performance speedup to cache this on group initialization? That way it's not making tuples every time it's hash checked?

pyglet/experimental/multitexture_sprite.py
```diff
@@ -0,0 +1,331 @@
...
+    fragment_source.append("  return foreground * foreground.a + background * (1.0 - foreground.a);")
+    fragment_source.append("}")
+
+    fragment_source.append("void main() {")
+    fragment_source.append("  vec4 color = vec4(0.0, 0.0, 0.0, 1.0);")
+    fragment_source.extend([f"  color = layer(texture({name}, {name}_coords_frag{_SAMPLER_COORDS[tex.target]}), color);" for name,tex in images.items()])
+    fragment_source.append("  final_colors = color * vertex_colors;")
+    fragment_source.append("}")
+
+    fragment_source = '\n'.join(fragment_source)
```
caffeinepills, line 163: This area looks a lot cleaner.

Perhaps just inject what needs to be changed, so you can keep it all one string?

For example:
```py
in_tex_coords = "\n".join([f"in vec3 {name}_coords;" for name in images])

frag = f"""
#version 150 core
in vec3 translate;
in vec4 colors;
in vec2 scale;
in vec3 position;
in float rotation;
{in_tex_coords}
 """

print(frag)
```

caffeinepills (commented):
Looks a lot better, made a couple comments.

pyglet/experimental/multitexture_sprite.py
```diff
@@ -0,0 +1,331 @@
...
+
+    return pyglet.gl.current_context.create_program((vertex_source, 'vertex'), (fragment_source, 'fragment'))
+
+class MultiTextureSprite(pyglet.sprite.Sprite):
+    """Creates a multi-textured sprite."""
+    group_class = MultiTextureSpriteGroup
+
+    def __init__(self,
+                 images: list[AbstractImage | Animation],
```
caffeinepills, line 172: Should be dict now?

pyglet/experimental/multitexture_sprite.py
kingadami: Good idea.  Implemented requested change.

kingadami:
I think I have most of the functionality there.  The only things are related to the fact that I inherit from the Sprite class.  For instance, the multi-texture sprite class still has the frame_index attribute from the Sprite class which no longer makes sense.  I don't know what would be the recommended way to prevent it from being used: Delete the attribute? Override it with an exception?  Other than this, I think it is all there unless anyone has any suggestions.  I would eventually love to see this get moved to the sprite module, but understand if you want to keep it in the experimental module for now.  I believe that it is relatively generic, with the help of your feedback, and allows for the use of multi-texture sprites with the ability to customize it as needed.

kingadami:
Note for me to remove the __init__.py file from experimental.

benmoran56 (approved):

PR #1174, opened 2024-07-19
Screen and Display bases

einarf:
Quick test making bases for `Screen` and `Display` so it's possible to do typing. The `Screen` and `Display` types are dynamically overridden by the platform specific types so it's impossible to do anything related to typing with those.

Accidentally whitespace trimmer on. Can revert that if needed.


einarf:
There is another way to hack around his

PR #1182, opened 2024-07-25
Formatted text optimization

Kiera38:
some formatted text optimizations.
added `append_text` function to append text to the end of the document.

### DIFF
--- a/pyglet/text/document.py
+++ b/pyglet/text/document.py
@@ -247,7 +247,7 @@ class AbstractDocument(event.EventDispatcher):
         self._text = ""
         self._elements: list[InlineElement] = []
         if text:
-            self.insert_text(0, text)
+            self.append_text(text)
 
     @property
     def text(self) -> str:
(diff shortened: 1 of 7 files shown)


caffeinepills:
Hey Kiera, thanks for taking the time to do this. I have been wanting to look into this at some point as well.

One thing that doesn't work with this approach is `IncrementalTextLayouts`. With those (aka Textboxes) it just replaces the text entirely.

You can test this via examples/text/text_input.py (you may have to sync back to master branch as an import was just fixed to make this work). There are also interactive examples via tests/interactive/text

You may want to change it to only appending when the text is specifically added at the end and inserting at other times.

Kiera38:
Fixed it. The text is normally edited. `IncrementalTextLayout` incorrectly updated `owner_runs` in `on_insert_text`.

caffeinepills:
Thanks for fixing that. Upon further testing, it does appear there is still an issue somewhere.

It's now possible for a glyph box to be completely empty.

Steps: 
Open up text_input.py
For the text box "This is a test"
Replace "is" with "was"
Backspace all characters until no characters left. 
The resulting _GlyphBox will exist with no glyphs.

Kiera38:
Sorry for the long delay.  Fixed a small bug.  Now the text is deleted normally after editing.

caffeinepills:
Sorry I couldn't look at this sooner. Thanks for that. After testing it, I can't see any issues, seems good!

However, maybe this was overlooked, but changing the `text` property on the document doesn't actually take this new behavior into account. From what I can tell this only calls `append_text` when you are creating the document. Performance is roughly the same on my machine between master and your branch in this regard. However, I am only using a default label, this may be different on a different kind of document. Just curious what improvements you are seeing using this optimization?

Kiera38:
The difference between `insert_text` and `append_text` is only in `FormattedDocument`, when we add text in parts, as in decoders. There is not much difference for changing the `text` property, but `append_text` will be a slightly faster.

Also, `insert_text` has been optimized here, which now uses `RunList.insert_run` instead of `RunList.insert` + `RunList.set_run` to set the style in most cases. 

In addition, `append_run` is now also used to create `owner_runs`, which will be faster if the font changes frequently.

caffeinepills (approved):
Looks good.

caffeinepills:
My apologies I should have looked at this closer. @Kiera38 After looking further, it seems that this broke formatting with the attributed documents.

What it currently looks like:
image


What it should look like:
image

Issue #1175, opened 2024-07-19
Migration guide for major breaking changes due to audio backend rework


fladd:
As described [here](https://github.com/pyglet/pyglet/pull/975#issuecomment-2240422290), the major breaking audio changes that have been introduced in 2.0.12 have effects on other software that relies on Pyglet for audio playback (in my case https://github.com/zipped-album/zap).

Due to the lack of documentation of these changes, it is not straight-forward to make already existing (now broken) code - which worked fine in Pyglet 1 and Pyglet 2 (until 2.0.12) - work again.

Assuming that 2.0.10 will not be maintained in parallel in the future, a migration guide, which explains the changes in detail and gives step-by-step instructions on what parts need adaptation, would help with migrating existing software to 2.0.12 and beyond.

caffeinepills:
Thanks for the report.

Can you provide information to help us with your issue?  Including the information specified in the bug report form, if you can share the output of the `python -m pyglet.info` command. If this is not possible, please provide the OS information, Python information, Codec (If ffmpeg, which version), audio driver being used (PulseAudio, OpenAL, etc), and if possible an example of audio having issues (or format).

fladd:
@caffeinepills Thanks for getting back to me.

I indeed had to override the `__init__` of `FFmpegSource` for two reasons:
1. The original did not support 24bit FLAC files (it does specify 24 bits, but as far as I could tell, files when loaded in Pyglet were 32bit padded, so I added that and it worked)
2. I needed to implement dithering, because OpenAL does not support 24 bits, so I had to resample, but I wanted to do this properly (i.e with dithering and noise shaping)

I actually just had a look at my customized `__init__` yesterday and it seemed pretty up to date. I only had to change the part that used `asbytes_filename`, as this does not seem to be supported since 2.0.12 anymore.

The temp file loading I do is because I play files directly from a ZIP file, and for this to work, I apparently need to first unzip the file I want to play into a temporary file on disk.

I am getting somewhere, though! With one simple change (apparently the DirectSound`buffer_size` and OpenAL`ideal_buffer_size` have been unified to `_buffered_data_ideal_size` across all drivers in 2.0.12), I can now get my `AudioPlayer` class to work! I have audio and it seeks correctly it appears.

[...]

PR #1184, opened 2024-07-27
Fix for issue uncovered in #1175

Square789:
This fixes the issue of `PreciseStreamingSource`s not passing through their wrapped source's attributes by removing it entirely, moving the logic that buffers audio from imprecise sources into the `AudioPlayer` base class.

Wrapping sources like that is quite messy, the truly solid alternative would be forwarding source attributes entirely via property overrides (See https://github.com/Square789/pyglet/tree/precise_streaming_fix_a). I did not open that branch as a PR because i believe this solution is preferrable.  
This commit series also removes an instance in `Player` where the `PreciseStreamingSource` wrapper had to be created parallel to the `Source.get_queue_source`, removing a duplication of logic.

### DIFF
--- a/pyglet/media/codecs/base.py
+++ b/pyglet/media/codecs/base.py
@@ -389,9 +389,9 @@ class Source:
         """Return the ``Source`` to be used as the queue source for a player.
 
         Default implementation returns ``self`` if this source is precise as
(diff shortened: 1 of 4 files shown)


benmoran56 (approved):

benmoran56:
Thanks Square!

PR #1189, opened 2024-08-03
Gui widget fixes1

shenjackyuanjie:
fix some typing in `event`
and fix `TextEntry` background

image

### DIFF
--- a/pyglet/gui/widgets.py
+++ b/pyglet/gui/widgets.py
@@ -118,7 +118,7 @@ class WidgetBase(EventDispatcher):
         raise NotImplementedError('Value depends on control type!')
 
     @value.setter
-    def value(self, value: int | float | bool):
+    def value(self, value: int | float | bool) -> None:
         raise NotImplementedError('Value depends on control type!')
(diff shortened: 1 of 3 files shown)


shenjackyuanjie:
image

oh, I forgot to reinstall it, it's fine

pyglet/gui/widgets.py
```diff
@@ -445,7 +445,7 @@ def __init__(self, text: str,
                 Optional parent group of text entry widget.
         """
         self._doc = pyglet.text.document.UnformattedDocument(text)
-        self._doc.set_style(0, len(self._doc.text), dict(color=text_color))
```
benmoran56, line 448: Doesn't the linter prefer `dict(...)` instead of `{...}`?

benmoran56:
Closing because this work is also in the other PR.

pyglet/gui/widgets.py
shenjackyuanjie: it prefer `{...}` better

PR #1200, opened 2024-09-03
Fixed two simple typos in docs, missing "t" and "an" before "Window" 

lizard-wizard-git:
… which is a word that begins with a consonant sound, so it should be "a Window"

(Sorry if something is wrong, it's my first pull request)


benmoran56 (approved):

benmoran56:
Thank you for this. Everything looks good. And congratulations on your first pull request!

lizard-wizard-git:
Thank you! Feels good to do something useful :)

PR #1212, opened 2024-09-30
Support OpenGL structures in UBO's

caffeinepills:
Adds/fixes support to allow Structures from OpenGL to be created with UBO's. This also works now with nested structures and structure arrays.

Doing this is much more performant than updating dozens or hundred of individual uniforms for game objects.

Added a test to retrieve the data after setting it to ensure it's entered correctly. Need to eventually add a projection back to the Shadow Window, this way we can use a FBO to determine if the shader actually set the right value by looking at the pixel data of a sprite. Otherwise this should be good.

Feel free to let me know if you find any bugs.

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,3 +1,30 @@
+pyglet 2.0.17
+
+Changes and Improvements
+------------------------
+- image: Use dedicated shader for blitting to fix blending issues. (#1170)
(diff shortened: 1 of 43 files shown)


benmoran56 (dismissed):

benmoran56:
This looks great! I'll test this a bit on my end, but I don't see any issues with the implementation.

benmoran56 (approved):

PR #1216, opened 2024-10-05
Fix F1,F2,.. handling on Mac

mnowotnik:
Regression was created in 1.5.28 after switching to NSApplication event loop. F1,F2,F3,.. keys stopped being properly handled. Only keyup events are registered, but not keydown.

Attempt to fix by sending F* keydown events to window. They seem to get lost in the propagation chain, but I don't know why.

### DIFF
--- a/pyglet/app/cocoa.py
+++ b/pyglet/app/cocoa.py
@@ -6,7 +6,25 @@ from pyglet.app.base import PlatformEventLoop, EventLoop
 from pyglet.libs.darwin import cocoapy, AutoReleasePool, ObjCSubclass, PyObjectEncoding, ObjCInstance, send_super, \
     ObjCClass, get_selector
 
-NSApplication = cocoapy.ObjCClass('NSApplication')
+
+class PygletNSApplication_Implementation:
+
+    _F_KEYS = (cocoapy.NSF1, cocoapy.NSF2, cocoapy.NSF3, cocoapy.NSF4, cocoapy.NSF5, cocoapy.NSF6, cocoapy.NSF7,
+               cocoapy.NSF8, cocoapy.NSF9, cocoapy.NSF10, cocoapy.NSF11, cocoapy.NSF12, cocoapy.NSF13, cocoapy.NSF14,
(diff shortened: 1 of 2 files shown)


caffeinepills:
Hey @mnowotnik good catch! Sorry about that.

The use of `NSApplication` is still kind of a WIP to address issues between various Python versions and MacOS versions.

Looking into this, I think the fix may be a little simpler, but it looks to be because of the `performKeyEquivalent_` function. If it returns `False` it means the event was unhandled and gets passed to the responders, if it returns `True` it doesn't go through the responder chain.

I am not 100% sure why it was done this way.  Maybe it has to do with the fact some Function keys also have other capabilities like volume or brightness. Furthermore I am not sure why we only caught just the Home, End, Page Up, Page Down keys. As far as why we are doing that, the notes say `Overriding this method helps prevent system beeps for unhandled events.` however I can't seem to replicate any 'beeps' using my keyboard when commenting out the function itself.

https://github.com/pyglet/pyglet/blob/7387e594bee7e84664f931c2d9486636e4c28f8c/pyglet/window/cocoa/pyglet_view.py#L218-L233

To allow all function keys through we can do this:
```py
    def performKeyEquivalent_(self, nsevent):
        # Let arrow keys and certain function keys pass through the responder
        # chain so that the textview can handle on_text_motion events.
        modifierFlags = nsevent.modifierFlags()
        if modifierFlags & cocoapy.NSNumericPadKeyMask:
            return False
        if modifierFlags & cocoapy.NSFunctionKeyMask:
            return False
        # Send the key equivalent to the main menu to perform menu items.
        NSApp = cocoapy.ObjCClass('NSApplication').sharedApplication()
        NSApp.mainMenu().performKeyEquivalent_(nsevent)
        # Indicate that we've handled the event so system won't beep.
        return True
```
If you could test and see if this resolves it on your end.

If you want to test further and remove the function entirely and see how it impacts things, that would be nice. See if it can produce 'beeps'. If that can be resolved, we can probably just remove the overridden function entirely now.

benmoran56 (approved):

PR #1223, opened 2024-10-06
Revert "Fix F1,F2,.. handling on Mac (#1216)"

mnowotnik:
This reverts commit 686697c0bfc38d0f62022d8e95bcffbcc69a5668.

Tested newest master branch and after changes from @caffeinepills this fix is no longer needed.


caffeinepills:
Thanks for testing. I have reverted these changes already so we should be good.

benmoran56:
Just to add, it tests OK on my M1 mac as well.

PR #1226, opened 2024-10-13
Move py.typed to include it in wheel

viblo:
This moves py.typed marker file into the pyglet folder, since that is what is put into the wheel and later installed when pyglet is installed. (Right now its not included in the published wheel)


benmoran56 (approved):

benmoran56:
Ah, thanks for that.

PR #1229, opened 2024-10-22
Stop F5 from autocompleting on mac

mnowotnik:
F5 triggers list of autocompletions on MacOS and because NSTextView is global for the whole app, it shows in the bottom left corner. This change disables it.


benmoran56:
Hi @mnowotnik, thank you for the PR, and sorry for the long delay.

For reference, can you say which hardware platform you tested against? IE:  M1 Macbook Pro, etc.
There seems to be slightly different behavior between different architecture Macs.

mnowotnik:
Just M1 MacBook Pro

benmoran56 (approved):

benmoran56:
Thanks Michał!

PR #1241, opened 2024-12-05
Add support for FFmpeg 7.1

caffeinepills:
Update FFMpeg to support new versions as lots of deprecations and re-orderings were made to the structures.

Tested on Windows 10 x64 with version Ffmpeg 7.1 and Python 3.10 and 3.12.


benmoran56 (approved):

benmoran56:
Works fine on Arch Linux with FFmpeg n7.1

PR #1251, opened 2025-01-03
Fixes for typing

pvcraven:
In working with the Arcade library, I noticed the latest rc3 had some issues with imports and typing. Pyright was not happy.

This PR fixes those issues and Arcade is able to import things and keep pyright happy.


eruvanos:
Could you also remove the `__match_args__` from the Vec classes (like: https://github.com/pyglet/pyglet/blob/093206c7e9152455d4552549eaa2a3ee5a1a5a7f/pyglet/math.py#L54)

That makes mypy unhappy about to much arguments for Vec2-Vec4

pvcraven:
> Could you also remove the __match_args__ from the Vec classes

Pushed an update with that removed. It does fix over 100 errors in arcade's mypy run.

benmoran56 (approved):

PR #1265, opened 2025-02-01
Small fixed in quickstart docs and examples

cdeil:
I think the `__docformat__` and `__version__` in the examples is useless?
(git blame shows it wasn't touched since 17 years)

If correct I could send another PR to systematically remove it from all example files or add that here if you prefer.

### DIFF
--- a/doc/programming_guide/quickstart.rst
+++ b/doc/programming_guide/quickstart.rst
@@ -190,8 +190,8 @@ Specify ``streaming=False`` in this case::
     sound = pyglet.resource.media('shot.wav', streaming=False)
     sound.play()
 
-The `examples/media_player.py` example demonstrates playback of streaming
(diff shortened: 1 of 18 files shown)


benmoran56 (approved):

benmoran56:
Thanks for the PR Christoph. 

Yes, if you have the ambition, please go ahead with removing those markers. They are not needed, but I guess nobody touched them since they are harmless.  I suppose it could be confusing for new users who might think they are something important.

cdeil:
> go ahead with removing those markers

@benmoran56 - Done in 8e80d8a .

> I suppose it could be confusing for new users who might think they are something important.

Yeah, I want to use Pyglet / Arcade examples to teach kids coding and think the simpler / less distractions the better.

benmoran56 (approved):

PR #1267, opened 2025-02-01
Feature shapebase shader rescaling

cteberg:
Added rescaling to ShapeBase objects using shaders. The implementation is based on sprite rescaling pattern.


benmoran56 (changes requested):
Looks good! However, we should add a small note to the module docstring, and the `ShapeBase.__contains__` method docstring. The Python `in` operator won't take scaling into account, so we should mention that it does not account for scaling.

benmoran56 (approved):

PR #1269, opened 2025-02-02
Simplify super().__init__ calls using Python 3

cdeil:
When reading the tutorial I saw that you still had a few cases of the old Python 2 syntax to call `super()` in `__init__` methods. This PR updates all such cases to the modern Python 3 syntax. Should behave identically.


benmoran56 (approved):

benmoran56:
Thanks for the effort!

PR #1274, opened 2025-02-11
Simplify docs version handling

cdeil:
This is a proposal to simplify the docs version display and version cross-linking on RTD so that it's zero maintenance, no manual edits in RST docs needed as you push new versions.

Note that currently you have a link to v1.5 docs but that gives a 404 "not found" error. I see you have that branch on Github, so if you think users need those old docs you could go to the RTD config and build it and it would show up In the RTD version selector widget.

Screenshot 2025-02-11 at 23 43 03

I've tried to maintain many-version docs with cross links pointing out older and newer releases before and it was always painful, especially pointing to newer releases. I see e.g. https://docs.astropy.org/en/stable/ and other projects also have settled on the simple solution to just show the version on the landing page.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1,13 +1,9 @@
 pyglet Documentation
 ====================
 
-.. ATTENTION::
-   This documentation is for the pyglet 2.1 series, which has a few small API
-   changes from the 2.0 series. Previous documentation can be found at:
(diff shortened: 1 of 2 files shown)


benmoran56 (approved):

benmoran56:
I think the original notice has served it's purpose.  When we release the next big major version, I may again add a temporary message, but for now this is OK to remove.

PR #1277, opened 2025-02-15
Fix Xlib screen mode being dotclock instead of the true hz.

csd4ni3l:
Fix Xlib screen mode being dotclock instead of the true hz.  In Windows, the .rate returns the hz of the display, but on Linux, it returns dotclock, which is not the same. This fixes that issue.


benmoran56 (approved):

benmoran56:
Thank you!

csd4ni3l:
Np. The only problem with this is its not 100% accurate, so users might need to round it up.

PR #1281, opened 2025-02-24
Updated the torus example to work with the material updates in 3.1

Kaundur:
Updated the opengl torus example to work with the shader and material update in 3.1


benmoran56 (approved):

benmoran56:
Thanks Kaundur

PR #1290, opened 2025-03-11
DocString indicating arc_start is in radians instead of degrees

Brett208:
I noticed when updating my usage of pyglet to the 2.1 series, I think arc was switched to utilize degrees publicly, but the start_angle is still listed as being in radians in one place.

Correct the docstring for the start_angle of an arc to indicate it is in degrees and not radians

Thanks,
~Brett


benmoran56 (approved):

benmoran56:
Good catch, and thank you for the PR Brett!

PR #1294, opened 2025-03-19
Text fixes, additions, and improvements

caffeinepills:
This is quite a large pull request with lots of text additions, so it should be tested more thoroughly on other machines. So far I have fixed all issues I have encountered.

Below is what I've done
### All OS's
* Add `harfbuzz` library support for text shaping. This should improve spacing, ligatures, and substitutions for people who need more accurate text positioning and glyph shaping. This is supported for all of the main font renderers for all 3 main OS's.
* Added `font.get_text_size` function. This will return the width and height of a string of text without having to render it. This will greatly improve speed for those who need to spot check sizes, or determine a scale, for their own purposes before rendering. It may be beneficial to cache the character sizes yourself. Does not take kerning or multilines into account. This may be improved later. (Note: That this returns the width and height of the combined glyphs. Text Layouts `content_width` and `content_height` may return slightly different sizes, as a layout may return the actual font height `(font.ascent - font.descent)`. and include kernings or shaping depending on the OS, however it should be fairly close.)
* Default text shader now multiplies text color. Previously this was inaccurate due to non-DPI aware windows AND MacOS rendering text as black. Now that both of these are resolved, this should allow color bitmap fonts to display properly. Note that they are currently affected by the color run of the text they are a part of.
* Add initial support for fallback fonts. This should help in situations where a glyph is not present in one font, but present in another. `font.add_fallback(font)`. 
* Added a GlyphPosition that is paired with each glyph for per-run unique positions.
* Optimize get_grapheme_cluster function to be quicker.

### New options:
- text_antialiasing: Disable font renderers from anti-aliasing text. This may improve the appearance of more pixel style fonts at larger scales. This is global and the results are cached. [If you want to have only certain fonts be affected by this, set this option, call `font.get_glyphs(characters)` with all of the characters you want to cache, then you can change the setting again.]
- text_shaping:     
     * ``False``, Disables the shaping process for text. This may increase performance as it reduces the amount
        of calls during rendering. If your font is simple, monospaced, or you require no advanced OpenType features,
        this option may be useful.
     * ``'platform'``, Uses platform's font system for shaping. Supported by Windows (DirectWrite) and Mac (CoreText).
     * ``'harfbuzz'``, Utilize the harfbuzz library for font shaping. This requires an optional dependency, if not
     found, it will fallback to platform/no shaping.

### Windows
* Refactor DirectWrite into it's own module, separate it from the lib as it was growing large.
* Fix DirectWrite bugs where glyphs that are merged or missing due to shaping may crash.
* Refactor DirectWrite to use callbacks for it's custom renderer system, this gives us shaping information without having to make a bunch of calls on our end.
* Separate WIC into it's implementation and lib file. Made it easier for other modules to interact with it.

### Linux
* Now supports different weight styles, and stretch styles.
* Update freetype to support color glyphs. (Note: With a limitation, some bitmap glyphs are a fixed size, other font renderers will scale the size of the bitmaps, but freetype will not. This may be adjusted in the future on our end.)
* Retrieving glyphs should be much faster now, as we were using fontconfig (they themselves recommend not to use it). 
* Properly sets sizes on fixed fonts.
* Freetype no longer returns data in one channel, it should now return in RGBA. (Function for 1 channel conversion may have to be optimized later)

### MacOS
* CoreText can now render by glyph index.
* Optimize fetching glyph data from a buffer (twice as fast)
* Now supports specifying different weight and stretch styles.
* Fix missed freeing of CoreText resources.
* Now renders text in white instead of black. 
* Now flips the image in the context so we don't have to flip texture coordinates.

### Other
* Fix examples and tests.
* Add a User defined bitmap font example using a test font we have.
* Clean up import issues. A bulk of the files are just import changes, which I had to take care of due to star import issues with _lib files on Windows


benmoran56 (dismissed):

DigiDuncan:
Absolutely required for Charm. Very excited to see this in the library.

benmoran56 (approved):

PR #1295, opened 2025-03-22
Improve missing or underdeveloped math doc

pushfoo:
This is a prelude to a much bigger series of programming guide updates.

- [ ] Quaternion doc
- [ ] `Vec2` fixes:
  - [ ] Explain what `Vec2`-like means somewhere to avoid [uncrossreferenced vagueness](https://github.com/pyglet/pyglet/blob/81552da62c1df4539620205e493c4f40ba79a743/pyglet/math.py#L270)
  - [ ] Cross-ref or explain [`.rotate()` direction](https://github.com/pyglet/pyglet/blob/master/pyglet/math.py#L280)
  - [ ] Rephrase clamp for clarity
- [ ] `Vec3` fixes:
  - [ ] Rephrase [top-level docstring](https://github.com/pyglet/pyglet/blob/81552da62c1df4539620205e493c4f40ba79a743/pyglet/math.py#L371-L378)
    - [ ] cross-ref `len`
    - [ ] swamp note to `.. important::`
  - [ ]  [`s/simply//`](https://github.com/pyglet/pyglet/blob/81552da62c1df4539620205e493c4f40ba79a743/pyglet/math.py#L548-L553)
  - [ ] cross-ref [`lerp` method with a lerp explanation](https://github.com/pyglet/pyglet/blob/81552da62c1df4539620205e493c4f40ba79a743/pyglet/math.py#L576-L584)
  - [ ] explain distance [func args](https://github.com/pyglet/pyglet/blob/81552da62c1df4539620205e493c4f40ba79a743/pyglet/math.py#L591-L597) better 

## In The follow-up PR to come

1. Centralize "unit vector" summary with cross-refs to external math for:
   -  [`Vec2.normalize()`](https://github.com/pyglet/pyglet/blob/81552da62c1df4539620205e493c4f40ba79a743/pyglet/math.py#L298-L302), as well as for `Vec3` and `Vec4`
 2. Matrix explanations of:
   - (briefly) how to matmul in Python / why
   - cross-refs to external info

### DIFF
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -1396,7 +1396,17 @@ class Mat4(_typing.NamedTuple):
 
 
 class Quaternion(_typing.NamedTuple):
-    """Quaternion."""
+    """Quaternions are special 4-dimensional vectors useful for 3D rotations.
+
+    This type is separate from the :py:class:`Vec4` type for clarity when:
+
+    * organizing helper methods
+    * annotating types
+
+    .. tip:: Store method results if you will need them more than once.
+
+             Otherwise, you will be re-calculating values wastefully.
+    """
 
     w: float = 1.0
     x: float = 0.0
@@ -1412,6 +1422,8 @@ class Quaternion(_typing.NamedTuple):
         raise NotImplementedError
 
     def to_mat4(self) -> Mat4:
+        """Calculate a 4 by 4 transform matrix which applies a rotation."""
+
         w = self.w
         x = self.x
         y = self.y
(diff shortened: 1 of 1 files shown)


pyglet/math.py
```diff
@@ -1437,6 +1449,8 @@ def to_mat4(self) -> Mat4:
         return Mat4(a, b, c, 0.0, e, f, g, 0.0, i, j, k, 0.0, 0.0, 0.0, 0.0, 1.0)
 
     def to_mat3(self) -> Mat3:
+        """Convert the quaternion to a 3 by 3 rotation matrix."""
```
benmoran56, line 1452: This should probably be "Calculate", similar to the `to_mat4` method. "Convert" could be confusing.

pyglet/math.py
```diff
@@ -1396,7 +1396,17 @@ def __repr__(self) -> str:
 
 
 class Quaternion(_typing.NamedTuple):
-    """Quaternion."""
+    """Quaternions are special 4-dimensional vectors useful for 3D rotations.
```
benmoran56, line 1399: I would avoid mentioning the Vec4 class entirely here. Quaternions are not really vectors, but a 4d number that is a type of complex number (as I understand it).

benmoran56 (approved):

benmoran56:
Going to merge this in as it is, so the already completed work isn't left dangling.

PR #1299, opened 2025-04-01
Fix the position range of the slider knob when repositioning the slider

Gato-X:
When a Slider is re-positioned, the position range of the knob is not updated. This fixes that issue.


benmoran56:
Hi Gato, thank you very much for the pull request. I made one small comment.

benmoran56:
Thank you, this will be in the next point release.

Issue #1309, opened 2025-04-26
MouseStateHandler mouse positions stored in a dictionary.


RealTactic:
Hi I have a very simple addition to the code for the MouseStateHandler class in "pyglet/window/mouse.py".

Currently the position data is stored as follows:

```python
    def __init__(self) -> None:  # noqa: D107
        self.data = {
            'x': 0,
            'y': 0,
        }
``` 

I want to add the following to this:
```python
    def __init__(self) -> None:  # noqa: D107
        self.data = {
            'x': 0,
            'y': 0,
        }
        self.x = 0
        self.y = 0
``` 

and update it in a way where code written previously will still work the same:
```python
    def on_mouse_motion(self, x: int, y: int, dx: int, dy: int) -> None:  # noqa: ARG002
        self.data['x'] = x
        self.data['y'] = y
        self.x = x
        self.y = y

    def on_mouse_drag(self, x: int, y: int, dx: int, dy: int, buttons: int, modifiers: int) -> None:  # noqa: ARG002
        self.data['x'] = x
        self.data['y'] = y
        self.x = x
        self.y = y
``` 

This will give us another "easier/cleaner" option to retrieve mouse coordinates when using MouseStateHandler class.

```python
win = pyglet.window.Window()
mouse_state = mouse.MouseStateHandler()
win.push_handlers(mouse_state)
#previously
mouse_position = (mouse_state["x"], mouse_state["y"])

# instead use
mouse_position = (mouse_state.x, mouse_state.y)
``` 

Just checking to see if there are any issues with this before submitting a PR.

benmoran56:
Hi,

Rather than adding the duplicate setting code and x/y attributes `__getattr__` could be added to do this. The code is identical to `__getitem__`. Give this a try (before creating an instance) to see if it accomplishes what you need: 

```py
MouseStateHandler.__getattr__ = MouseStateHandler.__getitem__
```

If that works, you could make a PR like: 
```py
    def __getattr__(self, item: str) -> int:
        return self.data[item]
```

PR #1311, opened 2025-04-28
Add x and y attributes to MouseStateHandler with __getattr__

RealTactic:
Please see issue #1309 for more info.

I have added the `__getattr__` function to the class and have tested to see that it does what I want as described in the issue.
I have also updated the documentation in the docstrings for that class.

### DIFF
--- a/pyglet/window/mouse.py
+++ b/pyglet/window/mouse.py
@@ -23,7 +23,8 @@ class MouseStateHandler:
         False
 
 
-    Mouse coordinates can be retrieved by using the ``'x'`` and ``'y'`` strings.
+    Mouse coordinates can be retrieved by using the ``'x'`` and ``'y'`` strings
+    or by using ``'mouse_state.x'`` and ``'mouse_state.y`` attributes
 
     For example::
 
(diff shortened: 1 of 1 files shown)


pyglet/window/mouse.py
```diff
@@ -64,6 +71,9 @@ def on_mouse_drag(self, x: int, y: int, dx: int, dy: int, buttons: int, modifier
 
     def __getitem__(self, key: str) -> int | bool:
         return self.data.get(key, False)
+    
+    def __getattr__(self, item: str) -> int:
+        return self.data.get(item, default=0)
```
benmoran56, line 76: It may be better to avoid .get, and just use:
```py
    def __getattr__(self, item: str) -> int:
        return self.data[item]
```
Otherwise it will return '0' when querying _any_ unrecognized attribute, like `mouse.q`.

benmoran56 (changes requested):

benmoran56 (approved):

benmoran56:
Thanks! This will go into the next point release.

PR #1313, opened 2025-05-06
Pyglet Clipboard Support

csd4ni3l:
This Draft adds clipboard support to pyglet.

First, it adds the `on_clipboard_copy` and `on_clipboard_paste` events to `Caret`.

Then, the `Caret` is now an `EventDispatcher`, and is registered for these events. There is a new optional window argument, which sets _window and will be used to set the clipboard content.

In `on_text_motion`, if the motion is `MOTION_COPY`, it will dispatch `on_clipboard_copy`, and will copy the text.

If the motion is `MOTION_PASTE`, it will paste the text inside the layout, overriding if something else was selected, and then dispatching `on_clipboard_paste` to the window that was passed in.

Please, feel free to correct my code or even push your local changes if you want to.

### DIFF
--- a/pyglet/text/caret.py
+++ b/pyglet/text/caret.py
@@ -18,16 +18,17 @@ from typing import TYPE_CHECKING, Any, Pattern
 
 from pyglet import clock, event
 from pyglet.window import key
+from pyglet.event import EventDispatcher
 
 if TYPE_CHECKING:
     from pyglet.graphics import Batch
+    from pyglet.window import Window
     from pyglet.text.layout import IncrementalTextLayout
 
-
-class Caret:
+class Caret(EventDispatcher):
     """Visible text insertion marker for `pyglet.text.layout.IncrementalTextLayout`.
 
-    The caret is drawn as a single vertical bar at the document `position` 
+    The caret is drawn as a single vertical bar at the document `position`
     on a text layout object.  If ``mark`` is not None, it gives the unmoving
     end of the current text selection.  The visible text selection on the
     layout is updated along with ``mark`` and ``position``.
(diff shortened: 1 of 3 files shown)


pyglet/text/caret.py
```diff
@@ -110,6 +116,8 @@ def __init__(self, layout: IncrementalTextLayout, batch: Batch | None = None,
 
         self.visible = True
 
+        self._window = window
+        self.push_handlers(window)
```
caffeinepills, line 120: No longer need to push the window as a handler if the Caret is dispatching. (Also the value could be None, so this would fail regardless)

pyglet/window/__init__.py
```diff
@@ -874,6 +874,18 @@ def on_scale(self, scale: float, dpi: int) -> EVENT_HANDLE_STATE:
         during runtime.
         """
 
+    def on_clipboard_copy(self, text: str) -> EVENT_HANDLE_STATE:
```
caffeinepills, line 877: Move to the Caret.

pyglet/window/__init__.py
```diff
@@ -1754,6 +1770,8 @@ def on_text_motion_select(self, motion: int) -> EVENT_HANDLE_STATE:
 BaseWindow.register_event_type('on_file_drop')
 BaseWindow.register_event_type('on_draw')
 BaseWindow.register_event_type('on_refresh')
+BaseWindow.register_event_type('on_clipboard_copy')
```
caffeinepills, line 1773: Remove.

caffeinepills (changes requested):
Looks good, just needs cleaning up.

caffeinepills (approved):

caffeinepills (approved):

Issue #693, opened 2022-10-14
Transparent Window not working on Windows 11


fregapple:
Hi there! I see that in 1.5.20 you added experimental support for transparent windows / overlay. However, I can't find anything in the official docs on how to set it up!

I have this as my code:

```
import pyglet

window = pyglet.window.Window(style=pyglet.window.Window.WINDOW_STYLE_OVERLAY)

label = pyglet.text.Label('Hello, World',
                          font_name='Times New Roman',
                          font_size=36,
                          x=window.width//2, y=window.height//2,
                          anchor_x='center', anchor_y='center')

@window.event 
def on_draw():
    window.clear()
    label.draw()

pyglet.app.run()
```

However, it looks like this when run:

![image](https://user-images.githubusercontent.com/82219216/195791002-65e7a7f3-61ea-4f46-8790-bcea3f08c54c.png)

Is there some other command I need to make the background Transparent?

benmoran56:
Hi, 
There is a new window style argument. There is an example here that should work on Linux and Windows. (macOS support is not yet implemented)
https://github.com/pyglet/pyglet/blob/master/examples/window/transparent_window.py

fregapple:
Hey, yeah I just found it. Howerver it doesn't appear transparent:

![image](https://user-images.githubusercontent.com/82219216/195793309-3f860a31-3f46-4fb5-ae67-b66fdb5dea8e.png)

benmoran56:
If you're using Wayland on Linux, unfortunately that's not yet implemented either. It will run using XWayland.

fregapple:
Sorry, I am running Windows 11 Pro. I should have mentioned that

fregapple:
Using python ver - 3.9.13

fregapple:
I just trialed the example for the overlay window as well and comes back the same! Perhaps a windows 11 issue?

benmoran56:
Thanks for the information. Yes, it's likely that more work is necessary for Windows 11 support, as the existing code was tested on 10. Nobody is currently working on this, but let's leave this ticket open for now.

fregapple:
No worries! Look forward checking it out when it is working.

caffeinepills:
Could you try doing a `pyglet.gl.glClearColor(0.0, 0.0, 0.0, 0.0)` after window creation and see if that works?

fregapple:
Hi, after:
`window = Window(500, 500, style=Window.WINDOW_STYLE_TRANSPARENT)`

[...]

PR #1314, opened 2025-05-07
Fix window transparency for Windows OS.

caffeinepills:
Enforce an alpha_size on configs if using a transparent or overlay window.
Add missing DWM definition.
Tested on Windows 10 x64.

This will allow a transparent framebuffer without having to do NVIDIA GDI compatibility tricks mentioned here https://github.com/pyglet/pyglet/issues/693


shenjackyuanjie:
image

works well!

shenjackyuanjie:
image

works well on windows 10 also

Issue #1315, opened 2025-05-09
Some WAV files play at insane volumes.


DigiDuncan:
**Describe the bug**
Some WAV files (specifically, seemingly, those encoded in Adobe Audition) play at ear-splitting volumes. 

⚠️ I'm going to warn you now, if you're testing this, **please turn down your headphones**, it's genuinely dangerous volumes.

Reencoding the file (and by that I mean opening it in Audacity and saving it as a WAV again) produces a clean version.

**System Information:**
```
Platform
------------------------------------------------------------------------------
platform:  Windows-11-10.0.22621-SP0
release:   11
machine:   AMD64

Python
------------------------------------------------------------------------------
implementation: CPython
sys.version: 3.13.2 | packaged by Anaconda, Inc. | (main, Feb  6 2025, 18:49:14) [MSC v.1929 64 bit (AMD64)]
sys.maxint: 9223372036854775807
os.getcwd(): C:\Users\DigiDuncan\Documents\GitHub\GDG_jam_2025-1

pyglet
------------------------------------------------------------------------------
pyglet.version: 2.1.5
pyglet.compat_platform: win32
pyglet.__file__: C:\Users\DigiDuncan\miniconda3\envs\jam\Lib\site-packages\pyglet\__init__.py
pyglet.options['audio'] = ('xaudio2', 'directsound', 'openal', 'pulse', 'silent')
pyglet.options['debug_font'] = False
pyglet.options['debug_gl'] = True
pyglet.options['debug_gl_trace'] = False
pyglet.options['debug_gl_trace_args'] = False
pyglet.options['debug_gl_shaders'] = False
pyglet.options['debug_graphics_batch'] = False
pyglet.options['debug_lib'] = False
pyglet.options['debug_media'] = False
pyglet.options['debug_texture'] = False
pyglet.options['debug_trace'] = False
pyglet.options['debug_trace_args'] = False
pyglet.options['debug_trace_depth'] = 1
pyglet.options['debug_trace_flush'] = True
pyglet.options['debug_com'] = False
pyglet.options['debug_win32'] = False
pyglet.options['debug_input'] = False
pyglet.options['debug_x11'] = False
pyglet.options['shadow_window'] = True
pyglet.options['vsync'] = None
pyglet.options['xsync'] = True
pyglet.options['xlib_fullscreen_override_redirect'] = False
pyglet.options['search_local_libs'] = True
pyglet.options['win32_gdi_font'] = False
[...]

PR #1316, opened 2025-05-09
Fix WMFDecoder delivering audio data as float (#1315)

Square789:
- Do not consider float data to be uncompressed [EDIT: bit misleading, of course it's still "uncompressed", but indigestable by pyglet], change the debug messages accordingly
- Additionally safeguard `WAVEFORMATEX` structs (this might cause breakage in isolated cases, but worth it to prevent further ear-splitting audio incidents).

### DIFF
--- a/pyglet/media/codecs/ffmpeg.py
+++ b/pyglet/media/codecs/ffmpeg.py
@@ -385,16 +385,6 @@ def ffmpeg_stream_info(file: FFmpegFile, stream_index: int) -> StreamAudioInfo |
             context.sample_rate,
             channel_count,
         )
-        if context.format in (AV_SAMPLE_FMT_U8, AV_SAMPLE_FMT_U8P):
-            info.sample_bits = 8
-        elif context.format in (AV_SAMPLE_FMT_S16, AV_SAMPLE_FMT_S16P,
-                                AV_SAMPLE_FMT_FLT, AV_SAMPLE_FMT_FLTP):
-            info.sample_bits = 16
-        elif context.format in (AV_SAMPLE_FMT_S32, AV_SAMPLE_FMT_S32P):
-            info.sample_bits = 32
-        else:
-            info.sample_format = None
-            info.sample_bits = None
     else:
         return None
     return info
@@ -634,35 +624,29 @@ class FFmpegSource(StreamingSource):
                 self._video_stream = stream
                 self._video_stream_index = i
 
-            elif isinstance(info, StreamAudioInfo) and info.sample_bits in (8, 16, 24) and self._audio_stream is None:
+            elif isinstance(info, StreamAudioInfo) and self._audio_stream is None:
                 stream = ffmpeg_open_stream(self._file, i)
 
+                channels_out = min(2, info.channels)
+                channel_input = self._get_default_channel_layout(info.channels)
+                channel_output = self._get_default_channel_layout(channels_out)
+
+                sample_bits = info.sample_bits
+                if info.sample_format in (AV_SAMPLE_FMT_U8, AV_SAMPLE_FMT_U8P):
+                    self.tgt_format = AV_SAMPLE_FMT_U8
+                else:
+                    # No matter the input format, produce S16 samples.
+                    sample_bits = 16
+                    self.tgt_format = AV_SAMPLE_FMT_S16
+
                 self.audio_format = AudioFormat(
-                    channels=min(2, info.channels),
-                    sample_size=info.sample_bits,
+                    channels=channels_out,
+                    sample_size=sample_bits,
                     sample_rate=info.sample_rate)
                 self._audio_stream = stream
                 self._audio_stream_index = i
 
-                channel_input = self._get_default_channel_layout(info.channels)
-                channels_out = min(2, info.channels)
-                channel_output = self._get_default_channel_layout(channels_out)
-
-                sample_rate = stream.codec_context.contents.sample_rate
-                sample_format = stream.codec_context.contents.sample_fmt
-
-                if sample_format in (AV_SAMPLE_FMT_U8, AV_SAMPLE_FMT_U8P):
-                    self.tgt_format = AV_SAMPLE_FMT_U8
-                elif sample_format in (AV_SAMPLE_FMT_S16, AV_SAMPLE_FMT_S16P):
-                    self.tgt_format = AV_SAMPLE_FMT_S16
-                elif sample_format in (AV_SAMPLE_FMT_S32, AV_SAMPLE_FMT_S32P):
-                    self.tgt_format = AV_SAMPLE_FMT_S32
-                elif sample_format in (AV_SAMPLE_FMT_FLT, AV_SAMPLE_FMT_FLTP):
-                    self.tgt_format = AV_SAMPLE_FMT_S16
-                else:
-                    raise FFmpegException('Audio format not supported.')
-
-                self.audio_convert_ctx = self.get_formatted_swr_context(channel_output, sample_rate, channel_input, sample_format)
+                self.audio_convert_ctx = self.get_formatted_swr_context(channel_output, info.sample_rate, channel_input, info.sample_format)
                 if not self.audio_convert_ctx:
                     swresample.swr_free(self.audio_convert_ctx)
                     raise FFmpegException('Cannot create sample rate converter.')
--- a/pyglet/media/codecs/wave.py
+++ b/pyglet/media/codecs/wave.py
@@ -25,6 +25,9 @@ class WaveSource(StreamingSource):
 
         nchannels, sampwidth, framerate, nframes, comptype, compname = self._wave.getparams()
 
+        if nchannels not in (1, 2):
+            raise WAVEDecodeException(f"incompatible channel count {nchannels}")
+
         if sampwidth not in (1, 2):
             raise WAVEDecodeException(f"incompatible sample width {sampwidth}")
 
--- a/pyglet/media/codecs/wave.py
+++ b/pyglet/media/codecs/wave.py
@@ -25,6 +25,9 @@ class WaveSource(StreamingSource):
 
         nchannels, sampwidth, framerate, nframes, comptype, compname = self._wave.getparams()
 
+        if sampwidth not in (1, 2):
+            raise WAVEDecodeException(f"incompatible sample width {sampwidth}")
+
         self.audio_format = AudioFormat(channels=nchannels, sample_size=sampwidth * 8, sample_rate=framerate)
 
         self._bytes_per_frame = nchannels * sampwidth
--- a/pyglet/media/codecs/wmf.py
+++ b/pyglet/media/codecs/wmf.py
@@ -542,8 +542,13 @@ class WMFSource(Source):
             source_subtype_guid = com.GUID(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
             source_sample_size = c_uint32()
             source_channel_count = c_uint32()
+
             imfmedia.GetGUID(MF_MT_SUBTYPE, byref(source_subtype_guid))
-            imfmedia.GetUINT32(MF_MT_AUDIO_BITS_PER_SAMPLE, byref(source_sample_size))
+            try:
+                # Some formats such as mp3 do not report this value
+                imfmedia.GetUINT32(MF_MT_AUDIO_BITS_PER_SAMPLE, byref(source_sample_size))
+            except OSError:
+                source_sample_size.value = 0
             imfmedia.GetUINT32(MF_MT_AUDIO_NUM_CHANNELS, byref(source_channel_count))
 
             if (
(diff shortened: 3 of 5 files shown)


caffeinepills:
It actually looks like this should work, I think the issue is this:
https://github.com/pyglet/pyglet/blob/9bf89395254b90a3f2cd5b045a60d5251dbf9fed/pyglet/media/drivers/xaudio2/interface.py#L29-L37

I am forcing it to `WAVE_FORMAT_PCM` when the format is not guaranteed to be PCM.

benmoran56:
@caffeinepills  Could we do a small mapper to find the right values? Something like?
```py
wfx.wFormatTag = {32: WAVE_FORMAT_IEEE_FLOAT, 16: WAVE_FORMAT_PCM}[audio_format.sample_size]
```

Square789:
> It actually looks like this should work, I think the issue is this:
> 
> https://github.com/pyglet/pyglet/blob/9bf89395254b90a3f2cd5b045a60d5251dbf9fed/pyglet/media/drivers/xaudio2/interface.py#L29-L37
> 
> I am forcing it to `WAVE_FORMAT_PCM` when the format is not guaranteed to be PCM.
> 
> If I hardcode `wfx.wFormatTag = lib.WAVE_FORMAT_PCM` to `wfx.wFormatTag = 3` (which is `WAVE_FORMAT_IEEE_FLOAT`, from the source `WAVEFORMATEX`) then it does not distort.

Doing that instructs the voice to play float samples, which causes voice initialization to fail when playing the "healthy" sound

Square789:
> @caffeinepills Could we do a small mapper to find the right values? Something like?
> 
> ```python
> wfx.wFormatTag = {32: WAVE_FORMAT_IEEE_FLOAT, 16: WAVE_FORMAT_PCM}[audio_format.sample_size]
> ```
> 
> A simple lookup table to cover 8, 16, 24, and 32 of course. If you think it's worth supporting 32bit audio. I think it's pretty rare, just due to how huge those files can be.

pyglet atm is hardwired to a somewhat common standard (At least going by (XAudio2)[https://learn.microsoft.com/en-us/windows/win32/api/xaudio2/nf-xaudio2-ixaudio2-createsourcevoice] and (OpenAL)[https://github.com/kcat/openal-soft/blob/98d2afe8522db77048e3041c342063a9c29c4704/include/AL/al.h#L380]) of playing back samples in u8 or s16 format, only 1 or 2 channels supported. Testing for a sample size of 32 is technically not enough to be sure we are dealing with float samples (There appear to be more formats that require the [WAVEFORMATEXTENSIBLE](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ksmedia/ns-ksmedia-waveformatextensible#remarks) structure because they only use 20 or 24 of a given 32 bits) (Additionally, there seem to be very rare cases where 16bit floats are used, so allowing both PCM and Float might cause confusion there). 32bit audio probably delivers higher quality, so i can understand why it should be supported, but to be safe i would want to add another attribute to `AudioFormat` specifying whether we are dealing with PCM or Float, or outright create generalized formats that can be handled by all audio backends.

Square789:
For reference: https://github.com/pyglet/pyglet/issues/1315#issuecomment-2868021391
My intention with this PR is to make the WMFDecoder similar to the FFMpegDecoder, both resampling f32 to u16.
Although it does happen implicitly, as 
```py
mf_mediatype.SetGUID(MF_MT_MAJOR_TYPE, MFMediaType_Audio)
mf_mediatype.SetGUID(MF_MT_SUBTYPE, MFAudioFormat_PCM)
```
does not specify the sample size (which can probably be done through `[MF_MT_AUDIO_BITS_PER_SAMPLE` as per https://learn.microsoft.com/en-us/windows/win32/medfound/uncompressed-audio-media-types). 16 bits apparently just happens to be the default.

benmoran56:
Thanks for the info, that makes sense.  I vaguely remember running into what you are describing a while ago, with trying to support 24bit Wav files. 
We should probably just be defaulting to the built-in/standard library decoder on Windows for .wav files as well (fraction less resource usage). That would also help enforce the baseline.

caffeinepills:
Why not try to convert to 32 bit float via:
```
mf_mediatype.SetGUID(MF_MT_MAJOR_TYPE, MFMediaType_Audio)
mf_mediatype.SetGUID(MF_MT_SUBTYPE, MFAudioFormat_Float)
```
If that fails, fallback to PCM?

Square789:
Unless updated too, it would render OpenAL inoperable on Windows (which admittably is very unlikely to be used there) and amend pyglet's decoder contract by "On Windows, decoders may return float samples", which i guess is also fine for the WMFDecoder.
Still, creating a platform discrepancy like that and assuming that 32bit <=> float is flaky. According to the XAudio2 docs i linked, 32bit size PCM samples also exist.
A good solution would be to create more certainty on the data by including sample types in the audio format that can then with certainly be resolved to e.g. `PA_SAMPLE_S16LE`, `AL_FORMAT_{STEREO|MONO}16` and `WAVE_FORMAT_PCM` + `16`.
Before something like that, imo it's safest to stick to "8 bits means u8" and "16 bits means s16" and enforce that in the decoders.
For OpenAL playing back floats requires an extension and may not be entirely guaranteed. I am not sure just how commonplace openal-soft is.

caffeinepills:
> Unless updated too, it would render OpenAL inoperable on Windows (which admittably is very unlikely to be used there) and amend pyglet's decoder contract by "On Windows, decoders may return float samples", which i guess is also fine for the WMFDecoder. Still, creating a platform discrepancy like that and assuming that 32bit <=> float is flaky. According to the XAudio2 docs i linked, 32bit size PCM samples also exist. A good solution would be to create more certainty on the data by including sample types in the audio format that can then with certainly be resolved to e.g. `PA_SAMPLE_S16LE`, `AL_FORMAT_{STEREO|MONO}16` and `WAVE_FORMAT_PCM` + `16`. Before something like that, imo it's safest to stick to "8 bits means u8" and "16 bits means s16" and enforce that in the decoders. For OpenAL playing back floats requires an extension and may not be entirely guaranteed. I am not sure just how commonplace openal-soft is.

We could query OpenAL for the support. However, in these cases, perhaps we need to have the drivers define which data types it supports, and can effectively use the `AudioFormat` correctly, and perhaps the codecs can optimize based on that?

In the worst case (for WMF anyways), we can always try to convert Float to PCM. If we drop the ` or guid_compressed == MFAudioFormat_Float`, it should put WMF on to the path to convert itself to PCM. 
With testing his example, the `WAVEFORMATEX` after does convert down: `WAVEFORMATEX(wFormatTag=1, nChannels=2, nSamplesPerSec=44100, nAvgBytesPersec=176400, nBlockAlign=4, wBitsPerSample=16, cbSize=0)`. The `wFormatTag` became 1.  If you want to add an exception to ensure the format is always PCM after converting, then that's quick fix. 

However I am unsure about this, as this silently lowering their audio quality. (On the other hand, at least it's not crashing their app or preventing them from using their audio), so perhaps a warning should be placed when it downsamples?

Square789:
Created a few more files that should cover most of the wav formats and current behavior:

|   | PCM Integer, 8 bit: u8 | PCM Integer, 16 bit: s16 | PCM Integer, 24 bit: s24 | PCM Integer, 32 bit: s32 | PCM Integer, 64 bit: s64 | IEEE754 Float, 32 bit: f32 | IEEE754 Float, 64 bit: f64 |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Used decoder | wave | wave | wave¹ | wave¹ | wave¹ | WMF/FFmpeg | WMF/FFmpeg |
| Pre-PR, XAudio2 | Ok | Ok | Ok² | Ok² | WinError when creating Voice | Painful sound | WinError when creating Voice |
| Pre-PR, OpenAL | Ok | Ok | MediaException, previously no AudioFormat | MediaException, previously no AudioFormat | MediaException, previously no AudioFormat | Downsampled to s16 | MediaException, previously no AudioFormat |
| With this PR, XAudio2 | Ok | Ok | MediaException | MediaException | MediaException | Downsampled to s16, implicitly | Downsampled to s16, implicitly |

¹ Since Python 3.12. This is a funny edge case, as, outside of Windows, s24 or s32 files that previously may have been decoded by FFmpeg (albeit without an AudioFormat) will now produce an AudioFormat and then an error during OpenAL format selection. Not too interesting, nothing is getting played either way.
² We only use the `WAVEFORMATEX` struct (with the format tag `WAVE_FORMAT_PCM`). The Xaudio2 docs state: "If wFormatTag is WAVE_FORMAT_PCM, then wBitsPerSample should be equal to 8 or 16.". This type of data should not be returned from the decoder and works with XAudio2 unofficially.

The FFmpeg decoder handles a more limited amount of sample sizes and simply provides an `AudioFormat` of `None` if it encounters an unsupported format. It should be easy to alter it to also downsample 24bit, integer 32bit and 64bit audio.

In reference to ², this PR breaks playability of Integer PCM files that would actually play before, if the sample size is >16 bit. Officially this working is not guaranteed, but i can undo the AudioFormat guards.

I also noticed that floats appear to be the go-to sample format in WebAudio, which is relevant for pyglet 3.0 and worth considering. The only problem with supporting it is that OpenAL can't be entirely guaranteed to have the `AL_EXT_float32` extension, so if it were to be supported, pyglet's requirements need to be updated to mention that the extension is required. That change is probably not a good idea inbetween minor versions.

> so perhaps a warning should be placed when it downsamples

Pyglet states (although in a poor place, the [`AudioFormat` docstring](https://github.com/pyglet/pyglet/blob/14a03d72943b324b8934f8dc344138e050201859/pyglet/media/codecs/base.py#L26) only) and assumes that it only ever handles samples in the u8 and s16 formats, a limit primarily given by the extensionless OpenAL support range. The downsampling not happening is, in my eyes, an oversight in both the wave and WMF decoder.

My primary intention with this PR was to fix the very loud and distorted result from interpreting float audio as integer PCM in Xaudio2. But seeing that it also breaks some other formats as visible in the table, now i'm not sure what to do:

1. Decide that raising the `MediaException`s is correct and accept the PR as-is (fixing the decoders seperately and keeping the formats to u8, s16 for now.)
2. Downsample float audio to s16 only and keep the wave+WMF grey zones intact (undo the XAudio2+DirectSound format checks)
3. Downsample float audio to s16, change the WaveDecoder to raise a MediaError when >16bit, so the FFmpegDecoder takes over again
4. Change the WMFDecoder so it will downsample all Integer PCM formats >16bit to 16bit
5. Expand the FFmpegDecoder to downsample s24, s32, f64 and s64
6. Modify all audio backends to support at least f32 right now (maybe s24, s32 as well), taking OpenAL extension presence for granted. f64 and s64 can not be played back by any backend (and are also complete overkill), so will still need to be downsampled or otherwise protected against.

Square789:
After a discussion with @benmoran56, what i'm trying to do now is to have decoders be mindful of the data they return, keeping it u8, s16 and at 1 or 2 channels. Supposedly s16 is fine for normal audio playback, anything higher is audible only in specific conditions: [1](https://www.reddit.com/r/audioengineering/comments/1c68k2/what_exactly_is_the_difference_between_16bit/), [2](https://www.youtube.com/watch?v=n3hFVgBqTr4)
So far, only the wave, FFmpeg and WMF decoders check for both channel count and sample size.
There was another idea on making AudioDrivers report formats they are capable of playing back, which decoders can then respect, but so far that is not implemented. Should it be implemented in this PR?

These commits make the WMFDecoder very likely to fall into the resampling/decoding branch (as the specific conditions it tests for are covered by the standard wave decoder and i'm not sure another format exists that produces the `MFAudioFormat_PCM` subtype.)
The WMFDecoder is now probably too verbose in its target format description. @caffeinepills let me know whether this is fine.

The FFMpegDecoder would always apply a resampler anyways and now always resamples to s16, unless the decoded data is u8. It also had some limits on formats and sample sizes, which have been removed.

The wave decoder now raises an exception when encountering a channel count `not in (1, 2)` or a sample width (in bytes) `not in (1, 2)`, to have a decoder with resampling capabilities take over.

benmoran56:
> There was another idea on making AudioDrivers report formats they are capable of playing back, which decoders can then respect, but so far that is not implemented. Should it be implemented in this PR?

There was the idea that we can add maximum bitrate (and potential other) arguments to the MediaDecoder class inits. This can be a solution for users that know their audio driver (backend) will support Source data in that format. It will also require a user to specifically make an instance of the Decoder, so would apply to people who are targeting a specific platform and formats.

IMO, lets leave this to a next PR so as not to over complicate things.

benmoran56 (approved):

fladd:
@benmoran56 @caffeinepills @Square789

Mmh, I think there is a better way of handling this. Pyglet and all of its supported audio drivers are perfectly capable of playing 32 bit int and float formats (well, OpenAL needs extensions for this, but at least the float32 extension is usually present, including on macos). Note that 24-bit audio is most often put into a 32bit int container with the least significant bits being padded with zeros. So this is handled by 32bit int support as well. What is needed is indeed a way to differentiate between 32 int and 32 float within Pyglet. One simple way to do this is to give the `audio_format` definition a property called `sample_type` and then include this in the logic to select the right output format given the input format. In the case one of the 32bit format is not available (usually int for OpenAL), one can convert to the other (usually float for OpenAL) without quality loss. Only in case both are not available (probably OpenAL on Windows), one should convert down to 16bit, but this should optimally involve dithering with noise-shaping in order to retain most of the dynamic range of 24/32 bit.

I have implemented all of this (but only for FFmpeg) in my Zipped Album Player, where I (1) detect which of Pyglet's drivers are available on a given system, (2) monkey patch all of them to support 32bit int and float (if available), (3) make each driver know the difference between int float representation, and (4) subclass FFmpegSource to deal with all of this correctly. I also gave FFmpegSource a class variable that can be set from the outside to always force a certain output format irrespective of the input (like you do now forcing everything to 16bit) if someone wishes to do that.

Have a look at https://github.com/zipped-album/zap/blob/main/zap/player.py for details. If this is something that you are interested in having in Pyglet natively, I am happy to send a pull request. Just let me know.

benmoran56:
Hi @fladd, I pinged you on the Discord server, but we can continue the discussion here if you prefer.

PR #1320, opened 2025-05-22
Add minimum window size to Windows and MacOS

csd4ni3l:
I already added minimum window size for Linux, but i didnt to Windows and MacOS. This fixes that by using a minimum of 1 when setting width and height. (I also got a report from someone on Windows having a crash because there was no minimum of 1)

### DIFF
--- a/pyglet/window/win32/__init__.py
+++ b/pyglet/window/win32/__init__.py
@@ -1197,7 +1197,7 @@ class Win32Window(BaseWindow):
             self.dispatch_event('on_show')
         w, h = self._get_location(lParam)
         if not self._fullscreen:
-            self._width, self._height = w, h
+            self._width, self._height = max(1, w), max(1, h)
         self._update_view_location(self._width, self._height)
 
         if self._exclusive_mouse:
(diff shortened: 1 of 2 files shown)


pyglet/window/win32/__init__.py
```diff
@@ -1197,7 +1197,7 @@ def _event_size(self, msg: int, wParam: int, lParam: int) -> int | None:
             self.dispatch_event('on_show')
         w, h = self._get_location(lParam)
         if not self._fullscreen:
-            self._width, self._height = w, h
+            self._width, self._height = max(1, w), max(1, h)
```
caffeinepills, line 1200: This is modifying the event from Windows. We should be limiting the size on what the user is capable of setting.

caffeinepills (changes requested):

pyglet/window/win32/__init__.py
csd4ni3l: The problem is that if they resize the window too much(i dont know how you can to 0, never tried), but one user who reported to me could, and thats why i tried to set the maximum here as well

PR #1326, opened 2025-06-02
Improve documentation of example camera class' end method

Brett208:
We successfully used the camera example as a springboard, so thank you for providing it.

While reading through the end method in the example, I thought it read oddly although the gist of the existing code comments are correct. Took a turn a trying to improve the language.

I understand if you don't want to make the change. Thank you for making pyglet available.

~Brett


benmoran56:
Thank you Brett,  documentation changes are always appreciated!  This reads well to me.

PR #1344, opened 2025-08-06
Added the option for windows to be minimized upon creation

TaqiHamoda:
I am currently using `pyglet` for rendering and saving scenes programmatically and I needed the option to have the window minimized on creation as it was being very disruptive to my workflow.

These changes were tested on Ubuntu 24.


benmoran56:
Hi, thanks for opening the PR.
Unfortunately after some discussion, we have to decline this. The reason being:
The same thing can be accomplished with a subclass (or a single call to minimize after window creation). 
You can already create non-visable (or headless, depending on platform) Windows for programmatic workflows. 

We're happy to receive PRs, but not everything fits the scope of the project. I would encourage you to join the Discord server if you are interested in contributing, as it's a good place to discuss ideas like this.

PR #1358, opened 2025-09-16
Linux platform event loop:  switch from select/poll to epoll & timerfd.

benmoran56:


benmoran56:
Some additional work is needed to prevent failure during the unit tests for multiple start/stop of the event loop.

benmoran56:
This is a working experiment locally, but will possibly abandon this for now since some of the special FDs are only available in more recent Python releases.

PR #1362, opened 2025-09-24
replacing the list implementation of window's event queue by a double ended queue

gergely-elias:
I mentioned this a few days ago on discord, didn't get too much echo.
If there's any good reason to decline this PR, please feel free to do so.

In my case, this change does fix a frozen application. In case the event queue grows too large, list.pop(0) is very inefficient, leading to performance problems.
One can argue that the event queue growing too large is the result of a poor architecture on my side (I'm mixing a pyglet main window with a tkinter dialog, this happens when the tkinter dialog is open), and that the problem can be worked around (I currently do it with setting the pyglet window's visibility temporarily to False), or even properly solved (on discord I got detailed instructions about using a pyglet-based dialog window), so maybe this update is not needed. I can't see why it would hurt though.
But I don't know what other consequences the change can cause.

Thanks for taking this into consideration!

### DIFF
--- a/pyglet/window/__init__.py
+++ b/pyglet/window/__init__.py
@@ -91,6 +91,7 @@ from __future__ import annotations
 
 import sys
 from abc import abstractmethod
+from collections import deque
 from typing import TYPE_CHECKING, Any, Callable, Sequence
 
 import pyglet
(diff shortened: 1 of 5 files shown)


benmoran56 (approved):

benmoran56:
Thanks for opening the PR. I think this is safe to merge. For typical pyglet apps/games where the event loop is not stalled, this event queue is unlikely to ever contain more than a few items.  Still, even though the deque object technically could be slower (at tiny sizes) due to small built-in overhead, it seems immeasurable in my tests.

PR #1369, opened 2025-10-17
Properly type MouseStateHandler.

nicolasdespres:
Key can be str (for coordinate) and int (for button)


benmoran56 (approved):

benmoran56:
Thanks for this

PR #1373, opened 2025-10-19
Fix camera group example

nicolasdespres:


benmoran56 (approved):

benmoran56:
Thanks @nicolasdespres!

Issue #1298, opened 2025-03-29
pyglet.text.decode_attributed not works with non-system fonts


skilletfun:
**Describe the bug**

First I load font and check it available:

```
pyglet.font.add_file("./assets/fonts/Agency FB.ttf")
pyglet.font.load("Agency FB Cyrillic", 16)
print(f"Agency FB Cyrillic exists: {pyglet.font.have_font('Agency FB Cyrillic')}")
```

Print returns `True` in console, so font are available
```
> python main.py 
Agency FB Cyrillic exists: True
```

Then I create two `DocumentLabel` with `decode_html` and `decode_attributed` 
```
text1 = pyglet.text.decode_html("<font size='5' face='Agency FB Cyrillic'>Hello, html Agency!</font>")
text2 = pyglet.text.decode_attributed("{font_name 'Agency FB Cyrillic'}Hello, attributed Agency!")
```

And after run I get this
![Image](https://github.com/user-attachments/assets/26d84d9e-413a-4220-a050-42064dcf2093)

But if I remove line with `add_file` and move font file into system folder with fonts (`.fonts` in home directory) I got correct result

Edited code
```
# pyglet.font.add_file("./assets/fonts/Agency FB.ttf")  # Commented line
pyglet.font.load("Agency FB Cyrillic", 16)
print(f"Agency FB Cyrillic exists: {pyglet.font.have_font('Agency FB Cyrillic')}")
```
Print also returns `True`
And right label with correct font
![Image](https://github.com/user-attachments/assets/2fce7302-a6f2-41bf-87a8-e83ec2693b22)

So can `decode_attributed` be used with loaded by `add_file`, non-system fonts?

I tried several fonts, so it not depends of font in example above
No any errors in console happens while try this

**System Information:**
Python 3.11.1, Pyglet 2.1.3
Manjaro Linux, Kernel 6.1.119

**How To Reproduce**
Here full code

```
import pyglet

img = pyglet.image.SolidColorImagePattern((255,255,255,255)).create_image(1280, 720)
window = pyglet.window.Window(1280, 720)

pyglet.font.add_file("./assets/fonts/Agency FB.ttf")
pyglet.font.load("Agency FB Cyrillic", 16)
print(f"Agency FB Cyrillic exists: {pyglet.font.have_font('Agency FB Cyrillic')}")

text1 = pyglet.text.decode_html("<font size='5' face='Agency FB Cyrillic'>Hello, html Agency!</font>")
[...]

PR #1379, opened 2025-10-28
Fix issue #1298. 

caffeinepills:
Font objects should no longer accept None for some of it's parameters. So this ensures attribute text styles pass the correct types.

Seems to have affected the Linux FreeType font loader the most as it prevented attributed font names being loaded. Tested using WSL.


benmoran56 (approved):

benmoran56:
That looks good to me.

PR #1399, opened 2025-12-27
Fix multi-draw array call by adding mode parameter

eruvanos:
Following the docs, the mode parameter was missing:
https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_multi_draw/multiDrawArraysWEBGL

Caused error when used:

image


caffeinepills:
Looks good. Seems to be an oversight due to no big tests for multi-draw.

PR #1400, opened 2025-12-27
Add context_options for Pyodide rendering contexts

Cleptomania:
This adds a new `context_options` field to the Pyodide options in pyglet, which allows an application to set values in a dictionary that will be passed to the `canvas.getContext` function when the webgl2 context is created.

This allows applications to do things like turn off pre-multiplied alpha, or turn off alpha on the back-buffer.

As a sidenote, Pyglet may want to explore setting `alpha` to False in these options, as it's much more in line with how desktop OpenGL handles the backbuffer. With default WebGL settings if you do any alpha blending as you normally would without pre-multiplied alpha you will get the color of the page behind the canvas leaking through the alpha blending which leads to things like edge artifacts.

### DIFF
--- a/pyglet/__init__.py
+++ b/pyglet/__init__.py
@@ -54,6 +54,12 @@ class PyodideOptions:
     page, and do not want to alter your code, then modify this option.
     """
 
+    context_options: dict[str, Any] = field(default_factory=lambda: {})  # noqa: PIE807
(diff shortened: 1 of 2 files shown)


benmoran56:
@Cleptomania since we now have a dedicated WebGLConfig, we can close this. Any objections?

PR #1405, opened 2026-01-26
bugfix: fixing rotation behaviour of Sector

igt-tsteiner:
The rotation of Sector is used when creating vertices in _get_vertices method. This has the effect of doubling the drawn rotation when e.g. setting radius, angle, etc. 

__contains__ method mixes deg and rad when comparing angles, and does not account for negative start_angle and/or negative angle.


benmoran56:
Thank you for the PR. This looks good.

igt-tsteiner:
Thank you! Cheers, Thomas

PR #1407, opened 2026-02-04
Label Class's __init__ font_family type hinting update

Brett208:
Update Label.__init__'s font_family attribute type hint to indicate it can optionally take a font family in the form of a list.

It appears the underlying code supports receiving a font family as a list of strings, but the type hint does not allow it.

Thanks for maintaining pyglet!

~Brett


benmoran56:
Thank you @Brett208

PR #1409, opened 2026-02-13
Provide concrete typing for Window based on selected platform.

Brett208:
Currently, static type checkers are unable to deduce the concrete type used to represent `BaseWindow` and will provide errors similar to below (this is from mypy).

> error: Cannot instantiate abstract class "BaseWindow" with abstract attributes "_create", "_recreate", ... and "switch_to" (8 methods suppressed)  [abstract]

The code that produced this mypy error was:

```python
self._window = Window(1280, 720)
```

I only tested on my development machine (Windows), so I did not fully verify type checking results on other platforms. For me it resolved to `XlibWindow`. I am not a expert on typing in Python, so hopefully this is an amenable solution.

Thanks!
~Brett

### DIFF
--- a/tests/unit/math/test_quat.py
+++ b/tests/unit/math/test_quat.py
@@ -11,16 +11,6 @@ def test_create():
     assert q.length() == 1
 
 
-def test_from_mat3():
-    with pytest.raises(NotImplementedError):
-        Quaternion.from_mat3()
-
-
-def test_from_mat4():
-    with pytest.raises(NotImplementedError):
-        Quaternion.from_mat4()
-
-
 def test_to_mat3():
     # TODO: Needs improvement. Just checking identity matrix.
     q = Quaternion()


benmoran56:
Hi @Brett208, thanks for opening the PR.  I believe we had some issues with this approach in the past, but also the typing situation in Python has continued to evolve since then. The CI pipeline does pass now :thinking: 

One thing that comes to mind however, is that we don't necessary have full typing information (or docstrings) in the platform subclasses.  Have you tried with something simple like? : 
```py
import BaseWindow as Window
```
If that doesn't work, perhaps a "safe" solution would be to make a dummy class just for the purposes of type checking. 
Just an idea, but something like: 
```py
if TYPE_CHECKING:
    ...
    class Window(BaseWindow):
        def missing_abstract_method_a():
             pass

        def missing_abstract_method_b():
             pass
```

This way the BaseWindow could remain the single point that has full typing and docstrings for everything. 
If you could also give me a small example of how you're testing with mypy, I can test it on my end.

Brett208:
@benmoran56,

I verified both of your suggested changes prevent mypy from complaining. I went with the import since it is smaller and would require less work to maintain if an abstract method ever changed (although I assume Window is very stable). It does seem odd to import a class at the top of the file it is defined in.

---

To recreate on your machine:

Define Requirements.txt as:
```
pyglet
mypy
```

Define main.py as:
```python
import pyglet

window = pyglet.window.Window(width=400, height=300)
pyglet.app.run()
```

run terminal command 'mypy .'

You should see the following mypy error:

> main.py:3: error: Cannot instantiate abstract class "BaseWindow" with abstract attributes "_create", "_recreate", ... and "switch_to" (8 methods suppressed)  [abstract]
> Found 1 error in 1 file (checked 1 source file)

Make the changes to Window's __init__.py and rerun 'mypy .' and the error should go away.

---

I am using VSCode. I receive documentation on hover both before and after the change. If I use the goto definition/declaration feature, VSCode will take me to the xlib implimentation of the method, not the abstract (same behavior before and after the change).

image

Realizing typing support can get messy...

~Brett

Brett208:
Forgot to mention, I was using a virtual environment with python 3.14.3 for the sample, although I am encountering the same issue on 3.12.x.

benmoran56:
I don't usually use VS Code, but I tried it to test the behavior. It's a bit funny, because every time I mouse-over `Window()` it gives me a random subclass.  The method mouse-overs _usually_ seem to correctly resolve to BaseWindow, but occasionally give a different subclass as well. I think that may be just a quirk with pylance :thinking: 

In any case, the change does fix mypy, so lets merge this in.

Brett208:
@benmoran56  Thanks for the quick review! :rocket:

PR #1412, opened 2026-02-25
Added Atari CX Wireless Controller configuration

Ralac07:
I set this up for the wireless joystick that came with the atari 2600+


benmoran56:
Hi @Ralac07, thanks for opening the ticket.  The built-in controllerdb in pyglet is automatically generated based on SDL sources, and shouldn't be manually edited.  The built-in db is limited to the most common devices, and somewhat slow to update. One solution is to include additional mappings with your game/app in the form of a text file, and import this at runtime. The most popular community sourced mappings are: https://github.com/mdqinc/SDL_GameControllerDB
I would recommend to open a PR there, to get the this controller included in that DB.

You can also just add the mapping as a string inside of your game/app, if you are only concerned with this particular device.

Ralac07:
Thank you, I had the mapping as a string in my project and thought I should try to figure out the best way to share the mapping
I'll open a pr there, thanks for the help

benmoran56:
Sure no worries.  That community mapping file is common to see in shipped games. It catches more devices, and also gives end users a way to update it themselves if they want to.

PR #1415, opened 2026-03-11
Better 32-bit audio support

fladd:
This PR implements more robust 32-bit audio support as suggested in #1316. Most importantly, it introduces unified audio sample format definitions and enables audio drivers to explicitly distinguish between 32-bit integer and floating-point formats.

**Note**: For now, the FFmpeg decoder is the primary implementation utilizing these new capabilities. It serves as a reference for adapting other decoders in the future and is designed not to break the current behavior of other drivers.

### Changes
- **`pyglet.media`**: Added `AUDIO_SAMPLE_FORMAT_U8`, `AUDIO_SAMPLE_FORMAT_S16`, `AUDIO_SAMPLE_FORMAT_S32`, `AUDIO_SAMPLE_FORMAT_F32` string constants
- **`media.load`**: Added argument `audio_sample_format` (defaults to `None`) to set a fixed target format (currently only supported by `FFmpegDecoder`)
- **`media.base.AudioFormat`**: Added `SAMPLE_TYPE_INT`, `SAMPLE_TYPE_UINT`, `SAMPLE_TYPE_FLOAT` string constants
- **`media.base.AudioFormat`**: Added argument `sample_type` (defaults to `SAMPLE_TYPE_INT`)
- **`media.base.AudioFormat`**: Added property `sample_format` which returns a `media.AUDIO_SAMPLE_FORMAT_*` string
- **`media.ffmpeg`**: Added `AUDIO_SAMPLE_FORMATS` dict that maps supported `media.AUDIO_SAMPLE_FORMAT_*` values to FFmpeg-specific `AV_SAMPLE_FMT_*` values
- **`FFmpeg decoder`**: Added `audio_sample_format` argument to `FFmpegDecoder.decode`, and `FFmpegSource.__init__`
- **`FFmpegSource`**: Added new default behaviour (`audio_sample_format=None`) to automatically set the target format based on the input format
- **`FFmpegSource`**: Added fallback mechanisms if automatic target format is not supported by current audio driver
- **`FFmpegSource`**: Added noise-shaping dither when target format bit-depth is lower than source format
- **`OpenAL` audio driver**: added support for playing 32-bit int and/or float formats, depending on available extensions (`AL_EXT_FLOAT32`, etc.)
- **All other audio drivers**: added support for playing 32-bit int and float formats (and also 24-bit)
- **All audio drivers**: added mappings from `media.AUDIO_SAMPLE_FORMAT_*` formats to driver-specific formats, and  added property `sample_formats` which returns the `media.AUDIO_SAMPLE_FORMAT_*` formats they support

### DIFF
--- a/pyglet/media/__init__.py
+++ b/pyglet/media/__init__.py
@@ -127,17 +127,17 @@ def load(filename: str, file: BinaryIO | None = None,
         audio_sample_format:
             A specific audio sample format you wish the decoder to ouput,
             rather than relying on automatic detection. For possible values see
-            the AUDIO_SAMPLE_FORMAT_* constants.
+            ``pyglet.media.AUDIO_SAMPLE_FORMAT_*`` constants.
             NOTE: currently only supported by FFmpegDecoder!
         audio_sample_rate:
             A specific audio sample rate (in Hz) you wish the decoder to
             output, rather than relying on automatic detection. For possible
-            values see the AUDIO_SAMPLE_RATE_* constants.
+            values see ``pyglet.media.AUDIO_SAMPLE_RATE_*`` constants.
             NOTE: currently only supported by FFmpegDecoder!
         audio_channels:
             A specific number of channels you wish the decoder to output,
             rather than relying on autimatic detection. For possible values see
-            the AUDIO_CHANNELS_* constants.
+            ``pyglet.media.AUDIO_CHANNELS_*`` constants.
             Note: currently only supported by FFmpegDecoder!
         audio_resample_hq:
             Whether to use high-quality resampling when resamplig is required
--- a/pyglet/media/__init__.py
+++ b/pyglet/media/__init__.py
@@ -72,24 +72,39 @@ AUDIO_SAMPLE_RATE_96000 = 96000
 AUDIO_SAMPLE_RATE_176400 = 176400
 AUDIO_SAMPLE_RATE_192000 = 192000
 
-_VALID_AUDIO_SAMPLE_FORMATS = (AUDIO_SAMPLE_FORMAT_U8,
-                               AUDIO_SAMPLE_FORMAT_S16,
-                               #AUDIO_SAMPLE_FORMAT_S24,  # Could be considered as well
-                               AUDIO_SAMPLE_FORMAT_S32,
-                               AUDIO_SAMPLE_FORMAT_F32)
+AUDIO_CHANNELS_MONO = 1
+AUDIO_CHANNELS_STEREO = 2
+AUDIO_CHANNELS_DUAL_MONO = -2
 
-_VALID_AUDIO_SAMPLE_RATES = (AUDIO_SAMPLE_RATE_22050,
-                             AUDIO_SAMPLE_RATE_44100,
-                             AUDIO_SAMPLE_RATE_48000,
-                             AUDIO_SAMPLE_RATE_88200,
-                             AUDIO_SAMPLE_RATE_96000,
-                             AUDIO_SAMPLE_RATE_176400,
-                             AUDIO_SAMPLE_RATE_192000)
+_VALID_AUDIO_SAMPLE_FORMATS = (
+    AUDIO_SAMPLE_FORMAT_U8,
+    AUDIO_SAMPLE_FORMAT_S16,
+    #AUDIO_SAMPLE_FORMAT_S24,  # Could be considered as well
+    AUDIO_SAMPLE_FORMAT_S32,
+    AUDIO_SAMPLE_FORMAT_F32,
+)
+
+_VALID_AUDIO_SAMPLE_RATES = (
+    AUDIO_SAMPLE_RATE_22050,
+    AUDIO_SAMPLE_RATE_44100,
+    AUDIO_SAMPLE_RATE_48000,
+    AUDIO_SAMPLE_RATE_88200,
+    AUDIO_SAMPLE_RATE_96000,
+    AUDIO_SAMPLE_RATE_176400,
+    AUDIO_SAMPLE_RATE_192000,
+)
+
+_VALID_AUDIO_CHANNELS = (
+    AUDIO_CHANNELS_MONO,
+    AUDIO_CHANNELS_STEREO,
+    AUDIO_CHANNELS_DUAL_MONO,
+)
 
 def load(filename: str, file: BinaryIO | None = None,
          streaming: bool = True, decoder: MediaDecoder | None = None,
          audio_sample_format: str | None = None,
-         audio_sample_rate: int | None = None) -> Source | StreamingSource:
+         audio_sample_rate: int | None = None,
+         audio_channels: int | None = None) -> Source | StreamingSource:
     """Load a Source from disk, or an opened file.
 
     All decoders that are registered for the filename extension are tried.
@@ -114,9 +129,15 @@ def load(filename: str, file: BinaryIO | None = None,
             the AUDIO_SAMPLE_FORMAT_* constants.
             NOTE: currently only supported by FFmpegDecoder!
         audio_sample_rate:
-            A specicif audio sample rate (in Hz) you wish the decoder to
-            output, rather than relying on automatic detection.
+            A specific audio sample rate (in Hz) you wish the decoder to
+            output, rather than relying on automatic detection. For possible
+            values see the AUDIO_SAMPLE_RATE_* constants.
             NOTE: currently only supported by FFmpegDecoder!
+        audio_channels:
+            A specific number of channels you wish the decoder to output,
+            rather than relying on autimatic detection. For possible values see
+            the AUDIO_CHANNELS_* constants.
+            Note: currently only supported by FFmpegDecoder!
     """
 
     audio_driver = get_audio_driver()
@@ -138,6 +159,12 @@ def load(filename: str, file: BinaryIO | None = None,
                 f"Expected one of: "
                 f"{', '.join([str(x) for x in _VALID_AUDIO_SAMPLE_RATES])}")
 
+    if audio_channels:
+        if audio_channels not in _VALID_AUDIO_CHANNELS:
+            raise ValueError(
+                f"Invalid audio_channels '{audio_channels}'. "
+                f"Expected one of: "
+                f"{', '.join([str(x) for x in _VALID_AUDIO_CHANNELS])}")
 
     if decoder:
         if type(decoder).__name__ == "FFmpegDecoder":
@@ -145,7 +172,8 @@ def load(filename: str, file: BinaryIO | None = None,
                 filename, file, streaming=streaming,
                 audio_sample_format=audio_sample_format,
                 audio_driver_sample_formats=audio_driver.sample_formats,
-                audio_sample_rate=audio_sample_rate)
+                audio_sample_rate=audio_sample_rate,
+                audio_channels=audio_channels)
         else:
             return decoder.decode(filename, file, streaming=streaming)
 
--- a/pyglet/media/__init__.py
+++ b/pyglet/media/__init__.py
@@ -104,7 +104,8 @@ def load(filename: str, file: BinaryIO | None = None,
          streaming: bool = True, decoder: MediaDecoder | None = None,
          audio_sample_format: str | None = None,
          audio_sample_rate: int | None = None,
-         audio_channels: int | None = None) -> Source | StreamingSource:
+         audio_channels: int | None = None,
+         audio_resample_hq: bool=False) -> Source | StreamingSource:
     """Load a Source from disk, or an opened file.
 
     All decoders that are registered for the filename extension are tried.
@@ -138,10 +139,16 @@ def load(filename: str, file: BinaryIO | None = None,
             rather than relying on autimatic detection. For possible values see
             the AUDIO_CHANNELS_* constants.
             Note: currently only supported by FFmpegDecoder!
+        audio_resample_hq:
+            Whether to use high-quality resampling when resamplig is required
+            (e.g. when a specific sample rate is requested), at the cost of
+            increased CPU usage.
+            Note: currently only supported by FFmpegDecoder!
+
     """
 
     audio_driver = get_audio_driver()
-    if audio_sample_format:
+    if audio_sample_format is not None:
         if audio_sample_format not in _VALID_AUDIO_SAMPLE_FORMATS:
             raise ValueError(
                 f"Invalid audio_sample_format '{audio_sample_format}'. "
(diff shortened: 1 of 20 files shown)


Square789:
This looks very promising. I'm seeing a few issues remaining:

Realistically, those files are very rare, but for 8-bit wav files (U8), the AudioFormat reports a sample type of S8, which doesn't exist as a playback option for any driver, so those now fail to be played back.
`sample_type` should default to `None` and in that case be set to `"uint"` for a `sample_size` of 8 and `"int"` otherwise.

`AudioFormat.__eq__` and `__repr__` should be updated to include the sample type. `__eq__` is especially important. If you queue an S32 and F32 wav file on the same player, the underlying playback stream/voice is not swapped out, and floats are interpreted as int on all drivers but OpenAL, leading to the same issue as in #1315.

The extension detection for OpenAL happens when no context exists, so it always fails; the supported formats should be queried and stored inside of `OpenALContext`.

I'm assuming the Low Shibata resampling method is some kind of gold standard? Not an audiophile by any means, so i can't make out a difference. However, it's indented with two levels of indentation.

The following are nitpicks/thoughts, nothing severe:

The coupling from FFmpegSource now importing the driver does not feel very good. It works, and similar hacks already exist i.e. in the XAudio2 driver where it imports its Audio Device Manager, so it's not the end of the world. Still, Ben suggested in the Discord to instead have `media.load` be responsible for importing the driver and passing in the supported formats to `.decode`.
To figure out which formats a driver supports in case no formats are given, the FFmpegSource could, instead of importing the active audio driver, fall back to the implicit standard and use S16.
But I'm not saying this must be done, since this method works in the end.

The `"F32", "U16"` etc. magic strings are standing out to me, but pyglet has similar ones in buffer object data format descriptions already. I guess the tradeoff of not importing a bunch of constants is alright.

fladd:
> Realistically, those files are very rare, but for 8-bit wav files (U8), the AudioFormat reports a sample type of S8, which doesn't exist as a playback option for any driver, so those now fail to be played back. `sample_type` should default to `None` and in that case be set to `"uint"` for a `sample_size` of 8 and `"int"` otherwise.

Ah, good point. This is a remainder from a previous implementation where I only used int and float (and then had the sample_format property append U to any 8-bit int automatically). I will fix this.

> `AudioFormat.__eq__` and `__repr__` should be updated to include the sample type. `__eq__` is especially important. If you queue an S32 and F32 wav file on the same player, the underlying playback stream/voice is not swapped out, and floats are interpreted as int on all drivers but OpenAL, leading to the same issue as in #1315.

I will have a look at it.

> The extension detection for OpenAL happens when no context exists, so it always fails; the supported formats should be queried and stored inside of `OpenALContext`.

That is strange, it correctly detects the float32 extension here on my Macbook just fine. Can you elaborate on this? Is this specific to Linux/Windows?

> I'm assuming the Low Shibata resampling method is some kind of gold standard? Not an audiophile by any means, so i can't make out a difference. However, it's indented with two levels of indentation.

Yes, low-shibata is a very common moderate noise shaping that helps with shaping the dithering noise away from the perceivable spectrum. I will fix the double indention.

> The following are nitpicks/thoughts, nothing severe:
> 
> The coupling from FFmpegSource now importing the driver does not feel very good. It works, and similar hacks already exist i.e. in the XAudio2 driver where it imports its Audio Device Manager, so it's not the end of the world. Still, Ben suggested in the Discord to instead have `media.load` be responsible for importing the driver and passing in the supported formats to `.decode`. To figure out which formats a driver supports in case no formats are given, the FFmpegSource could, instead of importing the active audio driver, fall back to the implicit standard and use S16. But I'm not saying this must be done, since this method works in the end.

Okay, I can move this to the `load` function.

> 
> The `"F32", "U16"` etc. magic strings are standing out to me, but pyglet has similar ones in buffer object data format descriptions already. I guess the tradeoff of not importing a bunch of constants is alright.

I tried to come up with some "meta" format descriptions that are free from implementation details like endianness and such. I also defined constants for those strings in pyglet.media though. I think that should be okay. Do you want me to (re)use some other format descriptions?

benmoran56:
@Square789 Good catch! 
> AudioFormat.__eq__ and __repr__ should be updated to include the sample type.

@fladd 
> (OpenAL) That is strange, it correctly detects the float32 extension here on my Macbook just fine. Can you elaborate on this? Is this specific to Linux/Windows?
Just a guess, but perhaps it's permanently resident on Mac, since it's baked into the OS :thinking: In any case, it does make logical sense to be attached to the context. 

Regarding keeping the decoders de-coupled, since I didn't elaborate earlier: 
It's ideal if the decoders can be imported and used without any driver at all. It makes them easier to test in isolation (to be fair we don't yet have these kinds of unit/integration tests :sweat:). It also would allow you to decode audio on a device without a sound card, etc.

fladd:
> Just a guess, but perhaps it's permanently resident on Mac, since it's baked into the OS 🤔 In any case, it does make logical sense to be attached to the context.

Okay, I moved this into the OpenALContext now. So basically, the global SAMPLE_FORMATS in interface will be empty now by default for OpenAL, and then is filled during driver creation after the context is set.

> Regarding keeping the decoders de-coupled, since I didn't elaborate earlier: It's ideal if the decoders can be imported and used without any driver at all. It makes them easier to test in isolation (to be fair we don't yet have these kinds of unit/integration tests 😓).

I decoupled it already and moved the format detection of the current driver into the `pyglet.media.load` method.

> It also would allow you to decode audio on a device without a sound card, etc.

Well, it would always work with the Silent driver (which supports all sample formats)...

Square789:
> That is strange, it correctly detects the float32 extension here on my Macbook just fine. Can you elaborate on this? Is this specific to Linux/Windows?

Yes, can confirm that the extension presence was reported as 0 when it was checked for at module initialization and 1 when the context was made current (using `examples.media.media_player` and some print statements).
With the current changes, the global constant is modified on context creation, which isn't too ideal. But OpenAL buffers are tied to a device, and if the formats were only available in a context, they'd need to be coupled to a context.
How about this:
Only set `SAMPLE_FORMATS` once as before; contains all formats, no matter whether supported or not:
```py
SAMPLE_FORMATS = {
	"U8":  (None, {1: al.AL_FORMAT_MONO8, 2: al.AL_FORMAT_STEREO8}),
	"S16": (None, {1: al.AL_FORMAT_MONO16, 2: al.AL_FORMAT_STEREO16}),
	"S32": (b"AL_EXT_32bit_formats",
		    {1: al.alGetEnumValue(b"AL_FORMAT_MONO_I32"),
             2: al.alGetEnumValue(b"AL_FORMAT_STEREO_I32")}),
	"F32": (b"AL_EXT_float32",
		    {1: al.alGetEnumValue(b"AL_FORMAT_MONO_FLOAT32"),
             2: al.alGetEnumValue(b"AL_FORMAT_STEREO_FLOAT32")}),
}
al.alGetError()  # Clear error from possibly unknown enum values
```
This should work, the OpenAL specification states that enum values are context-independent and thus i assume baked into the implementation. Unknown ones will resolve to 0 and should then never be accessed anyways.

Then, during context creation, the context can gather the ones it supports like
```py
self._supported_formats = []
for fmt, (required_extension, _) in SAMPLE_FORMATS.items():
    if required_extension is None or al.alIsExtensionPresent(required_extension):
        self._supported_formats.append(fmt)
```
`OpenALBuffer.data` can then just access `SAMPLE_FORMATS` and doesn't need to concern itself with the context the buffer will end up playing in.
The only disadvantage there is that, if for whatever reason, unsupported data were to be loaded into a buffer, it wouldn't fail there (as the buffer only is a descriptor/container) but a few microseconds later with a more cryptic OpenAL error when it'd be queued on an incompatible source. Ways around that would be to instead run the format compatibility check in `adaptation.py`, or modifying `OpenALBuffer.data` to take a context's supported formats.

> I tried to come up with some "meta" format descriptions that are free from implementation details like endianness and such. I also defined constants for those strings in pyglet.media though. I think that should be okay. Do you want me to (re)use some other format descriptions?

As i said, it's alright. I was just rambling there because no matter how much sense it might make, any magic string sets off an alarm bell for me. (And these format strings are totally fine; anyone working on audio backends is going to understand what they mean.)

One last nitpick i have is that `AudioFormat.sample_format` is a property. Since `AudioFormat` is more of a constant descriptor class/its attributes aren't supposed to be changed after instance creation, i think it would be fine to just compute that once during `__init__`. It would also be possible to go the other route and turn `bytes_per_frame/second` into properties, but i'd argue against that since formats are never modified after their creation, and it would come at a minor performance cost.

fladd:
> > That is strange, it correctly detects the float32 extension here on my Macbook just fine. Can you elaborate on this? Is this specific to Linux/Windows?
> 
> Yes, can confirm that the extension presence was reported as 0 when it was checked for at module initialization and 1 when the context was made current (using `examples.media.media_player` and some print statements). With the current changes, the global constant is modified on context creation, which isn't too ideal. But OpenAL buffers are tied to a device, and if the formats were only available in a context, they'd need to be coupled to a context. How about this: Only set `SAMPLE_FORMATS` once as before; contains all formats, no matter whether supported or not:
> 
> ```python
> SAMPLE_FORMATS = {
> 	"U8":  (None, {1: al.AL_FORMAT_MONO8, 2: al.AL_FORMAT_STEREO8}),
> 	"S16": (None, {1: al.AL_FORMAT_MONO16, 2: al.AL_FORMAT_STEREO16}),
> 	"S32": (b"AL_EXT_32bit_formats",
> 		    {1: al.alGetEnumValue(b"AL_FORMAT_MONO_I32"),
>              2: al.alGetEnumValue(b"AL_FORMAT_STEREO_I32")}),
> 	"F32": (b"AL_EXT_float32",
> 		    {1: al.alGetEnumValue(b"AL_FORMAT_MONO_FLOAT32"),
>              2: al.alGetEnumValue(b"AL_FORMAT_STEREO_FLOAT32")}),
> }
> al.alGetError()  # Clear error from possibly unknown enum values
> ```
> 
> This should work, the OpenAL specification states that enum values are context-independent and thus i assume baked into the implementation. Unknown ones will resolve to 0 and should then never be accessed anyways.
> 
> Then, during context creation, the context can gather the ones it supports like
> 
> ```python
> self._supported_formats = []
> for fmt, (required_extension, _) in SAMPLE_FORMATS.items():
>     if required_extension is None or al.alIsExtensionPresent(required_extension):
>         self._supported_formats.append(fmt)
> ```

Yeah I can implement that.

> One last nitpick i have is that `AudioFormat.sample_format` is a property. Since `AudioFormat` is more of a constant descriptor class/its attributes aren't supposed to be changed after instance creation, i think it would be fine to just compute that once during `__init__`. It would also be possible to go the other route and turn `bytes_per_frame/second` into properties, but i'd argue against that since formats are never modified after their creation, and it would come at a minor performance cost.

I can implement that, too, but note that currently there is only a property and no setter, so it already is read-only and cannot be changed after instance creation.

fladd:
@Square789 I implemented your suggested changes now. I do find it a bit awkward now with the actually supported formats being calculated in OpenALContext, but it works.

Hope this can be merged now.

Square789:
The supported formats are context-dependent, so i wouldn't know a better place to put that. I agree that the format compatibility check in `OpenALBuffer.data` is awkward, but if we do want to keep it, something like this is our only option.

But one last thing: The format code for signed int is now `I` instead of `S`. Up to you decide whether you want to reintroduce the mapping or rename `S32/16` to `I32/16`.

caffeinepills:
Regarding the naming of types, we have started to use the Python struct string values as data types, (such as shader module and buffers) it might be a good spot to also use it here. https://docs.python.org/3/library/struct.html#format-characters

Either way works, but just thought I would mention it here though.

fladd:
I reintroduced the correct mapping (i.e. "S16/32"). This makes most sense in an audio context.

fladd:
While I was at it, I also now added support for (optionally) setting `audio_sample_rate` to resample to a specific target sample rate (only FFmpeg for now).

Square789:
Looks good, I couldn't find any more issues testing this branch. Just that the most recent commit has a typo (specicif).

fladd:
And now I also implemented setting the audio channels as well, including a new option for dual-mono (since single channel can sometimes be on left channel only, depending on system configuration).

fladd:
Two more thoughts:

1) **Resampling algorithm**:
When setting `audio_sample_rate` to something other than `None` (i.e. automatic), then my current approach is to use the default `swr` resampler in FFmpeg. However, looking at https://src.infinitewave.ca, you can see that the `soxr` resampler, is much better. I would suggest to use `soxr` if available, and otherwise fall back to `swr`. Furthermore, when the target format larger than 16 bit, I would furthermore set the `precision` option for `soxr` to `28`. In either case, I would enable the `cheby` option, as this will improve "irrational" ratios, such as 44.1 to 48, which is probably the most use case.

   Implementing this also requires to move from using `swr_alloc_set_opts` and `swr_alloc_set_opts2` to setting all values of the context via `av_opt_set_int` calls in `get_formatted_swr_context`, because otherwise the resampler changes are not having any effect.

   I tried to implement this, but didn't manage to get the c_cyptes right for the calls. I will try again soon, but if anyone has any pointers here, that would be very helpful.

2) **Multi-channel audio:**
I think all audio drivers are nowadays perfectly capable of playing more than two channels of audio (with openal again needing an extension though). Pyglet currently artificially limits the number of channels to 2 though. I think this limitation can be lifted, and we could then allow more formats to go into `pyglet.media.load` (e.g. `AUDIO_CHANNELS_2_1 = 3`, `AUDIO_CHANNELS_QUAD = 4`, `AUDIO_CHANNELS_5_0 = 5`, `AUDIO_CHANNELS_5_1 = 6`, `AUDIO_CHANNELS_6.1 = 7`, `AUDIO_CHANNELS_7_1 = 8`).

fladd:
I tried adding soxr again, but couldn't make it work. So for now I implemented the option to use a more hiqh-quality resampling based on swr (with stricter filter settings). In the future, this could check if soxr is available, then use it, otherwise fall back to this high-quality swr resampling.

fladd:
Is there anything else I can do for this PR? Shall I give implementing multi-channel output a try? Or do you want this rather in another PR?

benmoran56:
@fladd I would say it would be better to target multi-channel audio for pyglet v3.0. 
First lets get this merged into master, then I will sync the changes into v3.0. From there we can expand on it if you want to take it forward. 

This looks good to merge for me, but I had one small suggestion. Instead of explicitly checking for the decoder type when loading: 
```py
        if type(decoder).__name__ == "FFmpegDecoder":
            return decoder.decode(...)
        else:
            return decoder.decode(filename, file, streaming=streaming)
```
It would be better to extend the base `MediaDecoder` class to add the extra arguments. 
It's OK that other decoders do not support them, if they are just treated as "hints". It does leave it open for us to extend some of the platform decoders if/where possible (for example, the GStreamerDecoder can likely add this support). For others where it doesn't make sense, the decoder can just ignore the hints. 

What do you think?

fladd:
@benmoran56 Done. Please check if this is what you had in mind.

On another note: While testing my multi-channel downmixer to dual-mono,  I found a bug where opening audio files with multiple channels (i.e. source channel count > 2), leads to `ffmpeg_read` getting stuck after a few audio packages (in my test around 70)  in the call to `avformat.av_read_frame(file.context, packet)`. And the Python process freezes. Some further testing revealed that this is related to the new way pyglet reads FFmpeg streams from memory. Reverting to the tempfile hack I use solves the bug and multi-channel audio plays fine. Very strange...

benmoran56 (approved):

PR #1429, opened 2026-04-05
Fix type hints and types to get unit tests to pass successfully on Python 3.8.20 on MacOS.

joaufi:
### Description

I fixed the type hints and types of various objects to get the command `pytest tests/unit` to run successfully with Python 3.8.20 on a MacOS.

Changes:
- I added `from __future__ import annotations` where missing.
- I fixed the type hints and type of [pyglet.libs.darwin.coreaudio.err_str_db](https://github.com/pyglet/pyglet/blob/development/pyglet/libs/darwin/coreaudio.py#L199) so that `err_str_db` is now actually immutable, as I imagine was intended.
- I added a unit test to cover the immutability of `err_str_db`. 
- I fixed a link in a comment also in `coreaudio.py`.
- I updated gitignore to ignore any dir starting with `.venv` to support virtualenv names like `.venv-3.8.20`.
- I added Python 3.8 and 3.9 to the `python-version` matrix of the unit tests GitHub Action runner for `macos-latest`. 
  - Note: I have not fixed the **integration** tests for MacOS Python 3.8, but since the `unittests.yaml` has it ignored I felt it safe to add back in those Python versions. I will work on fixing the integration tests next.

### Running

*Ensure you are running with Python 3.8.20 on MacOS.*

Run:

```bash
pytest tests/unit
```

And all tests should run successfully.

### Discussion

Although the changes I made got the unit tests to run for Python 3.8.20, there are a lot of areas in the code that may not be covered by unit tests that use language constructs not available to Python 3.8 (such as using the `|` operator as a replacement for `Union` in type hints, which was introduced in Python 3.10). 

Both [Python 3.8 and Python 3.9 are end-of-life and no longer supported](https://devguide.python.org/versions/). Current [pyglet development documentation references using Python3.8+](https://pyglet.readthedocs.io/en/latest/internal/virtualenv.html#development-environment) for development. It may be worthwhile to develop a roadmap for no longer supporting Python 3.8 / 3.9 and update all development and user documentation to reflect that coming change.

There is [a discussion thread for this PR in the Discord at this link](https://discord.com/channels/438622377094414346/1490438039909040248) as well.

### DIFF
--- a/pyglet/graphics/buffer.py
+++ b/pyglet/graphics/buffer.py
@@ -23,7 +23,7 @@ if TYPE_CHECKING:
 
 
 ByteSource = Union[bytes, bytearray, memoryview]
-DataSource = Union[Sequence[Union[float, int]], Array[Any]]
+DataSource = Union[Sequence[Union[float, int]], Array]
 
 _DATA_TYPE_TO_CTYPE: dict[str, type[Any]] = {
     "?": ctypes.c_bool,
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -38,7 +38,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        python-version: [ '3.10', '3.11', '3.12', '3.13', '3.14']
+        python-version: [ '3.8', '3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
     name: Python ${{ matrix.python-version }} macOS
     steps:
       - uses: actions/checkout@v4
--- a/.gitignore
+++ b/.gitignore
@@ -6,9 +6,7 @@ __pycache__
 # Virtualenvs (from project root)
 /env
 /.env
-/.venv
-/.venv2
-/.venv3
+/.venv*
 
 # Pyenv
 .python-version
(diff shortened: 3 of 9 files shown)


pyglet/graphics/buffer.py
```diff
@@ -23,7 +23,7 @@
 
 
 ByteSource = Union[bytes, bytearray, memoryview]
-DataSource = Union[Sequence[Union[float, int]], Array[Any]]
+DataSource = Union[Sequence[Union[float, int]], Array]
```
joaufi, line 26: Initial bug here is that `ctypes.Array` is not sub-scriptable in Python 3.8.

After reading [the documentation for ctypes.Array](https://docs.python.org/3.8/library/ctypes.html#ctypes.Array) I am positive that my solution is not the correct one. 

However, reading through [the ctypes doc for creating Array types](https://docs.python.org/3.8/library/ctypes.html#arrays), I am also unsure of how to best proceed, as I do not know what the `DataSource` type is supposed to *be*. 

Any suggestions here would be welcome. Thanks!

joaufi:
Closing this pull request due to Python 3.8 not really being of importance to support, from a user or development perspective. I will open a PR that reflects a pyglet support stance somewhere in the documentation to help define development contributing paths and also user-support paths in the future.

For those interested in the discussion behind this PR closure, [please see this Discord thread](https://discord.com/channels/438622377094414346/1490438039909040248).

Issue #1430, opened 2026-04-06
Update development and user documentation to include Python version support information.


joaufi:
### Description

There isn't a place in the documentation- either in the user-facing documentation or the development-contributing documentation- that defines what pyglet's Python version support stance is. Update the documentation so that pyglet's Python version support stance is documented.

Some useful discussion and places where the documentation could be updated can be found in [this Discord thread](https://discord.com/channels/438622377094414346/1490438039909040248).

The Python version support information might fit somewhere around [the pyglet Installation docs](https://pyglet.readthedocs.io/en/latest/programming_guide/installation.html), so that users who are starting out would be pushed to use a more recent version of Python when beginning. A note could also be added to [the pyglet Github Requirements docs](https://github.com/pyglet/pyglet?tab=readme-ov-file#requirements).

Additionally, it may be worthwhile to update [the Development Environment pyglet docs](https://pyglet.readthedocs.io/en/latest/internal/virtualenv.html#development-environment) to say `- The most recent non-End-Of-Life version of Python` instead of `- Python 3.8+`. That way anyone developing and contributing will start on a version of Python that is most likely to be supported by both pyglet and the Python Software Foundation.

joaufi:
This was mostly addressed with the linked merged PR. Also, with the release of pyglet 3.0, the developing documentation should automatically updated as it references [the value of pyglet.MIN_PYTHON_VERSION](https://github.com/pyglet/pyglet/pull/1434). Seems safe to close this issue.

PR #1433, opened 2026-04-06
Add Python Version Support section to Installation docs.

joaufi:
### Description

Added a Python Version Support section to the Installation user docs.

Partially addresses issue #1430. 

### Running

Pull and checkout this branch and run the following:

```bash
./make.py docs --open
```

Then navigate to the Installation portion of the Python docs in the static docs web page that opens in your brower. You should see a new heading underneath Installation titled "Python Version Support" with content that looks like the following:

image


joaufi:
Rebasing and targeting master.

Issue #1268, opened 2025-02-01
Window on_draw events skipped


cdeil:
**Describe the bug**

I am getting started with Arcade and Pyglet. Both describe that by default window on_draw should run at 60 fps. So I was surprised that I consistently run at lower fps even when doing almost nothing. It turns out the issue is that the `fps = 1 / dt` between `on_draw` callbacks sometimes is at 60 fps, but sometimes calls are skipped and it drops to 30 fps. When averaging multiple calls this results in some fps between 30 and 60, but not at the desired 60.

See script and plot at the end of the description.

x-ref #621 but I didn't test on other platforms yet.

**System Information:**

```
% python -m pyglet.info
Platform
------------------------------------------------------------------------------
platform:  macOS-15.3-arm64-arm-64bit-Mach-O
release:   24.3.0
machine:   arm64

Python
------------------------------------------------------------------------------
implementation: CPython
sys.version: 3.13.1 (main, Dec  3 2024, 17:59:52) [Clang 16.0.0 (clang-1600.0.26.4)]
sys.maxint: 9223372036854775807
PyObjC not available
os.getcwd(): /Users/cdeil/code/projects/cgames/checks

pyglet
------------------------------------------------------------------------------
pyglet.version: 2.1.2
pyglet.compat_platform: darwin
pyglet.__file__: /Users/cdeil/code/projects/cgames/.venv/lib/python3.13/site-packages/pyglet/__init__.py
pyglet.options['audio'] = ('xaudio2', 'directsound', 'openal', 'pulse', 'silent')
pyglet.options['debug_font'] = False
pyglet.options['debug_gl'] = True
pyglet.options['debug_gl_trace'] = False
pyglet.options['debug_gl_trace_args'] = False
pyglet.options['debug_gl_shaders'] = False
pyglet.options['debug_graphics_batch'] = False
pyglet.options['debug_lib'] = False
pyglet.options['debug_media'] = False
pyglet.options['debug_texture'] = False
pyglet.options['debug_trace'] = False
pyglet.options['debug_trace_args'] = False
pyglet.options['debug_trace_depth'] = 1
pyglet.options['debug_trace_flush'] = True
pyglet.options['debug_win32'] = False
pyglet.options['debug_input'] = False
pyglet.options['debug_x11'] = False
pyglet.options['shadow_window'] = True
pyglet.options['vsync'] = None
pyglet.options['xsync'] = True
pyglet.options['xlib_fullscreen_override_redirect'] = False
[...]

PR #1435, opened 2026-04-08
Set realtime thread priority in CocoaPlatformEventLoop.

joaufi:
### Description

_(Note: This PR description has had a few items update from when it was in WIP- please take the time to re-read it.)_

Sets the current process thread's policy/priority to realtime behavior.

All the functionality works as expected, but doesn't seem address the behavior seen in #1268 (running on my Macbook Air M4) . The changes are still worth reviewing and merging in as it seems standard to set the thread priority to realtime behavior (e.g. GameMaker does this as well) and pyglet applications will see a performance improvement from this change being the default.

A few things:
- ~I have not added a test explicitly testing the changes to `nsapp_step` and would welcome any suggestions on if a new test is needed and, if so, the best way to approach writing it.~ (Removed the `nsapp_step` changes as per review comments.)
- I tried to adhere to the naming conventions that I saw elsewhere in `pyglet/libs/darwin/cocoapy/cocoalibs.py` where C-stuff follows a `camelCase` naming pattern but Python-stuff follows a `snake_case` naming pattern.
- I tested the realtime thread policy with both the normal  `CocoaPlatformEventLoop` and the `CocoaAlternateEventLoop`. ~with the new changes to `nsapp_step`. I vaguely understand the `nsapp_step` changes mentioned in #1268 and it _does_ make a marked improvement in FPS timings (consistency of speed, max speed, avg speed, and mean speed). Someone who understands the particulars of this better than me may want to verify that there aren't any knock-on effects from these changes.~ There is less of an improvement to FPS when running with a non-alternate `CocoaPlatformEventLoop` but the timings are grouped more tightly and are more consistent.

### Running

All of these commands were run on a MacOS with Python 3.11.15.

Install Polars and Matplotlib into your virtual env by running:

```bash
pip install polars matplotlib
```

Copy this script into a `fps_test.py` in the root dir of the pyglet repo:

```python
import time
import pyglet

times = []

pyglet.options.osx_alt_loop = True

fps_display = pyglet.window.FPSDisplay(window=window)

@window.event
def on_draw():
    times.append(time.time())
    window.clear()
    fps_display.draw()

pyglet.app.run()

import polars as pl
import matplotlib.pyplot as plt

pl.DataFrame({"times": times}).write_csv("pyglet_times.csv")

times = pl.read_csv("pyglet_times.csv")["times"]
dt = times.diff()
fps = 1 / dt
plt.plot(fps, "o")
plt.axhline(y=fps.mean(), color='r', linestyle='--')
plt.text(0, fps.mean(), f'Mean FPS: {fps.mean():.2f}', color='r', va='bottom')
plt.xlabel("Sample number")
plt.ylabel("FPS rate (Hz)")
plt.savefig("fps_timings.png")
```

Run the above script by running:

```bash
python fps_test.py
```

And after letting the black window with the FPS counter run for a little while hit ESC to close the window.

The FPS timings will be output to a `pyglet_times.csv` once the window closes. Then, the rate is plotted into a data visual into an output PNG titled `fps_timings.png`. You can view either the the FPS timings or the PNG timing plot by running:

```python
open pyglet_times.csv
open fps_timings.png
```

### DIFF
--- a/pyglet/app/cocoa.py
+++ b/pyglet/app/cocoa.py
@@ -6,7 +6,7 @@ import time
 from pyglet import app
 from pyglet.app.base import PlatformEventLoop, EventLoop
 from pyglet.libs.darwin import cocoapy, AutoReleasePool, ObjCSubclass, PyObjectEncoding, ObjCInstance, send_super, \
-    ObjCClass, get_selector
+    ObjCClass, get_selector, set_realtime_thread_policy
 
 NSApplication = cocoapy.ObjCClass('NSApplication')
 NSMenu = cocoapy.ObjCClass('NSMenu')
@@ -161,6 +161,7 @@ class CocoaPlatformEventLoop(PlatformEventLoop):
 
     def __init__(self):
         super().__init__()
+        set_realtime_thread_policy()
         self._timer = None
 
         with AutoReleasePool():
--- a/pyglet/libs/darwin/cocoapy/cocoalibs.py
+++ b/pyglet/libs/darwin/cocoapy/cocoalibs.py
@@ -1,8 +1,13 @@
+from __future__ import annotations
+
+import logging
 from ctypes import *
 from ctypes import util
 
 from .runtime import send_message, ObjCInstance
 from .cocoatypes import *
+
+logger = logging.getLogger(__name__)
 
 ######################################################################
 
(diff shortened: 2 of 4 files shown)


pyglet/libs/darwin/cocoapy/cocoalibs.py
```diff
@@ -713,3 +716,104 @@ def cfarray_to_list(cfarray):
...
+    err = cocoa.thread_policy_get(
+        cocoa.mach_thread_self(), 
+        THREAD_TIME_CONSTRAINT_POLICY,
+        byref(extendedPolicy), # Send the address of the policy struct
+        byref(THREAD_TIME_CONSTRAINT_POLICY_COUNT),
+        byref(getDefault)
+    )
+    if err != KERN_SUCCESS:
+        raise RuntimeError('Failed to get darwin thread policy with thread_policy_get.')
+
```
joaufi: This error (and the other `if err != KERN_SUCCESS` errors below) concern me because it will cause the whole application to fail if the thread policy setting fails.  I don't know how likely that is to happen when interacting with Cocoa. It may be better to fail gracefully by not raising an error and simply continue on, but log the failure under a debug flag.

Any thoughts on these errors?

joaufi: Changed this to `logger.error` instead of raising an error. Failing the entire application for failing to get or set thread policy seems like the wrong move, as if it fails the app should still run on default thread priority.

pyglet/app/cocoa.py
```diff
@@ -229,7 +230,10 @@ def nsapp_start(self, interval):
 
     def nsapp_step(self):
         """Used only for CocoaAlternateEventLoop"""
-        self._event_loop.idle()
+        sleep_time = self._event_loop.idle()
+        while sleep_time:
+            time.sleep(sleep_time)
+            sleep_time = self._event_loop.idle()
```
joaufi: I haven't researched how to write a good unit/integration test for this changed code block. Any suggestions are welcome! 😄

caffeinepills: This does not look proper upon review. This is either calling idle multiple times per tick or worse blocking the tick longer than it should. What's the reason for this?

joaufi: This change is from https://github.com/pyglet/pyglet/issues/1268#issuecomment-2644725645 ; In the PR description body I mentioned that I did not understand the particulars of this change and only that it appeared to make the FPS times better and more consistent.

I'm happy to remove it, or attempt to improve upon it such that the spirit of the changes are kept and are also technically correct within the context of the app.

joaufi: Removed the changes and reverted back to the previous method. Please reopen this comment if you'd like to discuss more. Thanks!

tests/integration/libs/darwin/cocoapy/test_cocoalibs.py
```diff
@@ -0,0 +1,87 @@
...
+    assert not any(
+        record.levelname == "ERROR" for record in caplog.records
+    ), "Error was logged, call failed"
+
+    # A previous iteration of this test retrieved the thread priority
+    # before calling `set_realtime_thread_policy` to compare that value
+    # with the value after the call. Unfortunately, other tests run
+    # code paths that call `set_realtime_thread_policy` before this
+    # test is run, so we're only checking that the thread priority
+    # is a number indicating realtime behavior.
```
joaufi, line 86: If I added `pytest-forked` (https://pypi.org/project/pytest-forked/) or `pytest-isolate` (https://pypi.org/project/pytest-isolate/) to the `tests/requirements.txt` I could run this test in a separate process and write the test as described in this comment (not sure if it is worth adding the dependency for this one test). 

Any suggestions here (either leaving as-is or adding the dependency) are welcome. 😄

cbrnr:
Quick question, is this the new default or do we have to enable this behavior?

benmoran56:
This is the new default, and can be found in the upcoming v3.0 release (currently v3.0.dev5).

PR #1444, opened 2026-04-29
Add Frame class to gui.rst

abhro:


caffeinepills:
Good catch. Approved.

PR #1451, opened 2026-05-13
Update 3.0 info output.

caffeinepills:
Update the info output for pyglet 3.0.

- Includes information like the window pixel scale ratio and framebuffer size now.
- Adds an extension count, which can be useful to see how many are exposed. (Low number may be old hardware or bad driver)
- Only output first 12 extensions, add optional -extensions system arg if someone wants to output all extensions for advanced debugging.
- Add common graphics info output like max texture size.
- Only show wintab output on Windows OS's.
- Don't output GLX info for non-linux. Add WGL output for Windows.
- Add media codec lists that are found.
- Group some of the audio stuff under media instead of separately.

Also randomly: Moved wincodec's pure library file to libs folder.


benmoran56 (approved):
Looks good to me!

PR #1463, opened 2026-07-24
Fix Mat4.scale() to compose with self instead of only scaling the diagonal

### LINKED ISSUE #1462: Mat4.scale() only scales the diagonal, gives wrong result when self isn't diagonal
`Mat4.scale` only multiplies the three diagonal entries of the underlying tuple:

```python
def scale(self, vector: Vec3) -> Mat4:
    """Get a scale Matrix on x, y, or z axis."""
    temp = list(self)
    temp[0] *= vector[0]
    temp[5] *= vector[1]
    temp[10] *= vector[2]
    return Mat4(*temp)
```

That's only correct when `self` is diagonal (identity, or already a pure scale). For any matrix with off-diagonal entries in its first three columns (e.g. after a rotation has been composed in), it silently drops most of the scale.

Repro, using pyglet's own doc convention that matrices multiply on the left of column vectors (so `.scale(v)` should be equivalent to `self @ ScaleMatrix(v)`, exactly like `Mat4.translate` already does with `self @ Mat4(1,0,0,0, 0,1,0,0, 0,0,1,0, *vector, 1)`):

```python
from pyglet.math import Mat4, Vec3

M = Mat4().rotate(0.5, Vec3(0, 0, 1))
v = Vec3(2, 3, 4)

S = Mat4(v.x, 0, 0, 0,  0, v.y, 0, 0,  0, 0, v.z, 0,  0, 0, 0, 1)
expected = M @ S
got = M.scale(v)

print(expected)
print(got)
```

Output:

```
Mat4(1.7552, 0.9589, 0.0, 0.0
    -1.4383, 2.6327, 0.0, 0.0
    0.0, 0.0, 4.0, 0.0
    0.0, 0.0, 0.0, 1.0)
Mat4(1.7552, 0.4794, 0.0, 0.0
    -0.4794, 2.6327, 0.0, 0.0
    0.0, 0.0, 4.0, 0.0
    0.0, 0.0, 0.0, 1.0)
```

Row/column 0 and 1 (the ones actually touched by the rotation) diverge because only `self[0][0]` and `self[1][1]` got scaled, not the whole columns.

For an identity/diagonal `self` the two happen to agree, which is presumably why this hasn't been noticed - same root cause class as #1460 for `Mat3.scale`/`Mat3.translate` (missing test coverage for these methods let a convention mismatch slip through).

`Mat4.rotate` and `Mat4.translate` are correctly implemented as `self @ <transform matrix>`; `scale` should follow the same pattern.

Gooh456:
`Mat4.scale()` just multiplies `temp[0]`, `temp[5]`, `temp[10]` (the three diagonal entries) and returns that. That's only the same as `self @ ScaleMatrix(v)` when `self` is already diagonal. As soon as `self` has anything off-diagonal in its first three columns, like after a `rotate()`, the scale lands on the wrong elements and most of it just gets dropped.

`Mat4.translate()` already does this the right way with `self @ Mat4(1,0,0,0, 0,1,0,0, 0,0,1,0, *vector, 1)`. Made `scale()` match that pattern instead of poking at individual indices.

Same root cause as #1460 basically: no test existed for `Mat4.scale()` on a non-identity matrix, so the identity-only case (where the two approaches happen to agree) never caught it.

Added `test_mat4_scale`, same idea as the `test_mat3_scale`/`test_mat3_translate` tests added in #1460: build a rotated matrix, scale it both ways, compare. Full `tests/unit/math/` suite passes locally (116 passed).

Fixes #1462

### DIFF
--- a/pyglet/math.py
+++ b/pyglet/math.py
@@ -1256,11 +1256,7 @@ class Mat4(_typing.NamedTuple):
 
     def scale(self, vector: Vec3) -> Mat4:
         """Get a scale Matrix on x, y, or z axis."""
-        temp = list(self)
-        temp[0] *= vector[0]
-        temp[5] *= vector[1]
-        temp[10] *= vector[2]
-        return Mat4(*temp)
+        return self @ Mat4(vector[0], 0, 0, 0, 0, vector[1], 0, 0, 0, 0, vector[2], 0, 0, 0, 0, 1)
 
     def translate(self, vector: Vec3) -> Mat4:
         """Get a translation Matrix along x, y, and z axis."""
--- a/tests/unit/math/test_mat.py
+++ b/tests/unit/math/test_mat.py
@@ -147,6 +147,19 @@ def test_mat3_scale():
     assert result == Vec3(2, 3, 1)
 
 
+def test_mat4_scale():
+    # scale() must be equivalent to post-multiplying by a scale matrix, not just
+    # multiplying the diagonal of `self`. The two only agree when `self` is
+    # already diagonal, so use a rotated matrix to catch the difference.
+    matrix = Mat4().rotate(0.5, Vec3(0, 0, 1))
+    vector = Vec3(2, 3, 4)
+
+    scale_matrix = Mat4(vector.x, 0, 0, 0, 0, vector.y, 0, 0, 0, 0, vector.z, 0, 0, 0, 0, 1)
+    expected = matrix @ scale_matrix
+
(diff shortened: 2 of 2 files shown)


benmoran56 (approved):

benmoran56:
Thank you for the pull request. The Mat4.from_scale method already exists as a way to get a simple scaled identity matrix, so it makes sense for this to mimic translate as this is operating on an existing Matrix.

PR #1464, opened 2026-07-24
Expanding clock behavior with event chaining

caffeinepills:
This PR is to expand the clock module a bit to add generator-based event chaining to pyglet.clock, similar to async implementations that other event loop systems provide.

It's intended to make scheduling multi-step flows easier to manage, as well as, allow cancel, pause, and resume. Also to help remove a lot of boilerplate and frustration with trying to manage multiple `schedule_once` calls to chain together things.

Something like this:
```py
def part2(dt):
    print("part2")

def part1(dt):
    print("part1")
    pyglet.clock.schedule_once(part2, 0.5)

print("starting")
pyglet.clock.schedule_once(part1, 1.0)
```

Would become:
```py
@pyglet.clock.chain
def part_chain():
    print("starting")
    yield 1.0
    print("part 1")
    yield 0.5
    print("part 2")

chain = part_chain()
chain.start()
```

As you can see it can easily be created with a decorator and is much easier to understand the flow. You can even yield from another chain creating little pieces that can be re-used.

You can also create a chain without the decorator: `chain = pyglet.clock.Chain(part_chain())`

Cancel it with `chain.cancel()` or pause and resume via `chain.pause()`/`chain.resume()`.

It has lots of different helper functions for looping chains, parallel runs, race, timeouts, and integrating with other yielding return functions. 

There are also ChainGroups which help control life cycle. For example, you might have a character that has multiple chains being run on it. Maybe size changes, color changes, waits for actions being done, but then suddenly that character dies, you would want an easy way to stop those. That is where the group comes in. Furthermore you can create an inheritance of ChainGroups as well. For example, maybe you have a scene chain, and a ui chain. This way you can stop all of your scene chains during transitions or other features while keeping your UI separate.

Chains can also be used for separate clock instances as well. For example you can use the default clock as your game clock, but maybe you have a separate clock called UI clock. That way you can directly target chain control by different pieces of your application.

This could also be used for various animation features in the future.

At this point it might seem just easier to implement something like asyncio and then we could have features like that built in, but it would require a larger rewrite and would still be separate from the clock. 

Read the docs for more examples and explanations.

Looking for any input for improvements, or oversights, and just discussion on it.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.9', '3.10', '3.11', '3.12', '3.13', '3.14', 'pypy-3.10' ]
+        python-version: [ '3.10', '3.11', '3.12', '3.13', '3.14', 'pypy-3.10' ]
 
     name: Python ${{ matrix.python-version }} Ubuntu
     steps:
@@ -78,7 +78,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: [ '3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
+        python-version: [ '3.10', '3.11', '3.12', '3.13', '3.14']
 
     name: Python ${{ matrix.python-version }} Windows
     steps:
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -56,6 +56,7 @@ If you are migrating from an older version of pyglet, please read through
    programming_guide/rendering
    programming_guide/camera
    programming_guide/events
+   programming_guide/event_chains
    programming_guide/gui
    programming_guide/time
    programming_guide/context
--- a/doc/programming_guide/event_chains.rst
+++ b/doc/programming_guide/event_chains.rst
@@ -476,3 +476,81 @@ running::
         def destroy(self):
             self.chains.clear()
 
+
+Tweening values
+---------------
+
+Tweens update a value gradually over time. They are useful for visual changes
+such as fading, movement, scaling, or rotation, as well as audio levels,
+gameplay parameters, and any other state that changes smoothly. Yielding a
+tween keeps the update logic separate from the sequence that should continue
+after it finishes.
+
+:py:func:`~pyglet.clock.tween` drives an update callable with normalized
+progress over a duration. The callable receives one ``float`` argument. It
+can be a function, a lambda, a bound method, or any callable object. The tween
+returns a chain that can be yielded like any other child chain::
+
+    class FadeOut:
+        def __init__(self, target):
+            self.target = target
+            self.start_opacity = target.opacity
+
+        def update(self, progress):
+            self.target.opacity = self.start_opacity * (1.0 - progress)
+
+    @pyglet.clock.chain
+    def remove_sprite(sprite):
+        yield pyglet.clock.tween(0.5, FadeOut(sprite).update)
+        sprite.delete()
+
+For a one-off update, define a local function instead::
+
+    start_opacity = sprite.opacity
+
+    def update(progress):
+        sprite.opacity = start_opacity * (1.0 - progress)
+
+    yield pyglet.clock.tween(0.5, update)
+
+For a compact one-off update, a lambda works too. Capture the starting value
+before creating the lambda so that each update is based on the same value::
+
+    start_opacity = sprite.opacity
+    yield pyglet.clock.tween(
+        0.5,
+        lambda progress: setattr(sprite, 'opacity', start_opacity * (1.0 - progress)),
+    )
+
+With the default linear easing, the update callable receives ``0.0`` when the
+tween starts and ``1.0`` when it finishes. A zero-duration tween applies only
+the eased final value. Tweens are chains themselves, so they can be started
+directly or composed with :py:func:`~pyglet.clock.parallel` and
+:py:func:`~pyglet.clock.race`. They pause, resume, and stop like any other
+chain::
+
+    yield pyglet.clock.parallel(
+        pyglet.clock.tween(0.5, FadeOut(sprite).update),
+        pyglet.clock.tween(0.5, update_position),
+    )
+
+Pass an easing function to transform the linear progress. Easing functions
+receive a value from ``0.0`` through ``1.0``. Their result is not clamped,
+which permits curves that intentionally overshoot::
+
+    yield pyglet.clock.tween(
+        0.5,
+        FadeOut(sprite).update,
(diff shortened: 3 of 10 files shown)


pvcraven:
I think the idea is a good one. And it is pretty clean for the user.

Here's claude's feedback:

# Review: PR #1464 — "Expanding clock behavior with event chaining"

## Overview

Large feature addition (+3100/-28, 8 files) to `pyglet.clock` that adds generator-based
coroutine chaining (`Chain`, `@chain`, `ChainGroup`), composition helpers (`parallel`,
`race`, `repeat`/`repeat_until`/`repeat_duration`, `timeout`), a callback-adapter
(`yielding_callback`/`wait_for_event`), and a tweening system (`tween` + easing functions).
It ships with a 556-line doc page, ~340 new/updated unit tests, and two unrelated
drive-by changes: dropping Python 3.9 from CI and a micro-optimization to
`Clock._get_soft_next_ts`.

## Critical issue: breaks Python 3.9 (and the CI change hides it)

`pyglet/clock.py` adds:

```python
from typing import Any, Callable, Protocol, TypeAlias, TypeVar, cast
```

`typing.TypeAlias` was only added in **Python 3.10** (PEP 613). Since this is a plain
`import` statement (not an annotation), `from __future__ import annotations` does
**not** defer it — this line executes eagerly at import time. On Python 3.9 this will
raise `ImportError` as soon as `pyglet.clock` is imported.

`setup.py` still lists `'Programming Language :: Python :: 3.8'` and `'3.9'` as
supported classifiers, yet `.github/workflows/unittests.yml` in this same PR quietly
removes `'3.9'` from both the Ubuntu and Windows test matrices. That strongly suggests
the CI failure this caused was "fixed" by dropping the version from CI rather than
fixing the incompatibility (e.g., `try: from typing import TypeAlias except ImportError:
from typing_extensions import TypeAlias`, or simply dropping the `TypeAlias`
annotations since they're not required for the code to function). As it stands, this
PR silently breaks a still-advertised supported Python version. This should either be
fixed at the source or be an explicit, separate decision (bump `python_requires`/
classifiers, document the drop) rather than a silent side effect of the feature PR.

## Correctness bug: `repeat_duration(factory, 0)` doesn't match its docstring

```python
def repeat_duration(factory, duration, *, clock=None):
    """... If duration is 0, it repeats until stopped."""
    ...
    def _repeat_duration():
        winner_index, result = yield race(
            chain(_repeat_forever, clock=clock)(),
            timeout(duration, clock=clock),
        )
        return result if winner_index == 0 else None
```

There's no special-casing for `duration == 0`. `timeout(0)` still yields a delay of
`0`, which schedules via `Clock.schedule()` and fires on the very next tick — it does
**not** mean "never." So `repeat_duration(factory, 0.0)` will race the infinite
repeater against a timeout that wins almost immediately, stopping the "forever" loop
after (at most) one tick — the opposite of the documented behavior. There's also no
test covering `duration=0`, so this regression is untested. Either fix the doc or
special-case `duration <= 0` to bypass the race entirely and return the plain infinite
chain.

## Design inconsistency: stop propagation differs between plain `yield child` and `parallel`/`race`

- `Chain._wait_for_child`: if a chain waited on via `result = yield some_chain()` is
  stopped externally, the **parent silently stops too** (`on_stop` fires, no
  exception).
- `_ParallelOperation`/`_RaceOperation`: if a child chain stops, it's converted into a
  `RuntimeError` **failure** on the parent (`on_error` fires) unless
  `continue_on_stop=True`.

Composing the exact same child chain through these two mechanisms produces different
termination semantics (stop vs. error) for what's conceptually the same event. Worth a
deliberate decision and a doc note, since users will likely be surprised switching
between `yield child` and `yield parallel(child)`.

## Minor issue: `ChainGroup.start()` can raise on an already-finished chain

`ChainGroup.add()` guards against already-`done` chains (`if chain.done: return
chain`), but `ChainGroup.start()` calls `self.add(...)` then unconditionally calls
`chain.start()` regardless of that guard — which raises `RuntimeError("A completed or
stopped chain cannot be restarted")` per `Chain.start()`. The `done` check in `add()`
gives a false sense of defensiveness that `start()` doesn't actually provide.

## Scope creep

Three changes are unrelated to "event chaining" and bundled into this PR, making it
harder to review/revert independently:

- Dropping Python 3.9 from CI (see above — and it's masking a real bug).
- `pyglet/event.py`: lint-only fixes (`noqa: T201`, import reshuffle, docstring typo
  fix).
- `Clock._get_soft_next_ts`: replacing an in-place heap sort + linear scan with a
  `sorted()` snapshot + `bisect_left`. This looks correct (verified: it preserves the
  original "smallest `next_ts >= ts-e`, then check `<= ts+e`" semantics) and is a nice
  perf win, but it's a separate concern from chaining and would be easier to
  review/land on its own.

## Code quality — positives

- The `Chain` state machine (`_wait_token` staleness guard, `_pending_advance` while
  paused, unschedule/reschedule bookkeeping for pause/resume) is carefully written; I
  traced through the pause/resume/stop interactions for children, callback operations,
  and tweens and didn't find re-entrancy bugs beyond the two issues above.
- `_TweenManager` consolidating all tweens on a clock into one scheduled callback
  (rather than one per tween) is a sensible perf choice, and is actually tested
  (`test_clock_uses_one_scheduled_callback_for_multiple_tweens`).
- Test coverage is thorough for the core chain/parallel/race/tween mechanics,
  including pause/resume-during-wait and cancellation edge cases.
- Documentation is unusually complete for a PR still marked as being in discussion.

## Test coverage gaps

- No test for `repeat_duration` with `duration=0` (see bug above).
- No test exercising the stop-vs-error asymmetry noted above (plain `yield child` stop
  vs. `parallel`/`race` child stop).
- No test importing `pyglet.clock` under Python 3.9 semantics (can't easily be tested
  in-repo, but the CI matrix change is exactly what's hiding this).

## Security

No injection/untrusted-input concerns — this is a pure in-process scheduling API.
Exception handling is broad (`except Exception: ...` / `# noqa: BLE001` in several
spots) by design, to convert arbitrary user-callback exceptions into `fail()`/
`on_error`, which is reasonable for a coroutine framework, though it does mean
internal bugs in `_wait_for_child`/`_wait_for_operation` could get silently
reclassified as "chain failures" rather than surfacing loudly during development.

## Bottom line

The core chain engine is well thought-out and well tested, but the PR isn't safe to
merge as-is: the `typing.TypeAlias` import will break on Python 3.9 (a version the
project still advertises support for), and that breakage appears to be the real reason
CI's Python matrix was quietly trimmed in the same PR. I'd ask the author to fix the
`TypeAlias` compatibility (or explicitly bump minimum Python and update `setup.py`
classifiers as a conscious decision), fix or reword the `repeat_duration(duration=0)`
behavior, and consider splitting out the unrelated CI/lint/perf changes.

PR #1466, opened 2026-07-26
Fix crash in Cocoa event loop when a native event has no 'type' selector

### LINKED ISSUE #1465: Game crashes when controller disconnects
**Describe the bug**
When my controller disconnects the whole game crashes.

**System Information:**
MacOS: 26.5.2
pyglet: 3.0.dev3

**Error**
```
  File "/Users/xxx/workspace/arcade-station/src/arcade_station/__init__.py", line 20, in <module>
    main()
    ~~~~^^
  File "/Users/xxx/workspace/arcade-station/src/arcade_station/__init__.py", line 17, in main
    window.run()
    ~~~~~~~~~~^^
  File "/Users/xxx/workspace/arcade-station/.venv/lib/python3.14/site-packages/arcade/application.py", line 480, in run
    arcade.run()
    ~~~~~~~~~~^^
  File "/Users/xxx/workspace/arcade-station/.venv/lib/python3.14/site-packages/arcade/window_commands.py", line 175, in run
    pyglet.app.run(None)
    ~~~~~~~~~~~~~~^^^^^^
  File "/Users/xxx/workspace/arcade-station/.venv/lib/python3.14/site-packages/pyglet/app/__init__.py", line 87, in run
    event_loop.run(interval)
    ~~~~~~~~~~~~~~^^^^^^^^^^
  File "/Users/xxx/workspace/arcade-station/.venv/lib/python3.14/site-packages/pyglet/app/base.py", line 168, in run
    platform_event_loop.step(timeout)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/xxx/workspace/arcade-station/.venv/lib/python3.14/site-packages/pyglet/app/cocoa.py", line 267, in step
    event_type = event.type()
                 ^^^^^^^^^^
  File "/Users/xxx/workspace/arcade-station/.venv/lib/python3.14/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 1406, in __getattr__
    raise AttributeError(msg)
AttributeError: ObjCInstance b'_GCControllerAxisButtonInput' (31541734848) has no attribute type
```

SpiliosDimakopoulos:
Fixes #1465

## Problem
On macOS, `nextEventMatchingMask_untilDate_inMode_dequeue_` is documented
to only ever return an NSEvent (or None), but in practice it can hand back
a different native object -- observed as a private GameController
framework object (`_GCControllerAxisButtonInput`) right as a controller
disconnects. `CocoaPlatformEventLoop.step()` called `event.type()` on it
unconditionally, and since that object doesn't respond to `type`, the
ctypes/Objective-C bridge raised an uncaught `AttributeError` that crashed
the whole application.

## Fix
Added `CocoaPlatformEventLoop._safe_event_type()`, a small static helper
that calls `event.type()` inside a try/except AttributeError, issues a
`warnings.warn(...)` and returns `None` instead of letting the exception
propagate. `step()` now uses this helper, so a malformed native event is
skipped rather than crashing the run loop.

## Testing
Added a regression test in `tests/unit/platform/test_mac_objc.py` that
creates a real `NSObject` subclass without a `type` selector (matching
the existing test conventions in that file, which already builds throwaway
ObjC classes via `ObjCSubclass`/`ObjCClass`) and asserts
`_safe_event_type` returns `None` with a warning instead of raising.

Ran `ruff check` against the repo's own `pyproject.toml` config on the
changed file -- no new violations introduced (only pre-existing ones
already present in the file before this change).

Note: this test is mac-only (uses the real Cocoa/ObjC bridge), so I
wasn't able to execute it myself outside of macOS -- same constraint as
the rest of `test_mac_objc.py`.

### DIFF
--- a/0001-fix-cocoa-controller-disconnect-crash.patch
+++ b/0001-fix-cocoa-controller-disconnect-crash.patch
@@ -1,0 +1,85 @@
+--- a/pyglet/app/cocoa.py
++++ b/pyglet/app/cocoa.py
+@@ -2,6 +2,7 @@
+ 
+ import signal
+ import time
++import warnings
+ 
+ from pyglet import app
+ from pyglet.app.base import PlatformEventLoop, EventLoop
+@@ -241,6 +242,28 @@
+             self._timer.invalidate()
+             self._timer = None
+ 
++    @staticmethod
++    def _safe_event_type(event):
++        """Return ``event.type()``, or ``None`` if ``event`` doesn't behave like a normal NSEvent.
++
++        ``nextEventMatchingMask_untilDate_inMode_dequeue_`` is documented to only ever
++        return an NSEvent (or None), but macOS has been observed handing back a native
++        object that isn't a real NSEvent -- for example a private GameController
++        framework object such as ``_GCControllerAxisButtonInput`` right as a controller
++        disconnects. Such objects don't respond to ``type``, and calling it raises an
++        AttributeError from the ctypes/Objective-C bridge. Treat that as "nothing useful
++        to dispatch" instead of letting it crash the whole application.
++        See: https://github.com/pyglet/pyglet/issues/1465
++        """
++        try:
++            return event.type()
++        except AttributeError:
++            warnings.warn(
++                f"Ignored a native event without a usable 'type' (got {event!r}); "
++                f"this can happen around controller connect/disconnect on macOS."
++            )
++            return None
++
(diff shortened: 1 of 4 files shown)


caffeinepills:
Ty for this, however closing this as it solves the symptom not the cause. See #1469

Issue #887, opened 2023-07-06
Pyglet.shapes - Not all shapes are well behaved


nbroderick-code:
**Describe the bug**
I tested all of the shapes in the shape class, many do not behave a way that the documentation implies they would.
Here is a summary. There is a test code to help you see what I mean below.

- **Arc** # Fail, __contains__ not implemented, cannot set visible to False
- **BezierCurve** # Fail, could not initialize.
- **Circle**  # Pass
- **Ellipse** # Fail, cannot set visible to False
- **Sector** # Fail, __contains__ does not work
- **Line** # Fail, width does not update, rotation does not work, __contains__ does not work
- **Rectangle** # Fail, when rotating anchor is not center
- **BorderedRectangle** # Fail, when rotating anchor is not center, border width does not update
- **Triangle** # Fail, rotation does not work
- **Star**  # Pass
- **Polygon** # Fail, could not initialize

**System Information:**
```
_Platform
------------------------------------------------------------------------------
platform:  Windows-10-10.0.19045-SP0
release:   10
machine:   AMD64

Python
------------------------------------------------------------------------------
implementation: CPython
sys.version: 3.10.2 (tags/v3.10.2:a58ebcc, Jan 17 2022, 14:12:15) [MSC v.1929 64 bit (AMD64)]
sys.maxint: 9223372036854775807
os.getcwd(): C:\Users\Nate

pyglet
------------------------------------------------------------------------------
pyglet.version: 1.5.21
pyglet.compat_platform: win32
pyglet.__file__: C:\tools\Manim\Lib\site-packages\pyglet\__init__.py
pyglet.options['audio'] = ('xaudio2', 'directsound', 'openal', 'pulse', 'silent')
pyglet.options['debug_font'] = False
pyglet.options['debug_gl'] = True
pyglet.options['debug_gl_trace'] = False
pyglet.options['debug_gl_trace_args'] = False
pyglet.options['debug_graphics_batch'] = False
pyglet.options['debug_lib'] = False
pyglet.options['debug_media'] = False
pyglet.options['debug_texture'] = False
pyglet.options['debug_trace'] = False
pyglet.options['debug_trace_args'] = False
pyglet.options['debug_trace_depth'] = 1
pyglet.options['debug_trace_flush'] = True
pyglet.options['debug_win32'] = False
pyglet.options['debug_x11'] = False
pyglet.options['graphics_vbo'] = True
pyglet.options['shadow_window'] = True
pyglet.options['vsync'] = None
[...]

PR #1468, opened 2026-07-26
Fix Arc applying rotation twice, and implement Arc.__contains__

SpiliosDimakopoulos:
Related to #887

## Problem 1: Arc.rotation applied twice
`Arc._get_vertices()` computed its swept angle using
`start_angle - self._rotation`. This predates PR #943, which added a
shared shader-based `rotation` attribute to every shape and removed the
old per-shape rotation properties. The shader already applies
`-rotation` to every shape's local vertices, so Arc ended up rotating
at *double* the angle, in the *opposite* direction, compared to Sector,
Rectangle, and every other shape (Sector's `_get_vertices`, for
comparison, never references `self._rotation` at all).

Verified numerically by simulating the exact GLSL rotation matrix in
Python: before this fix, `world_angle = start_angle - 2*rotation`;
after, `world_angle = start_angle - rotation`, matching Sector.

## Problem 2: Arc.__contains__ not implemented
Every other shape overrides `ShapeBase.__contains__`; `Arc` didn't, so
`point in arc` raised `NotImplementedError`. Implemented it using the
same angle-wraparound logic as `Sector.__contains__`, adapted to test
an annulus (`abs(distance_to_center - radius) <= thickness / 2`)
instead of a filled disc, since Arc is a stroked shape.

## Testing
Added 9 tests to `tests/unit/shapes/test_shapes.py`, covering: points on/
off the ring, within/beyond half-thickness, inside/outside the swept
angle, and one test that isolates the double-rotation bug directly via
`_get_vertices()`, independent of the `__contains__` fix. All 9 fail
against unmodified pyglet and pass after this fix.

Ran the full existing unit test suite under Xvfb: 639 passed, 18
skipped, 2 xfailed, 2 failed -- the 2 failures are OpenAL/PulseAudio
driver tests unrelated to this change (no audio hardware in my
sandbox); confirmed identical failures on unmodified pyglet. Ran `ruff
check` on the changed file: zero new violations.

Note: this only addresses Arc from #887. That issue also covers several
other shapes (BezierCurve/Polygon init, Line, Triangle rotation,
Ellipse/BorderedRectangle) which are unrelated and unaffected by this
change.

### DIFF
--- a/pyglet/shapes.py
+++ b/pyglet/shapes.py
@@ -883,7 +883,7 @@ class Arc(ShapeBase):
         y = -self._anchor_y
         r = self._radius
         segment_radians = math.radians(self._angle) / self._segments
-        start_radians = math.radians(self._start_angle - self._rotation)
+        start_radians = math.radians(self._start_angle)
 
         # Calculate the outer points of the arc:
         points = [(x + (r * math.cos((i * segment_radians) + start_radians)),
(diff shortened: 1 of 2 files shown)


benmoran56:
Thanks for the pull request. The rotation fix looks good. 

I'm not sure if it's necessary to have such an accurate `contains` implementation. Similar to the Star, it might be enough to treat the shape as a simple circle for purposes of checking if something is within the boundary. That would be the more common use case, and would save on CPU.

PR #1470, opened 2026-07-28
Preserve fractional Win32 refresh rates

sldsmkd:
## Summary

Preserve fractional refresh rates for the active Win32 display mode.

`Win32Screen.get_mode()` currently obtains `rate` from
`DEVMODE.dmDisplayFrequency`, an integer field. On a display mode exposed by
Windows as 120000/1001 Hz, that API reports 119 Hz, which causes applications
using the value for presentation timing to run at the wrong nominal rate.

This change reuses pyglet's existing Display Configuration API path lookup to:

- match the active path to the screen by its GDI source name;
- obtain the active path's rational refresh rate;
- expose that precise value from `get_mode()` and from the matching active
  entry in `get_modes()`; and
- retain the legacy `DEVMODE` rate as a fallback when the modern query is
  unavailable, fails, races with a topology change, or returns invalid data.

Inactive enumerated modes continue to use their legacy integer rates because
the Display Configuration API describes the active path rather than every
available mode.

## Validation

- Native Windows 11, CPython 3.14.6 x64
- AMD Radeon RX 7900 XT, 3840x2160 display
- Live `ScreenMode.rate` at the Windows fractional 120 Hz mode:
  `119.88011988011988` (`120000/1001`), rather than `119`
- Focused Win32 unit tests: `10 passed`
- Linux unit collection: `629 passed, 19 skipped, 2 xfailed`; the three
  remaining failures were host audio/X11 environment failures unrelated to
  display handling
- BPS Linux suite against this checkout: `372 passed`
- `git diff --check`: passed

The focused tests cover multi-monitor GDI-name matching, buffer-size retry,
malformed rationals, precise-rate preservation, legacy fallback, and applying
the precise value only to the active enumerated mode.

### DIFF
--- a/pyglet/display/base.py
+++ b/pyglet/display/base.py
@@ -264,8 +264,8 @@ class ScreenMode:
     depth: int = None
     """Pixel color depth, in bits per pixel."""
 
-    rate: int = None
-    """Screen refresh rate in Hz."""
+    rate: int | float | None = None
+    """Screen refresh rate in Hz. May be fractional when provided by the operating system."""
 
     def __init__(self, screen: Screen) -> None:
         self.screen = screen
--- a/pyglet/display/win32.py
+++ b/pyglet/display/win32.py
@@ -45,7 +45,6 @@ if TYPE_CHECKING:
 
 _ERROR_SUCCESS = 0
 _ERROR_INSUFFICIENT_BUFFER = 122
-_QUERY_DISPLAY_CONFIG_RETRIES = 3
 
 
 def set_dpi_awareness() -> None:
@@ -107,7 +106,7 @@ class Win32Screen(Screen):  # noqa: D101
 
     def _get_display_config_path(self) -> DISPLAYCONFIG_PATH_INFO | None:
         """Get the active Display Configuration path for this screen."""
-        for _ in range(_QUERY_DISPLAY_CONFIG_RETRIES):
+        while True:
             path_count = UINT32()
             mode_count = UINT32()
 
@@ -153,7 +152,6 @@ class Win32Screen(Screen):  # noqa: D101
                     return path
 
             return None
-        return None
 
     def _get_friendly_name_display_config_api(self) -> str:
         """Get the friendly name of a monitor using the newer Display Configuration API.
--- a/pyglet/display/win32.py
+++ b/pyglet/display/win32.py
@@ -43,6 +43,10 @@ if WINDOWS_8_1_OR_GREATER:
 if TYPE_CHECKING:
     from ctypes.wintypes import HDC, HMONITOR, LPARAM, LPRECT
 
+_ERROR_SUCCESS = 0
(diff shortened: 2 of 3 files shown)


caffeinepills:
Thanks for the PR! So looking through this, I have a few questions. (The diff is hard to follow and compare, not sure why it's written like that where a few lines are removed and added 1 after the other.)

Some things
* My question would be why only tag the fractional rate for the active mode? Why not all modes? (Not saying one way or another is better, but just curious about the choice.)

* Not sure what `_QUERY_DISPLAY_CONFIG_RETRIES` is or why it's set to query the same code 3 times? Seems entirely unneeded.

> On a display mode exposed by Windows as 120000/1001 Hz, that API reports 119 Hz, which causes applications
using the value for presentation timing to run at the wrong nominal rate.

The current code shouldn't actually cause any failures from setting mode from what I can see. The data in `DEVMODE` is still being used to apply the mode (which is using the `dmDisplayFrequency` integer) regardless. So it should actually be setting the correct mode. The only way I can see this failing for some reason is if there is a real `119` refresh rate mode AND a `119.88011988011988` mode, where they are two distinct ones, in which case I could see wanting to tell the difference, but I have not seen a case of that happening before? Could you provide some additional information on the failure case?

sldsmkd:
Cleaned it up, should be easier to follow now. 

The retries has been updated to follow the MS documentation best practices as it can have a race condition. https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-querydisplayconfig

Regarding the failure case: this doesn’t prevent Windows from setting the correct display mode. My use case relies on the reported refresh rate to calculate presentation timing. When a 119.880 Hz mode is reported as 119 Hz, the engine targets half of that—59.5 FPS instead of approximately 59.94 FPS. That mismatch against the actual refresh cadence introduces periodic judder.
I only apply the precise rate to the active mode because that is the rate relevant to current presentation timing. The other enumerated modes have no bearing on this use case. Precise rates for inactive modes would become useful when selecting and switching display modes, particularly for fullscreen mode changes, but that is outside the scope of this fix.

caffeinepills:
Sorry for the delay, thank you for fixing the diff, much easier to read! I think this is pretty close, but the loop in the Microsoft example link you provided is to expand the buffer a couple times if it's too small. Currently your code does not expand the buffer during the loop, so it has no actual benefit.

sldsmkd:
The loop does reallocate — GetDisplayConfigBufferSizes is called at the top of each iteration and both buffers are allocated from the counts it returns, so an ERROR_INSUFFICIENT_BUFFER retry picks up the new sizes. That was easy to miss because the path-matching walk was also inside the loop body. I've split the query/retry into its own helper so the reallocation is the only thing in the loop, and added a test asserting the second attempt allocates the larger buffers.

caffeinepills (approved):
Looks good

PR #1484, opened 2026-08-23
Switch to epoll for the Linux event loop (instead of select)

benmoran56:
Switch to using epoll instead of select.


benmoran56:
As per the test suite, this has some bugs with PyPy, and may have contain race conditions. 

**Do not merge this yet until further investigation.**

PR #1494, opened 2026-09-06
Add basic policy and rules for AI agents.

caffeinepills:
As AI is getting more popular, we have had users requesting some additional rules, policies, and restrictions. This PR adds a basic policy and with rules for AI agents.

Feel free to discuss the proposed files and ask any questions. If you think something else belongs in the policy or agent rules, let us know what, and the reasoning for it.

The current rules generally boil down to allowing AI to assist with contributions, but it does not replace the contributor, so the human element is still required. If you submit it, you should understand the code and be able to discuss it.


benmoran56:
@caffeinepills one suggestion for AGENTS.md, regarding unit testing. Something like: 
* Tests that import platform resources (graphics, windowing, audio), should go under tests/integration/
* Tests that do not import platform resources (or can be minimally mocked) can go under tests/unit/

caffeinepills:
> @caffeinepills one suggestion for AGENTS.md, regarding unit testing. Something like:
> 
> * Tests that import platform resources (graphics, windowing, audio), should go under tests/integration/
> * Tests that do not import platform resources (or can be minimally mocked) can go under tests/unit/

Makes sense, I have added additional lines to ensure the tests are separated properly.

PR #1496, opened 2026-09-10
Preserve fractional XlibScreenModeXF86 and XlibScreenModeXrandr refresh rates

sldsmkd:
Updated both XlibScreenModeXF86 and XlibScreenModeXrandr to allow fractional refresh rates.  ScreenMode in pyglet.display.base.py was updated to support in #1470.

Tested on XWayland / Gnome on CachyOS. 

On a monitor with a fractional refresh rate set the results were the same:

XRandR  : 59.917811613014415
XF86VM  : 59.917811613014415
identical: True
raw dotclock kHz=712000 htotal=5312 vtotal=2237

### DIFF
--- a/pyglet/display/xlib.py
+++ b/pyglet/display/xlib.py
@@ -460,7 +460,7 @@ class XlibScreenXrandr(XlibScreen):
 
 
 class XlibScreenMode(ScreenMode):
-    def __init__(self, screen: XlibScreen, width: int, height: int, rate: int, depth: int):
+    def __init__(self, screen: XlibScreen, width: int, height: int, rate: float, depth: int):
         super().__init__(screen)
         self.width = width
         self.height = height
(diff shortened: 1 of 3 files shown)


benmoran56:
Hi @sldsmkd, thank you for the PR - this was lacking. 

Two things:  could you move the unit tests under tests/integration/platform? This is importing xlib, so we should try to keep it under integration. You can also use the existing pattern for detecting Linux: 

```py
from tests.annotations import Platform, require_platform

pytestmark = require_platform(Platform.WINDOWS)
```

sldsmkd:
Moved the test into integration and added pytestmark = require_platform(Platform.LINUX)

benmoran56 (approved):

benmoran56:
Thank you @sldsmkd !

PR #1501, opened 2026-09-20
Qoi image decoder

benmoran56:
This adds a pure Python QOI image codec, with both decoder and encoder. The format is dead simple, and I already had an encoder for round-trip testing, so both are included. Decoding & encoding is quite fast for pure Python, so it may be a nice alternative to the Python PNG codec for those environments that can't use anything else (web, etc.?).


caffeinepills (commented):
Looks good!

